PR #12, opened 2020-03-11
Testing linter app

UARTman:


UARTman:
Linter works as intended.

PR #37, opened 2020-04-18
Re-write DHT

justheuristic:
We had our doubts whether or not we can rely on bmuller's kademlia implementation.
In the last discussion with @mryab i pestered him into allowing me to merge kademlia into our repo with some changes.

Alas, a more detailed inspection revealed more issues that are not that easy to fix within the old implementation:
(1) the original implementation of [find_neighbors](https://github.com/bmuller/kademlia/blob/master/kademlia/routing.py#L182-L192) is incorrect if k > bucket size, which is necessary to speed-up lookups
(2) node depth [was computed](https://github.com/bmuller/kademlia/blob/master/kademlia/routing.py#L73) based on nodes in the bucket and not the range assigned to that bucket. As a result, sometimes buckets would not be split when they should have been split according to [section 4.2](https://pdos.csail.mit.edu/~petar/papers/maymounkov-kademlia-lncs.pdf)
(3) some algorithm parameters such as maximum [out-of-root split](https://github.com/bmuller/kademlia/blob/master/kademlia/routing.py#L166) ($b$) are hard-coded and hard to expose to a user
(4) beam search aka [SpiderCrawl](https://github.com/bmuller/kademlia/blob/master/kademlia/crawling.py) has unnecessary synchronizations between steps, therefore increasing alpha does not decrease latency

As a result, i would argue that it would be easy to redesign the DHT for our use case rather than modifying the original code. 

Here's a sketch:
img


justheuristic:
Implementation:
* [ ] Basic types: DHTID, KBucket
* [ ] RoutingTable
* [ ] KademliaProtocol
* [ ] DHTNode
* [ ] DHT caching

__Bugfix:__
* [ ] Expose dht parameters to user
* [ ] Fix invalid neighbor lookup - replaced with correct but inefficient implementation
* [ ] Fix KBucket.depth

__Test coverage:__
* [ ] Basic types
* [ ] RoutingTable
* [ ] Basic KademliaProtocol
* [ ] DHTNode for large number of nodes
* [ ] DHT benchmark

__Implementation differences:__
  * if a new node is rejected by RoutingTable (bucket is full and could NOT be split), 
       - bmuller's implementation would ping the first node in that bucket to *maybe* remove it if it doesn't respond; 
       - our implementation periodically pings random neighbors from all such buckets (same as with stale buckets)
   * if a node from RoutingTable doesn't respond,
        - bmuller's implementaton removes this node from routing table right away
        - [ ] our implementation does TODO

__Codebase differences:__
 * less files (7 -> 4). mostly because chunks of code were replaced with python system libraries
 * acyclic class dependencies: RoutingTable no longer calls KademliaProtocol, KademliaProtocol no longer calls DHTNode
  * DHTID is now just an integer with helper methods
  * typing and docstrings for (hopefully) all public classes and methods
  * updated to newer asyncio: ensure_future -> create_task, etc
  * our implementaton does not support saving/loading a persistent state
  * our implementaton does not supports pylint

hivemind/dht/node.py
```diff
@@ -0,0 +1,206 @@
...
+            loop.run_until_complete(self.find_nearest_nodes(query_id=self.node_id))
+
+    async def find_nearest_nodes(self, query_id: DHTID, k_nearest: Optional[int] = None,
+                                 beam_size: Optional[int] = None, exclude_self: bool = False) -> Dict[DHTID, Endpoint]:
+        """
+        Traverse the DHT and find :k_nearest: nodes to a given :query_id:, optionally :exclude_self: from the results.
+        :note: this is a thin wrapper over dht.search.beam_search, look there for more details
+        :returns: an ordered dictionary of [peer DHTID -> network Endpoint], ordered from nearest to farthest neighbor
+        """
+        k_nearest = k_nearest if k_nearest is not None else self.protocol.bucket_size
```
mryab, line 96:
```suggestion
        k_nearest = k_nearest or self.protocol.bucket_size
```

hivemind/dht/routing.py
```diff
@@ -0,0 +1,266 @@
...
+
+import hashlib
+import os
+import random
+
+import time
+import heapq
+from collections.abc import Iterable
+from itertools import chain
+from typing import Tuple, Optional, List, Dict, Set, Union, Any, Sequence, Iterator
```
mryab, line 11: `https://www.python.org/dev/peps/pep-0585/#id6`

mryab: (we'll fix that in a future PR)

PR #42, opened 2020-05-06
run_server update

justheuristic:
* torch1.5 support - fixed this bug https://circleci.com/gh/learning-at-home/hivemind/437
* better documentation for python test_utils.run_server --help
* TaskPool and connection handlers are now always ForkProcess
  - this fixes a bug that causes obscure torch errors after mp.set_start_method(other_than_fork)
* TaskPool and connections handlers are now daemons unless otherwise specified
  - this fixes a rare bug where failed background_server sometimes caused zombie processes
* minor code style fixes (line breaks, spaces)

### DIFF
--- a/tests/test_moe.py
+++ b/tests/test_moe.py
@@ -87,3 +87,8 @@ def test_compute_expert_scores():
                     "compute_expert_scores returned incorrect score"
     finally:
         dht.shutdown()
+
+
+if __name__ == '__main__':
+    test_determinism()
+    test_determinism() #TODO remove
--- a/tests/test_moe.py
+++ b/tests/test_moe.py
@@ -87,8 +87,3 @@ def test_compute_expert_scores():
                     "compute_expert_scores returned incorrect score"
     finally:
         dht.shutdown()
-
-
-if __name__ == '__main__':
-    test_determinism()
-    test_determinism() #TODO remove
--- a/tests/test_moe.py
+++ b/tests/test_moe.py
@@ -74,16 +74,17 @@ def test_compute_expert_scores():
         ii = [[4, 0, 2], [3, 1, 1, 1, 3], [0], [3, 2]]
         jj = [[2, 2, 1], [0, 1, 2, 0, 1], [0], [1, 2]]
         batch_experts = [
-            [hivemind.RemoteExpert(uid=f'expert.{ii[b][e]}.{jj[b][e]}') for e in range(len(ii[b]))]
+            [hivemind.RemoteExpert(uid=f'expert.{ii[batch_i][e]}.{jj[batch_i][e]}') for e in range(len(ii[batch_i]))]
             for b in range(len(ii))
         ]  # note: these experts do not exists on server, we use them only to test moe compute_expert_scores
         logits = moe.compute_expert_scores([gx, gy], batch_experts)
         torch.softmax(logits, dim=-1).norm(dim=-1).mean().backward()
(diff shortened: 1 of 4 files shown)


tests/test_moe.py
```diff
@@ -83,12 +83,7 @@ def test_compute_expert_scores():
 
         for b in range(len(ii)):
             for e in range(len(ii[b])):
-                assert torch.allclose(logits[b, e], gx[b, ii[b][e]] + gy[b, jj[b][e]]), "compute_expert_scores returned incorrect score"
+                assert torch.allclose(logits[b, e], gx[b, ii[b][e]] + gy[b, jj[b][e]]), \
```
mryab, line 86: Btw, we could use more explicit naming here :)

tests/test_utils/run_server.py
```diff
@@ -8,11 +8,11 @@
...
-def make_dummy_server(host='0.0.0.0', port=None, num_experts=1, expert_cls='ffn', hidden_dim=1024, num_handlers=None,
-                      expert_prefix='expert', expert_offset=0, max_batch_size=16384, device=None, no_optimizer=False,
-                      no_dht=False, initial_peers=(), dht_port=None, root_port=None, verbose=True, start=False,
-                      UID_DELIMETER=hivemind.DHTNode.UID_DELIMETER, **kwargs) -> hivemind.Server:
-    """ A context manager that creates server in a background thread, awaits .ready on entry and shutdowns on exit """
+def make_dummy_server(interface='0.0.0.0', port=None, num_experts=1, expert_cls='ffn', hidden_dim=1024,
+                      num_handlers=None, expert_prefix='expert', expert_offset=0, max_batch_size=16384, device=None,
+                      no_optimizer=False, no_dht=False, initial_peers=(), dht_port=None, root_port=None, verbose=True,
+                      UID_DELIMETER=hivemind.DHTNode.UID_DELIMETER, start=False, **kwargs) -> hivemind.Server:
+    """ Instantiate a server with several identical experts. See argparse comments below for details """
```
mryab, line 15: This is a function which is called separately from argparse (in fact, run_server.py is never called in tests). If we're trying to use this function elsewhere (think of IDE/Jupyter), the docstring with references to argparse is not going to be very useful. I suggest we move/copy the description of parameters here instead of argparse options

tests/test_utils/run_server.py
```diff
@@ -69,55 +69,69 @@ def make_dummy_server(host='0.0.0.0', port=None, num_experts=1, expert_cls='ffn'
...
-        runner.terminate()
-        runner.join()
+        if verbose:
+            print("Shutting down server...")
+        server.shutdown()
+        if verbose:
+            print("Server shut down successfully.")
 
 
 if __name__ == '__main__':
```
mryab, line 107: By the way, do we actually call run_server.py somewhere? I'm not sure this is the intended use for something located in `tests/test_utils`. If this is an example server launch script (which is used for debugging etc.), it should probably be located somewhere like `scripts/`

justheuristic: i routinely call python -m test_utils.run_server.py for prototyping, it will also be useful in tutorials. I agree that it should be moved out of test utils, but this is a topic for another pr.

justheuristic:
ya sdelal'

tests/test_utils/run_server.py
mryab: Okay, but we'd better write it down somewhere

PR #47, opened 2020-06-13
Dht documentation & cosmetic fixes

justheuristic:
* renamed Server -> HivemindServer, DHT -> HivemindDHT, updated docs
* fixed docstrings to be sphinx-compatible - mostly changed list formatting, added blank lines where necessary and changed Note -> :note:
* removed .pdf (based on prev review)
* [wontfix] tried to switch to typing from python3.9 everywhere (via `from __future__ import typing`), but pycharm failed to recognize that.


hivemind/dht/__init__.py
```diff
@@ -1,6 +1,6 @@
 """
 This sub-module implements a node in a Kademlia-based DHT. The code is organized as follows:
- * class DHT (below) - high-level class for model training. Runs DHTNode in a background process.
+ * class HivemindDHT (below) - high-level class for model training. Runs DHTNode in a background process.
```
mryab, line 3: Revert that and HivemindServer and we're good to go

PR #55, opened 2020-06-30
switch to uvloop, fix docs, fix congestion, add tests for empty routing table

justheuristic:
* hivemind.dht.DHT now uses uvloop (without affecting other processes)
* previously readthedocs failed when building both numpy, pytorch and protobuf (timeout), now it uses conda environment.yml to install a special "rtfd-only" env
* this reduced doc build time from >15 minutes to ≈5 minutes [example](https://readthedocs.org/projects/learning-at-home/builds/11357160/)
* removed tqdm and requests from requirements.txt (both unused)
* fixed broken congestion (max parallel_rpc)

### DIFF
--- a/docs/environment.yaml
+++ b/docs/environment.yaml
@@ -14,5 +14,6 @@ dependencies:
     - recommonmark
     - sphinx_rtd_theme
     - prefetch_generator>=1.0.1
+    - uvloop>=0.14.0
     - umsgpack
 
--- a/docs/environment.yaml
+++ b/docs/environment.yaml
@@ -1,4 +1,4 @@
-channels:
+catchannels:
   - anaconda
   - defaults
   - conda-forge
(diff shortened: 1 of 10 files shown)


docs/requirements.txt
```diff
@@ -1,2 +0,0 @@
-recommonmark
-sphinx_rtd_theme
```
justheuristic, line 2: moved to docs/environment.yml

docs/environment.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - pytorch>=1.3.0
+  - joblib>=0.13
+  - pip
+  - pip:
+    - recommonmark
+    - sphinx_rtd_theme
+    - prefetch_generator>=1.0.1
+    - uvloop>=0.14.0
+    - umsgpack
+
```
justheuristic, line 19: note: grpcio here is NOT 1.31, but it is enough of a mockup for docs to build correctly

PR #60, opened 2020-06-30
implement logarithmic lookup, add fast inverse lookup, make harsher tests

justheuristic:
* previously RoutingTable.get_nearest_neighbors searched through the whole heap, now it only considers a __subtree__ that is guaranteed to contain all k nearest neighbors (around 2.5-3.5x speedup for 100-1000 nodes)
* previously RoutingTable chose bucket with linear search, now it uses binary search 
   * practical speed-up is insignificant, the function itself runs ≈6x faster but it takes <1% of total time
   * added test for binary search correctness
* RoutingTable will now use dictionaries for quick DHTID => Endpoint lookup 
   * O(1) instead of O(log N) in theory, minor improvements (<5%) in practice, but its the canonical solution
   * also, RoutingTable can now do inverse lookup Endpoint => DHTID for eviction (was: return None)

### DIFF
--- a/hivemind/dht/routing.py
+++ b/hivemind/dht/routing.py
@@ -55,7 +55,7 @@ class RoutingTable:
         bucket = self.buckets[bucket_index]
         store_success = bucket.add_or_update_node(node_id, endpoint)
 
-        if node_id in bucket.nodes_to_addr or node_id in bucket.replacement_nodes:
+        if node_id in bucket.nodes_to_endpoint or node_id in bucket.replacement_nodes:
             # if we added node to bucket or as a replacement, throw it into lookup dicts as well
             self.uid_to_endpoint[node_id] = endpoint
             self.endpoint_to_uid[endpoint] = node_id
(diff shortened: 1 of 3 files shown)


hivemind/dht/routing.py
```diff
@@ -93,41 +108,45 @@ def get_nearest_neighbors(
...
         :param exclude: if True, results will not contain query_node_id even if it is in table
-        :returns: a list of tuples (node_id, endpoint) for up to k neighbors sorted from nearest to farthest
-
-        :note: this is a semi-exhaustive search of nodes that takes O(n * log k) time.
-            One can implement a more efficient knn search using a binary skip-tree in some
-            more elegant language such as c++ / cython / numba.
-            Here's a sketch
-
-            Preparation: construct a non-regular binary tree of depth (2 * DHTID.HASH_NBYTES)
-             Each leaf corresponds to a binary DHTID with '0' for every left turn and '1' for right turn
```
justheuristic, line 104: note: instead of traversing a skip-tree of nodes, we traverse a regular tree of buckets and select best nodes in buckets using heap-sort. This is 90% as fast (speculation!) but much simpler in terms of code

PR #67, opened 2020-07-06
Save checkpoints in Server

mryab:


hivemind/server/__init__.py
```diff
@@ -7,6 +7,7 @@
 
 from .connection_handler import handle_connection
 from .dht_handler import DHTHandlerThread
+from hivemind.server.checkpoint_saver import CheckpointSaver
```
justheuristic, line 10: relative import pls

hivemind/server/checkpoint_saver.py
```diff
@@ -0,0 +1,42 @@
...
+import time
+from datetime import datetime
+from pathlib import Path
+from shutil import copytree
+from tempfile import TemporaryDirectory
+from typing import Dict
+
+import torch
+
+from hivemind.runtime import ExpertBackend
```
justheuristic, line 11: relative pls. also consider merging with runtime

hivemind/server/__init__.py
mryab: Fixed

hivemind/server/checkpoint_saver.py
mryab: Fixed

PR #75, opened 2020-07-12
Simplify & explain moe.py

justheuristic:
* [ ] switch from tuple endpoints to string endpoints everywhere
* [ ] add a more feature-rich test for moe.py (with several DHT nodes and experts)
* [ ] simplify dht key format
   * __was:__ separate prefixes ["prefix.{uid_prefix}" => True], separate keys/values ["expert.{full_uid} => endpoint]
   * __new:__ just [uid_prefix => endpoint] for all prefixes (split by dot) including uid itself
   * expected benefit: make 1 less wave of requests in moe.py (2 => 1 for flat dht, 3 => 2 for our target size)
* [ ] re-write moe.py to make 1 less consecutive batches of requests
* [ ] re-write DHT.first_k_active to take advantage of get_many ( #53 )
* [ ] explain key format and guarantees in hivemind.dht.DHT docstring
* [ ] general: see if we can improve moe.py code quality
* [ ] see if we can implement parallel backward without utils/autograd.py. If we can, remove autograd.py alltogether


justheuristic:
[moved to https://github.com/learning-at-home/hivemind/pull/78 ]

PR #76, opened 2020-07-12
tuple endpoint => string endpoint

justheuristic:


hivemind/client/expert.py
```diff
@@ -22,20 +22,18 @@ class RemoteExpert(nn.Module):
...
     """
 
-    def __init__(self, uid, host='127.0.0.1', port=8080):
+    def __init__(self, uid, endpoint: Endpoint):
         super().__init__()
-        self.uid, self.host, self.port = uid, host, port
-        self._channel, self._stub = None, None
-        self._info = None
+        self.uid, self.endpoint = uid, endpoint
+        self._channel, self._stub, self._info = None, None, None  # to be filled via @property
```
mryab, line 31: IMO the comment is not necessary

PR #78, opened 2020-07-13
Simplify & explain hivemind.dht.DHT

justheuristic:
* [x] make DHT interface similar to DHTNode (e.g. initial_peers: List[str] instead of *initial_peers)
   * also: got rid of run_in_background & asyncio.run_coroutine_threadsafe in favor of async/await
   * also: DHT can now process concurrent queries on client side
* [x] simplify dht key format
   * __was:__ separate prefixes ["prefix.{uid_prefix}" => True], separate keys/values ["expert.{full_uid} => endpoint]
   * __new:__ just [uid_prefix => endpoint] for all prefixes (split by dot) including uid itself
   * expected benefit: make 1 less wave of requests in moe.py (2 => 1 for flat dht, 3 => 2 for our target size)
* [x] re-write DHT.first_k_active to take advantage of get_many ( #53 )
   * also: DHTNode.get_many/store_many now handle cancelling correctly (by stopping all unfinished workers)
* [x] explain key format and guarantees in hivemind.dht.DHT docstring
* [x] implement missing MPFuture funcitonality
   * [x] add tests for MPFuture
* Found a reason why tests hung unexpectedly (GLOBAL_EXECUTOR was corrupted due to forks)


hivemind/utils/mpfuture.py
```diff
@@ -1,147 +1,164 @@
...
+from typing import Optional
+
+from hivemind.utils.threading import run_in_background
 
 
-class MPFuture(Future):
-    """ Multiprocessing version of concurrent.futures.Future, interacts between two processes via Pipe """
-    STATES = 'pending', 'running', 'cancelled', 'finished', 'exception'
-    STATE_PENDING, STATE_RUNNING, STATE_CANCELLED, STATE_FINISHED, STATE_EXCEPTION = STATES
+class MPFuture(base.Future):
```
justheuristic, line 14: MPFuture now implements all public methods of concurrent.futures.Future (with functionality tests).

hivemind/utils/mpfuture.py
```diff
@@ -1,147 +1,164 @@
...
-            return self._result
-        elif self.state == self.STATE_EXCEPTION:
+    def __await__(self):
+        yield from asyncio.get_running_loop().run_in_executor(None, self._await_terminal_state, None).__await__()
+        if self._exception:
             raise self._exception
-        else:
-            assert self.state == self.STATE_CANCELLED
-            raise CancelledError()
+        return self._result
```
justheuristic, line 160: changed method name to match asyncio.Future

/* asyncio.Future itself is awaitable while .result / .exception return current values (if any) */

hivemind/utils/threading.py
```diff
@@ -1,27 +1,20 @@
...
 
 
 def run_in_background(func: callable, *args, **kwargs) -> Future:
     """ run func(*args, **kwargs) in background and return Future for its outputs """
-
+    global EXECUTOR_PID, GLOBAL_EXECUTOR
+    if os.getpid() != EXECUTOR_PID:
+        GLOBAL_EXECUTOR = ThreadPoolExecutor(max_workers=os.environ.get("HIVEMIND_THREADS", float('inf')))
+        EXECUTOR_PID = os.getpid()
     return GLOBAL_EXECUTOR.submit(func, *args, **kwargs)
```
justheuristic, line 16: GLOBAL_EXECUTOR used to hang non-deterministically when called in a forked process with system calls (e.g. pipe.poll/recv)

hivemind/utils/threading.py
```diff
@@ -1,27 +1,20 @@
...
-def run_forever(func: callable, *args, **kwargs):
-    """ A function that runs a :func: in background forever. Returns a future that catches exceptions """
-
-    def repeat():
-        while True:
-            func(*args, **kwargs)
-
-    return run_in_background(repeat)
-
-
```
justheuristic, line 24: run_forever is unused

hivemind/dht/__init__.py
```diff
@@ -62,16 +90,20 @@ def run(self) -> None:
...
+
+        async def _run():
+            node = await DHTNode.create(
+                initial_peers=list(self.initial_peers), listen_on=self.listen_on, parallel_rpc=self.parallel_rpc,
+                num_workers=self.max_workers or 1, **self.kwargs)
+            self._port.value = node.port
+            self.ready.set()
+
+            while True:
+                method, args, kwargs = await loop.run_in_executor(pipe_awaiter, self._pipe.recv)
```
mryab, line 103: JFYI asyncio has methods for [reading](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.connect_read_pipe) from pipes (from my understanding, only raw pipes, so you need to convert to bytes on your own). Not saying this needs to be implemented right away, but we can keep in mind that threads are not necessary here

hivemind/utils/mpfuture.py
```diff
@@ -1,147 +1,164 @@
...
 
-    async def async_exception(self, timeout=None):
-        await self._async_recv(timeout)
-        return self._exception
+    def __del__(self):
+        self._shutdown_trigger.set_result(True)
+        return self.connection.close()
```
mryab, line 164: Returning anything in `__del__` won't work, I guess?

tests/benchmark_dht.py
```diff
@@ -15,15 +15,15 @@ def random_endpoint() -> hivemind.Endpoint:
 
 
 def benchmark_dht(num_peers: int, initial_peers: int, num_experts: int, expert_batch_size: int, random_seed: int,
-                  wait_after_request: float, wait_before_read: float, wait_timeout: float, expiration_time: float):
-    old_expiration_time, hivemind.DHT.EXPIRATION = hivemind.DHT.EXPIRATION, expiration_time
+                  wait_after_request: float, wait_before_read: float, wait_timeout: float, expiration: float):
```
mryab, line 18: IMO expiration_time was more self-explanatory

tests/test_moe.py
```diff
@@ -88,4 +88,4 @@ def test_compute_expert_scores():
                                       gx[batch_i, ii[batch_i][expert_i]] + gy[batch_i, jj[batch_i][expert_i]]), \
                     "compute_expert_scores returned incorrect score"
     finally:
-        dht.shutdown()
+        dht.shutdown()
```
mryab, line 91: \n

hivemind/dht/__init__.py
```diff
@@ -16,39 +16,67 @@
...
 
     def __init__(self, listen_on: Endpoint = "0.0.0.0:*", initial_peers: Sequence[Endpoint] = (), *, start: bool,
-                 daemon: bool = True, max_workers: Optional[int] = None, parallel_rpc: Optional[int] = None, **kwargs):
+                 daemon: bool = True, max_workers: Optional[int] = None, parallel_rpc: Optional[int] = None,
+                 receiver_threads: int = 1, uid_delimiter: str = '.', expiration: float = 300, **kwargs):
         super().__init__()
         self.listen_on, self.initial_peers, self.kwargs = listen_on, initial_peers, kwargs
-        self.max_workers, self.parallel_rpc = max_workers, parallel_rpc
+        self.receiver_threads, self.max_workers, self.parallel_rpc = receiver_threads, max_workers, parallel_rpc
+        self.uid_delimiter, self.expiration = uid_delimiter, expiration
```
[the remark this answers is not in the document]
justheuristic: kk, will rollback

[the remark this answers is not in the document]
justheuristic: (agreed, will rollback)

hivemind/dht/__init__.py
justheuristic: noted
However, in this specific case we only need 1 thread and the only reason we need a thread is to avoid blocking the asyncio tasks while pipe.recv is waiting. So i would argue it is better to use executor than to implement a custom serializer for MPFuture :)

hivemind/utils/mpfuture.py
justheuristic: thx, fixed

PR #82, opened 2020-08-09
Speed up tests, shutdown threads via Event

mryab:


justheuristic (changes requested):
almost everything is fine, but
* please add some comment in moe.py that describes how gating function is averaged
* there's one edge case where moe.py won't forward on device other than cpu (see comment)

justheuristic (approved):

PR #97, opened 2020-09-06
Faster beam search: part1/2

justheuristic:
This PR has exploded into >500 SLOC, so we decided to split it into two parts:
This part: 
  * better caching & key reuse in DHTNode
  * fix rare bug in traverse_dht that caused ~0.1% change of not finding an existing key
  * update benchmarks & tests

Second part:
  * hivemind.dht.DHT.get_best_experts, hivemind.dht.DHT.batch_get_best_experts
     * the new beam search supports time_budget to balance between latency and accuracy (see part2))
  * remove async def beam_search from hivemind.client.RemoteMixtureOfExperts, replace with DHT.get_best_experts
  * more tests

TODO
* [x] add option for limited time_budget in beam search
* [x] add option to reuse concurrent traverse_dht runs for the same key
* [x] add option for automatic cache refresh
* [x] deduplicate concurrent get requests
* add tests for all new functionality
  * [x] new caching
  * [x] key reuse
  * [x] LocalStorage updates
* [x] verify no performance/accuracy drawdown in tests/benchmark_dht.py
* [x] update benchmark_dht.py

__traverse_dht bugfix:__ previously, if a worker chose a heap where all candidates are farther than upper bound, it would immediately terminate search. That's bad because there could have been other workers that would have found closer keys and continued search. Now, a search can only terminate if there are no suitable candidates AND no concurrent workers that can affect that heap. As a result, benchmark_dht now has exactly 100% accuracy for get operations at 1000+ nodes.

`python benchmark_dht.py --num_peers 1024 --num_experts 16384 --expert_batch_size 64 --expiration 99999 --increase_file_limit`
Before:
```
Increasing file limit - soft 1024=>32768, hard 1048576=>32768
Creating peers...
100%|███████████████████████████████████████████████████| 1024/1024 [01:51<00:00,  9.22it/s]
Sampled 16384 unique ids (after deduplication)
Storing peers to dht in batches of 64...
100%|█████████████████████████████████████████████████████| 256/256 [13:00<00:00,  3.05s/it]
Store success rate: 100.0% (48904 / 48904)
Mean store time: 0.015967, Total: 780.85
100%|█████████████████████████████████████████████████████| 256/256 [02:01<00:00,  2.11it/s]
Get success rate: 100.0 (16380 / 16384)
Mean get time: 0.00740, Total: 121.29011
Node survival rate: 100.000%
```

After:
```
Increasing file limit - soft 1024=>32768, hard 1048576=>32768
Creating peers...
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1024/1024 [01:45<00:00,  9.74it/s]
Sampled 16384 unique ids (after deduplication)
Storing peers to dht in batches of 64...
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 256/256 [12:07<00:00,  2.84s/it]
Store success rate: 100.0% (48920 / 48920)
Mean store time: 0.01487, Total: 727.46
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 256/256 [01:48<00:00,  2.35it/s]
Get success rate: 100.0 (16384 / 16384)
Mean get time: 0.00664, Total: 108.73952
Node survival rate: 100.000%
```

__Automatic cache refresh:__ every time DHTNode.get_* hits a local node's cache, it will mark this key as active. If an active key approaches expiration, DHTNode will launch a background task to fetch updates to that key. As a result, hivemind.DHT.get_experts and hivemind.DHT.first_k_active will no longer have temporary spikes in latency when cache gets invalidated.

__Key reuse:__ Old behavior: if you request the same key from DHT multiple times before the first request finishes, all requests would trigger a full DHT lookup and not benefit from caching (since cache wasn't available yet).
New behavior: DHTNode.get_* will check if there already are pending traverse_dht requests for the same key. If so, when one of these pending requests finishes, it will also return result to other requests for that key as long as expiration time is compatible.


docs/user/contributing.md
```diff
@@ -125,6 +126,6 @@ If one wants to account for these factors, one must introduce them manually by c
   
 
 #### Tips & tricks
-* You can find a wealth of pytorch debugging tricks at [their contributing page](https://github.com/pytorch/pytorch/blob/master/CONTRIBUTING.md).
+* You can find a wealth of pytorch debugging tricks at [their contributing page](https://tinyurl.com/pytorch-contributing).
```
justheuristic, line 129: old link wouldn't render correctly after `make html`

hivemind/__init__.py
```diff
@@ -3,4 +3,4 @@
 from hivemind.server import *
 from hivemind.utils import *
 
-__version__ = '0.8.2'
+__version__ = '0.8.3dev1'
```
justheuristic, line 6: let's call all non-released branches as such, whaddaya say?

hivemind/dht/protocol.py
```diff
@@ -265,24 +266,25 @@ class LocalStorage:
...
-
-    def remove_outdated(self):
-        while self.expiration_heap and (self.expiration_heap[0][0] < get_dht_time()
-                                        or len(self.expiration_heap) > self.cache_size):
+        self.data: Dict[DHTID, Tuple[BinaryDHTValue, DHTExpiration]] = dict()
+        self.expiration_heap: List[Tuple[DHTExpiration, DHTID]] = []
+        self.key_to_heap: Dict[DHTID, Tuple[DHTExpiration, DHTID]] = dict()
+        self.frozen = False  # if True, do not remove outdated elements
+
+    def _remove_outdated(self):
```
justheuristic, line 274: this function is no longer exposed to the user, reason: it doesn't always remove outdated elements and has a side-effect of updating heap, it isn't intended to be called by user

.circleci/config.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     steps:
       - checkout
       - python/load-cache
+      - run: pip uninstall -y pytest codecov  # temporary override for broken cache
```
justheuristic, line 12: i will check back in a week after their cache gets invalidated

hivemind/__init__.py
mryab: (discussed internally — going with option 2, versioning remains as is)

hivemind/dht/node.py
```diff
@@ -45,8 +51,10 @@ class DHTNode:
 
     """
     # fmt:off
-    node_id: DHTID; port: int; num_replicas: int; cache_locally: bool; cache_nearest: int; num_workers: int
-    refresh_timeout: float; protocol: DHTProtocol
+    node_id: DHTID; is_alive: bool; port: int; num_replicas: int; num_workers: int; protocol: DHTProtocol
+    refresh_timeout: float; cache_locally: bool; cache_nearest: int; cache_refresh_before_expiry: float
+    cache_refresh_available: asyncio.Event; cache_refresh_queue: LocalStorage
+    reuse_get_requests: bool; pending_get_requests: DefaultDict[DHTID, SortedList[_IntermediateResult]]
```
mryab, line 57: Maybe rewrite each statement on a separate line?

hivemind/dht/node.py
```diff
@@ -289,108 +319,198 @@ async def get(self, key: DHTKey, latest=False, **kwargs) -> Tuple[Optional[DHTVa
...
             kwargs["sufficient_expiration_time"] = float('inf')
         result = await self.get_many([key])
         return result[key]
 
-    async def get_many(
-            self, keys: Collection[DHTKey], sufficient_expiration_time: Optional[DHTExpiration] = None,
-            num_workers: Optional[int] = None, beam_size: Optional[int] = None
-    ) -> Dict[DHTKey, Tuple[Optional[DHTValue], Optional[DHTExpiration]]]:
+    async def get_many(self, keys: Collection[DHTKey], sufficient_expiration_time: Optional[DHTExpiration] = None,
+                       **kwargs) -> Dict[DHTKey, Union[Tuple[Optional[DHTValue], Optional[DHTExpiration]],
```
mryab, line 331: Side note: the type of return value probably means that we're doing something wrong... We should at least think of separating the logic in order to conform to the SRP

hivemind/dht/node.py
```diff
@@ -289,108 +319,198 @@ async def get(self, key: DHTKey, latest=False, **kwargs) -> Tuple[Optional[DHTVa
...
+
         :param keys: traverse the DHT and find the value for each of these keys (or (None, None) if not key found)
+        :param sufficient_expiration_time: if the search finds a value that expires after this time,
+            default = time of call, find any value that did not expire by the time of call
+            If min_expiration_time=float('inf'), this method will find a value with _latest_ expiration
+        :param kwargs: for full list of parameters, see DHTNode.get_many_by_id
+        :returns: for each key: value and its expiration time. If nothing is found, returns (None, None) for that key
+        :note: in order to check if get returned a value, please check if (expiration_time is None)
+        """
+        keys = list(keys)
```
mryab, line 344: Tuple?

hivemind/dht/node.py
```diff
@@ -400,3 +520,45 @@ async def _refresh_routing_table(self, *, period: Optional[float]) -> None:
                 await self.find_nearest_nodes(refresh_id)
 
             await asyncio.sleep(max(0.0, period - (get_dht_time() - refresh_time)))
+
+
+@dataclass(init=True, repr=True, frozen=False, order=False)
+class _IntermediateResult:
```
mryab, line 526: Maybe it's best to move the definition to the top before any classes that use it?

hivemind/dht/node.py
```diff
@@ -289,108 +319,198 @@ async def get(self, key: DHTKey, latest=False, **kwargs) -> Tuple[Optional[DHTVa
...
                 key_id, self.protocol.bucket_size, exclude=self.node_id))
 
+        # V-- this function will be called every time traverse_dht decides to request neighbors from a remote peer
         async def get_neighbors(peer: DHTID, queries: Collection[DHTID]) -> Dict[DHTID, Tuple[List[DHTID], bool]]:
             queries = list(queries)
             response = await self.protocol.call_find(node_to_endpoint[peer], queries)
             if not response:
                 return {query: ([], False) for query in queries}
 
             output: Dict[DHTID, Tuple[List[DHTID], bool]] = {}
```
mryab, line 406: Why not Tuple[Tuple[DHTID]]?

hivemind/dht/node.py
```diff
@@ -289,108 +319,198 @@ async def get(self, key: DHTKey, latest=False, **kwargs) -> Tuple[Optional[DHTVa
...
+                return {key_id: await search_result.future for key_id, search_result in search_results.items()}
+            except asyncio.CancelledError as e:  # terminate remaining tasks ASAP
+                for key_id, search_result in search_results.items():
+                    search_result.future.cancel()
+                raise e
+
+    def _reuse_finished_search_result(self, finished: _IntermediateResult):
+        expiration_time_threshold = max(finished.expiration_time or -float('inf'), finished.sufficient_expiration_time)
+        concurrent_requests = self.pending_get_requests[finished.key_id]
+        # note: this is a SortedList in the order of descending sufficient_expiration_time
```
mryab, line 439: "this is a" -> type annotation

tests/test_dht_experts.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def test_hivemind_dht():
+    peers = [hivemind.DHT(start=True)]
+    for i in range(10):
+        neighbors_i = [f'{LOCALHOST}:{node.port}' for node in random.sample(peers, min(3, len(peers)))]
+        peers.append(hivemind.DHT(initial_peers=neighbors_i, start=True))
+
+    you: hivemind.dht.DHT = random.choice(peers)
+    theguyshetoldyounottoworryabout: hivemind.dht.DHT = random.choice(peers)
```
mryab, line 16: Option 1: snake case :)
Option 2: other names for nodes?

mryab (changes requested):

docs/user/contributing.md
justheuristic: yep, its their main guarantee
so, not "absolutely guaranteed", but no worse than with a direct github link

hivemind/__init__.py
justheuristic: Agreed

hivemind/dht/node.py
justheuristic: i understand the concern, but it would be annoying without specialized IDEs like pycharm. Though, we can resolve this the same way we switched to absolute paths (with majority vote)

hivemind/dht/node.py
justheuristic: Agreed; Done

hivemind/dht/node.py
justheuristic: done, thanks

hivemind/dht/node.py
justheuristic: I agree with the concern, but the rest of the code is structured top-down (i.e. get_many -> get_many_by_id -> _update_cache), and its the same with other modules. If you don't mind, i would stick to the standard for now, but we can change it later in a less sneaky way.

hivemind/dht/node.py
justheuristic: Agreed, changed everywhere

tests/test_dht_experts.py
justheuristic: If we are to change it, i suggest we change it everywhere in a separate PR

hivemind/dht/node.py
mryab: Ok, let's create a discussion issue

tests/test_dht_experts.py
mryab: Let's create an issue for that as well (maybe collect all reformatting concerns in a single one)

mryab (approved):

SecretRose7 (commented):
FiireFox7 Financial Enterprise software systems

PR #104, opened 2020-10-04
Fix source links in RTFD

justheuristic:
There's a __[source]__ link after every class/method in RTFD
image
image

In master, this source link is broken since we switched from python setup.py develop to python setup.py install with this flag:
image

As a result, source links point to non-existant locations like
`https://github.com/learning-at-home/hivemind/envs/latest/lib/python3.7/site-packages/hivemind/dht/__init__.py#L131-L142`

This PR fixes the source link generator so that it works regardless of package installation path.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -209,14 +209,11 @@ todo_include_todos = True
 
 def setup(app):
     app.add_stylesheet("fix_rtd.css")
-    github_doc_root = 'https://github.com/rtfd/recommonmark/tree/master/doc/'  # TODO
     app.add_config_value('recommonmark_config', {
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -209,14 +209,11 @@
 
 def setup(app):
     app.add_stylesheet("fix_rtd.css")
-    github_doc_root = 'https://github.com/rtfd/recommonmark/tree/master/doc/'  # TODO
```
justheuristic, line 212: i blindly copied this line from the previous project, its not actually useful

mryab (approved):

PR #109, opened 2020-10-11
Switch to new beamsearch

justheuristic:
Changes:
* [ ] RemoteMixtureOfExperts now relies on the new n' fast beam search
* [ ] DHT.find_best_experts no longer supports first_k_active
* [ ] DHT.find_best_experts now has an option to cache _negative_ search results for EXPIRATION_TIME
* [ ] addressed all in-code TODOs from #107 
* [x] Value and expiration tuple will now use a special namedtuple type
* [ ] beam search now assembles beam top-k via a heap (better recall in few edge cases)
* [x] fixed an error that caused the new beam search to fail for 1-d grid
* [x] add a test for beam search
* [ ] add beam search correctness test for a full grid
* [x] allow arbitrary serializable sub-keys
* [ ] add test for update_routing_table that checks for key transfer


justheuristic:
moved to #111

Issue #95, opened 2020-08-30
Gating function averaging


justheuristic:
In our preliminary experiments, all peers have independent gating functions and we can only synchronize them manually. It would be great to implement some sort of builtin averaging mechanism.

For instance, every T seconds, assemble peers into groups at random, then perform all-reduce within each group. In case of failure, rollback and repeat T seconds later.

justheuristic:
Based on discussion with @mryab 

Formally speaking, you have a graph ![formula](https://render.githubusercontent.com/render/math?math=G(V,E)) where vertices are participants and edges represent peer relations. Each peer has a vector of parameters ![formula](https://render.githubusercontent.com/render/math?math=v_i) .

We want __all__ peers to converge to the global average 

Using the notation above, consider an (unknown) global average
![formula](https://render.githubusercontent.com/render/math?math=v^*%20%3D%20\sum_{v_i\in%20V}{v_i})

We consider two solutions:
1. each participant periodically sends his gating function to M random peers. When ![formula](https://render.githubusercontent.com/render/math?math=v_j), receives gating function from ![formula](https://render.githubusercontent.com/render/math?math=v_i), he updates his parameters with a moving average ![formula](https://render.githubusercontent.com/render/math?math=v_j%3A%3D{\alpha\cdot%20v_i%2B(1%2D\alpha)\cdot{v_j}})
2. participants ~somehow form groups of up to M members and run butterfly all-reduce between each other

The first approach is simpler to implement but requires O(NM) total communication for N peers, the second approach requires only O(N) communication but it can be tricky to implement with fault tolerance.


Notes:
* Should we send parameters with optimizer stats or just the parameters?

justheuristic:
Based on discussion with @shekhale

Using the notation above, consider an (unknown) global average ![formula](https://render.githubusercontent.com/render/math?math=v^*%20%3D%20\underset{v_i\in%20V}{E}{v_i})

We can postulate the problem of global averaging as an optimization problem with mean squared error objective
![formula](https://render.githubusercontent.com/render/math?math=L%3D\underset{v_i\in%20V}{E}{(v_i-v^*)^2})

The problem with approach 1 (prev post) is that each time one of the peers updates his parameters, the global average changes. As a result, the network will converge to a different global average depending on network structure and the order in which nodes perform averaging.

For approach 2, the AllReduce op for any subset of nodes will preserve the global average

Proof, sketch:
[...]

PR #115, opened 2020-10-28
[rfc] gating function averaging

justheuristic:
Based on #95 , discussion with @mryab 

__TL;DR__ This PR introduces two new objects, DecentralizedAverager and DecentralizedOptimizer, that allow participants to average their gating function (and whatnot) with other peers using a DHT.

__DecentralizedOptimizer__ works exactly like normal pytorch optimizer, but on .step() it will average your model with other peers according to the algorithm below. It runs DecentralizedAverager in background.

__DecentralizedAverager__ is a more low-level interface to averaging that does all the heavy lifting. It periodically finds forms groups with other nodes using DHT and performs allreduce. The algorithm for forming groups is designed so that the group size is small, but each participant converges to global average parameters after a small number of steps (see below, its tricky).

__Algorithm 1:__ DecentralizedAverager pseudocode
```python
def run_in_background(my_node: DHTNode, my_endpoint: Endpoint, num_buckets: int, group_size: int):
  my_bucket_index = randint(0, num_buckets)

  while True:
    store_ok = my_node.store(
        key='averaging.{my_bucket_index}', subkey=my_node.node_id, value=my_endpoint)
    if not store_ok:
      continue  # bad bucket configuration or DHT failure, throw a warning

    while True:
      bucket_nodes = dht_node.get("averaging.{my_bucket_index}")
      if my_node.id not in bucket_nodes:
        break  # expired too fast or DHT failure, throw a warning
      else if len(bucket_nodes) < group_size:
        continue  # wait for more nodes to join. 
      else:  # found at least group_size peers
        my_group = get_my_group(bucket_nodes)
        my_shard_index, active_peers = connect_to_peers(my_group)
        averaged_vector = allreduce(my_shard_index, active_peers)
        update_model(averaged_vector)

        # here's the tricky part. We know that my_shard_index is unique within group
        # AND we don't want to average with the same group again, SO we can just set
        my_bucket_index = my_shard_index   # ... to make sure we won't meet again
        break  # start over
```

__Subroutines:__
* __get_my_group(bucket_nodes: Sequence[Tuple[DHTID, Endpoint]]):__ repeatedly try to add peers in the same bucket starting from one's own node ID. If contacted by another node ID with earlier expiration time, yield priority to that node.
* __connect_to_peers(group: Sequence[Tuple[DHTID, Endpoint]]):__ establish connection with every peer, drop unresponsive peers, agree on (unique) shard ids within group.
* __allreduce(my_shard_index: int, active_peers: Sequence[Endpoint]):__ - butterfly-style allreduce, each node accumulates one part of the full vector according to its shard index, then distributes the result to all nodes.

__Notes:__
* we should adjust group size on the fly:
    * if there are less then group_size peers after K consecutive retries, halve the number of buckets (or give up and average regardless?)
    * If bucket size is larger than group_size * C, double the number of buckets
* shard indices must be shuffled within group to prevent potential harmful side-effects for repeated averaging (e.g. we shoult NOT order shards by ascending node id)
* make sure this won't result in huge-ass dictionaries with thousands of peers.
   * for instance, set up expiration time such that one node is simultaneously present in at most O(1) DHT entries
* if allreduce fails, we should form a new group and try again
* `my_bucket_index = my_shard_index` is a trick that is optimal if sqrt(total_trainers) <= group_size < total_trainers, but it may be suboptimal outside this range
    * if group_size equals total node size, we can average with the same group right away
    * if group size < sqrt(total_trainers), we need to avoid peers from previous T iterations. I'm unsure if we can generalize to that case (upd - it seems we can, see comments below).

* on the first round of averaging, buckets can be uneven due to random initial picks. However, after each allreduce run, the group will be uniformly distributed to other buckets. Hence, the distribution will eventually become uniform.

### DIFF
--- a/docs/modules/client.rst
+++ b/docs/modules/client.rst
@@ -17,3 +17,9 @@
 .. autoclass:: RemoteMixtureOfExperts
    :members:
    :member-order: bysource
+
+.. autoclass:: DecentralizedOptimizer
+   :members:
+
+.. autoclass:: DecentralizedAverager
+   :members:
--- a/docs/modules/client.rst
+++ b/docs/modules/client.rst
@@ -20,6 +20,8 @@
 
 .. autoclass:: DecentralizedOptimizer
    :members:
+   :member-order: bysource
 
 .. autoclass:: DecentralizedAverager
    :members:
+   :member-order: bysource
--- a/hivemind/client/__init__.py
+++ b/hivemind/client/__init__.py
@@ -1,2 +1,3 @@
 from hivemind.client.expert import RemoteExpert
 from hivemind.client.moe import RemoteMixtureOfExperts
+from hivemind.client.averaging import DecentralizedAverager, DecentralizedOptimizer
--- a/hivemind/client/__init__.py
+++ b/hivemind/client/__init__.py
@@ -1,3 +1,4 @@
 from hivemind.client.expert import RemoteExpert
 from hivemind.client.moe import RemoteMixtureOfExperts
-from hivemind.client.averaging import DecentralizedAverager, DecentralizedOptimizer
+from hivemind.client.averaging import DecentralizedAverager
+from hivemind.client.optimizer import DecentralizedOptimizer
--- a/hivemind/client/averaging.py
+++ b/hivemind/client/averaging.py
@@ -253,19 +253,18 @@ class DecentralizedAverager(mp.Process, averaging_pb2_grpc.DecentralizedAveragin
             async with self._state.one_request_at_a_time:
                 if not isinstance(self._state, (LookingForGroup, LeaderWaitingForPeers)):
                     if isinstance(self._state, LookingForGroup):
-                        logger.debug(f"Starting a new group as a leader. Group id: {self._state.group_id}")
                         self._state = LeaderWaitingForPeers(
                             leader_endpoint=self._state.my_endpoint, leader_expiration=self._state.my_expiration,
                             group_id=DHTID.generate().to_bytes(), group_endpoints={self._state.my_endpoint})
                         # note: we generate group_id as DHTID for convenience only. There is no relation to dht keys.
+                        logger.debug(f"Starting a new group as a leader. Group id: {self._state.group_id}")
 
         group_state = self._state
         assert isinstance(group_state, LeaderWaitingForPeers), "Internal error: i should be a leader at this point"
         assert peer not in group_state.group_endpoints
 
-
         try:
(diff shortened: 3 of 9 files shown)


justheuristic:
Update: we **can** generalize the my_bucket_index=my_shard_index trick to N dimensions:

__Sketch:__ Imagine we have a 10x10x10 "cube" and want to perform two averaging rounds along each dimension. Each averaging uses 10 nodes, there are 100 buckets. Let's view them as a 10x10 grid of buckets with first and second "digits"

* __First round:__ my_bucket_index = randint(100) = randint(10) * 10 + randint(10)
* __Second round:__ my_bucket_index = round1_shard_index * 10 + randint(10)
* __Third round:__ my_bucket_index = round2_shard_index * 10 + round1_shard_index
* __Subsequent rounds:__ my_bucket_index = prev_shard_index * 10 + prevprev_shard_index

justheuristic:
__Note:__ we probably should change the grid dimensions in case of undercomplete / overcomplete buckets. 
* if we found less than M/2 peers, drop one dimension
* if we found more than (M ^ 2) / 2 peers, increase grid dimension

/* maybe load-balance if our bucket is deep, but not too deep (pick random other bucket). But shouldn't we auto-balance after T steps anyway? */

There's probably a better way to set thresholds, think again before merging

justheuristic:
* [ ] DecentralizedAverager
   * [ ] implement allreduce protocol
       * [ ] add a basic test with manual group composition
   * [ ] implement LFG protocol
       * [ ] dht.get_my_endpoint? or use the same policy as in the server
   * [ ] on shutdown: gracefully send notifications to current group, remove keys from DHT, etc.
   * [ ] implement some test to verify that the averaged params are the same as latest model params, warn if not

* [ ] DecentralizedOptimizer
   * [ ] actually implement it
       * strategy: run averaging on a copy of params in background, synchronize only at optimizer.step()
   * [ ] allow specifying target allreduce frequency
   * [ ] add a basic test to verify that params after averaging are the same
   * [ ] test that params do not change midway through testing

justheuristic:
image
Upd: this PR has become unreviewable, we decided to split it into three chunks:

Preliminarily, 
1. basic allreduce protocol given a group
2. group search protocol
3. DecentralizedOptimizer

This PR will stay open while we process the parts

justheuristic:
This PR was split into a sequence of smaller PRs. We keep the original PR to #95 #123 #134 (and a few more to come)

Issue #98, opened 2020-09-08
v0.9 refactoring concerns


justheuristic:
As before ( #64 ), we have a number of things we may want to do right before the v0.9 release
* [ ] [RemoteMixtureOfExperts](https://github.com/learning-at-home/hivemind/blob/9c1e14aca11cdbc2011e0783c64a9cdc36df23e5/hivemind/client/moe.py#L44) uses grid_size while server side uses expert_pattern (e.g. `ffn.[0:256].[0:256]`), should we switch to expert_pattern everywhere? (@justheuristic )
* [ ] Should we use __slots__ for data structures such as [_IntermediateResult](https://github.com/learning-at-home/hivemind/blob/9c1e14aca11cdbc2011e0783c64a9cdc36df23e5/hivemind/dht/node.py#L526)
* [ ] Should we switch to declaring objects bottom-up vs top-down? For instance, [_IntermediateResult](https://github.com/learning-at-home/hivemind/blob/9c1e14aca11cdbc2011e0783c64a9cdc36df23e5/hivemind/dht/node.py#L526) is used before its declared (@mryab )
* [ ] [theguyshetoldyounottoworryabout](https://github.com/learning-at-home/hivemind/blob/9c1e14aca11cdbc2011e0783c64a9cdc36df23e5/tests/test_dht_experts.py#L16) rename it? (@mryab )
* [ ] Can we get rid of return_futures=True option in node.get_*
* [x] rename _IntermediateResult to SearchState, rationale: it is not necessarily a result, it also stores search data other than the result (@justheuristic )
  * [x] await SearchState.future -> await SearchState?
  * [x] SearchState.future.cancel() -> SearchState.finish_search()?
* [x] rename LocalStorage to TimedStorage (or similar)? reason: it is not always used as node's local storage (@justheuristic )
* [ ] make alias DHTID.from(x) = DHTID.generate(source=x)? (@justheuristic )
* [x] why are we using [umsgpack](https://github.com/vsergeev/u-msgpack-python) instead of just [msgpack](https://github.com/msgpack/msgpack-python#packingunpacking-of-custom-data-type)? (@justheuristic )
* [ ] we force the same vector compression rules during forward & backward pass #99 (@Vsevolod-pl )
* [x] update documentation on hivemind.dht: subkeys, beam_search, no first_k_active
* [x] update documentation on hivemind.server: compression
* [ ] consider setting maxsize to DHTValueType, e.g. 10k
[...]

PR #120, opened 2020-11-17
Process-wide channel cache for gRPC+aio

justheuristic:
* [x] process-wide grpc channel/stub caching
    * implemented hivemind.utils.grpc.ChannelCache
    * extracted TimedStorage from hivemind.dht.storage to hivmind.utils.timed_storage
    * [x] use in hivemind/client/expert.py
        * [x] verify no performance degradation
    * [x] use in hivemind/dht/protocol.py
        * [x] verify no performance degradation (upd: got faster)
    * [x] use in  hivemind.client.allreduce.py
        * [x] verify no performance degradation
* [ ] dht.get_endpoint() to discover this peer's current public endpoint from peers
* [ ] implement a basic synchronization check for get_dht_time()

### process-wide channel caching
Currently, each time we query a remote peer with gRPC, we have to create a new channel.
In contrast, gRPC best practices [recommend](https://docs.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-5.0) reusing channels for multiple rpc calls

hivemind/client/expert.py introduced channel caching, but it had a side-effect of keeping channels open forever

In this PR we implement a process-wide ChannelCache object that keeps track of open channels. The code is largely inspired by https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/_simple_stubs.py , but with the added support for grpc.aio channels.


justheuristic:
upd - from discussion with @mryab  - 
image

we decided to postpone these two to a #98 as they are not essential for ICML experiments

hivemind/client/allreduce.py
```diff
@@ -150,9 +150,7 @@ async def accumulate(self, source: Endpoint, part: torch.Tensor) -> torch.Tensor
         return await self.averaged_part
 
     def _get(self, peer: Endpoint) -> averaging_pb2_grpc.DecentralizedAveragingStub:
```
mryab, line 152: I believe _get is slightly overloaded here, how about _get_peer_stub?

hivemind/utils/grpc.py
```diff
@@ -1,17 +1,151 @@
 """
 Utilities for running GRPC services: compile protobuf, patch legacy versions, etc
 """
+from __future__ import annotations
+import os
+import threading
+from typing import NamedTuple, Sequence, Tuple, Optional, Union, Any, Dict, TypeVar, Type
```
mryab, line 7: If we're using `from __future__ import annotations`, do we need to import tuple and dict from typing?

hivemind/utils/grpc.py
```diff
@@ -1,17 +1,151 @@
...
+
+            return stubs[stub_type]
+
+    @classmethod
+    def _create_channel(cls, target: Endpoint, aio: bool, options: Sequence[Tuple[str, Any], ...],
+                        channel_credentials: Optional[grpc.ChannelCredentials],
+                        compression: Optional[grpc.Compression]) -> Union[grpc.Channel, grpc.aio.Channel]:
+        namespace = grpc.aio if aio else grpc
+        if channel_credentials is None:
+            logger.debug(f"Creating insecure {namespace} channel with options '{options}' " +
```
mryab, line 107:
```suggestion
            logger.debug(f"Creating insecure {namespace} channel with options '{options}' "
```

hivemind/utils/grpc.py
```diff
@@ -1,17 +1,151 @@
...
+                         f"and compression '{compression}'")
+            return namespace.insecure_channel(target, options=options, compression=compression)
+        else:
+            logger.debug(f"Creating secure {namespace} channel with credentials '{channel_credentials}', "
+                         + f"options '{options}' and compression '{compression}'")
```
mryab, line 112:
```suggestion
                         f"options '{options}' and compression '{compression}'")
```

hivemind/utils/grpc.py
```diff
@@ -1,17 +1,151 @@
...
+class ChannelCache(TimedStorage[ChannelInfo, Tuple[Union[grpc.Channel, grpc.aio.Channel], Dict]]):
+    """
+    A process-wide cache of gRPC channels, supports both normal and aio channels, secure/insecure channels, etc
+    Based on grpcio internal channel cache by Richard Belleville and Lidi Zheng (thanks!)
+    Unlike TimedStorage, ChannelCache actively evicts stale channels even if the cache is not accessed
+    Unlike grpc._simple_stubs.ChannelCache, this implementation supports aio and does not forcibly close active channels
+    """
+    MAXIMUM_CHANNELS = os.environ.get("GRPC_PYTHON_MANAGED_CHANNEL_MAXIMUM", 4096)
+    EVICTION_PERIOD_SECONDS = os.environ.get("GRPC_PYTHON_MANAGED_CHANNEL_EVICTION_SECONDS", 10 * 60)
+    logger.debug(f"Eviction period = {EVICTION_PERIOD_SECONDS}s, max channels = {MAXIMUM_CHANNELS}")
```
mryab, line 41: These constants should be module-level

hivemind/utils/grpc.py
```diff
@@ -1,17 +1,151 @@
...
+
+    _singleton: Optional[ChannelCache] = None
+    _singleton_pid: int = os.getpid()
+    _lock: threading.RLock = threading.RLock()
+    _update_eviction_evt: threading.Event = threading.Event()
+    _eviction_thread: threading.Thread
+    _nearest_expiration_time: DHTExpiration
+    _is_active: bool
```
mryab, line 49: These are instance attributes and are declared during init

mryab (changes requested):

hivemind/client/allreduce.py
justheuristic: its mostly a convention with DHTProtocol._get, should we rename both?

hivemind/utils/grpc.py
justheuristic: i agree, but i would request to keep it as is AND write a todo for #98 , ok?

hivemind/utils/grpc.py
justheuristic: (fixed)

hivemind/utils/grpc.py
justheuristic: (fixed, thanks!)

justheuristic: upd - it complicates testing, added to #98

mryab (approved):

PR #121, opened 2020-11-17
Fix DHT listening address, allow starting a server with no experts

mryab:


scripts/run_server.py
```diff
@@ -60,13 +57,12 @@ def main():
...
 
     compression_name = args.pop("compression")
-    compression = CompressionType.NONE
     if compression_name == "MEANSTD":
         compression = CompressionType.MEANSTD_LAST_AXIS_FLOAT16
     elif compression_name == "FLOAT16":
         compression = CompressionType.FLOAT16
     else:
-        compression = getattr(CompressionType, compression_name)
+        compression = getattr(CompressionType, 'NONE')
```
justheuristic, line 65: This code was meant to
A) allow user to directly specify compression, e.g. MEANSTD_LAST_AXIS_FLOAT16
B) but still support shortcuts such as MEANSTD (and potentially more)

You are correct that the code should be simplified, i merely suggest that you keep getattr(CompressionType, compression_name) . And maybe remove the FLOAT16 clause as it is covered by getattr

justheuristic (approved):
Great that you noticed it :)

PR #129, opened 2020-12-24
gRPC connection keepalive

mryab:


justheuristic:
Great job!

LGTM, but please increase ping interval to prevent it from affecting performance under a large number of peers

For the record
image

justheuristic (approved):

PR #132, opened 2020-12-26
Add anomaly detection to RemoteMixtureOfExperts

mryab:
This PR implements client-side anomaly detection in RemoteMixtureOfExperts: to prevent network contamination, any forward/backward pass will raise an error if passed a tensor with NaN/inf values. To prevent spreading errors from "broken" experts, we also ignore their outputs when collecting the forward/backward results of a mixture

### DIFF
--- a/hivemind/__init__.py
+++ b/hivemind/__init__.py
@@ -3,4 +3,4 @@ from hivemind.dht import *
 from hivemind.server import *
 from hivemind.utils import *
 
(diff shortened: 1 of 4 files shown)


justheuristic (approved):
Great job!
My only issue is that this behavior is not necessarily expected by the user (and potentially detrimental to performance).

It would be nice to implement it as an option that can be turned on if necessary.

PR #134, opened 2021-01-04
Address averaging corner cases, add benchmark_averaging.py, chunk averaged tensors, fix DHTNode get

justheuristic:
DecentralizedAverager functionality:
* [x] fixed several grpc-related bugs
* [x] resolved a deadlock that previously occurred when averagers made cyclic requests A->B->C->A
* [x] disabled redirect from NOT_A_LEADER (almost no performance benefit, significantly more network load)

Averaging stability & performance tweaks:
* [x] works under U(0, 1000)ms latency
* [x] works reliably over 1hour on our test cluster
   * [x] still works with random node kills (via instance hard reset)
* [x] works with 100 asynchronous averaging runs with 256 peers
* [x] works reasonably fast in crowded buckets (e.g. 256 nodes in a bucket of size 16)
* [x] tested matchmaking for larger groups, up to 256 nodes per group (matchmaking is no longer a bottleneck)
* [x] protobuf-serialized tensors are now split into chunks of fixed size (64KB) to speed up transfer under latency

Miscellaneous:
* [x] added a basic benchmark for averaging
* [x] DHTNode.get will now correctly forward kwargs (instead of ignoring them)
* [x] _SearchResult.add_candidate can now merge keys from multiple DictionaryDHTValue instances into one
  * motivation: previously, if a node was the last one to update a DictionaryDHTValue, than it would sometimes fail to fetch all keys even with DHTNode.get(..., __latest=True__) because its local value technically has latest expiration time
* [x] MPFuture will now handle EOF instead of throwing EOFError
* [x] Extracted the code that initialized event loop in DHT and DecentralizedAverager daemons into a separate function
* [x] Silenced CUDAInitialization error from cpu pytorch https://github.com/pytorch/pytorch/issues/47038
* [x] Added TimedStorage.clear() method that is equivalent to list.clear() and set.clear()


hivemind/dht/__init__.py
```diff
@@ -494,13 +490,14 @@ def declare_averager(self, group_key: GroupKey, endpoint: Endpoint, expiration_t
     async def _declare_averager(self, node: DHTNode, *, group_key: str, endpoint: Endpoint,
                                 expiration_time: DHTExpiration, looking_for_group: bool, future: MPFuture):
         try:
-            expiration_time = expiration_time if looking_for_group else nextafter(expiration_time, float('inf'))
+            expiration_time = expiration_time if looking_for_group else float(nextafter(expiration_time, float('inf')))
```
justheuristic, line 493: this caused serialization error (can't msgpack a np.float64)

mryab (changes requested):

mryab (approved):

PR #135, opened 2021-01-09
Remove duplicate log entries, report aggregate runtime performance and expert parameter count

mryab:
Also removed verbosity options in favor of logging and handled `KeyboardInterrupt`s in most processes/threads created inside Server

### DIFF
--- a/hivemind/client/averaging/__init__.py
+++ b/hivemind/client/averaging/__init__.py
@@ -23,7 +23,7 @@ from hivemind.proto import averaging_pb2, averaging_pb2_grpc, runtime_pb2
 StreamCallToLeader = grpc.aio.UnaryStreamCall[averaging_pb2.JoinRequest, averaging_pb2.MessageFromLeader]
 
 INITIAL_GROUP_NBITS = 3
-logger = get_logger(__file__)
+logger = get_logger(__name__)
 
 
 class DecentralizedAverager(mp.Process, averaging_pb2_grpc.DecentralizedAveragingServicer):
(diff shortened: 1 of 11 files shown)


hivemind/client/averaging/__init__.py
```diff
@@ -23,7 +23,7 @@
 StreamCallToLeader = grpc.aio.UnaryStreamCall[averaging_pb2.JoinRequest, averaging_pb2.MessageFromLeader]
 
 INITIAL_GROUP_NBITS = 3
-logger = get_logger(__file__)
+logger = get_logger(__name__)
```
mryab, line 26: There is only one logger per module, and using __file__ instead of __name__ leads to incorrect module name resolution and duplicate handlers

hivemind/server/__init__.py
```diff
@@ -69,7 +69,7 @@ def __init__(
     @staticmethod
     def create(listen_on='0.0.0.0:*', num_experts: int = None, expert_uids: str = None, expert_pattern: str = None,
                expert_cls='ffn', hidden_dim=1024, optim_cls=torch.optim.Adam, num_handlers=None, max_batch_size=4096,
-               device=None, no_dht=False, initial_peers=(), dht_port=None, verbose=True,
+               device=None, no_dht=False, initial_peers=(), dht_port=None,
```
mryab, line 72: Logging level is our tool of choice here for controlling verbosity

hivemind/server/runtime.py
```diff
@@ -1,9 +1,13 @@
 import multiprocessing as mp
 import multiprocessing.pool
 import threading
+from collections import defaultdict
 from itertools import chain
+from queue import SimpleQueue
 from selectors import DefaultSelector, EVENT_READ
-from typing import Dict
+from statistics import mean
```
justheuristic, line 8: i didn't know this was a member of python standard library, thanks :)

justheuristic (approved):
[reviewed, LGTM as is]

PR #137, opened 2021-01-14
Add DHT peer validation, add DHT.get_visible_address, add blacklist for unresponsive peers

justheuristic:
* DHT.get_visible_address will try to infer this peer's visible address in the shared network
* DHTNode will now validate that it is accessible by peers and synchronized within the allowed discrepancy
* DHTNode will now temporarily blacklist unresponsive peers from all store/get/find_nearest_nodes calls
   *  if the same peer fails to respond several times in a row without any successes, blacklist time will double (aka exponential backoff)
* DecentralizedAverager now has an optional "learning rate" parameter (useful to emulate gossip or add up deltas)


hivemind/dht/__init__.py
```diff
@@ -181,6 +180,53 @@ def shutdown(self) -> None:
...
+        assert not isinstance(peers, str) and isinstance(peers, Sequence), "Please send a list / tuple of endpoints"
+        future, _future = MPFuture.make_pair()
+        self.pipe.send(('_get_visible_address', [], dict(num_peers=num_peers, peers=peers, future=_future)))
+        return future.result()
+
+    async def _get_visible_address(self, node: DHTNode, num_peers: Optional[int], peers: Sequence[Endpoint],
+                                   future: Optional[MPFuture]):
+        if not peers and not node.protocol.node_info.endpoint:
+            peers_and_endpoints = node.protocol.routing_table.get_nearest_neighbors(
+                DHTID.generate(), num_peers or 1, exclude=node.node_id)
```
mryab, line 201: If there are no peers and peers was not passed, we should not silently replace this with 1

hivemind/dht/__init__.py
```diff
@@ -181,6 +180,53 @@ def shutdown(self) -> None:
...
 
+    def get_visible_address(self, num_peers: Optional[int] = None, peers: Sequence[Endpoint] = ()) -> Hostname:
+        """
+        Get this machine's visible address by requesting other peers or using pre-specified network addresses.
+
+        :param num_peers: if specified, ask multiple peers and chek that they perceive the same endpoint
+        :param peers: if specified, ask these exact peers instead of choosing random known peers
+        :note: if this node has no known peers in routing table, one must specify :peers: manually
+        """
+        assert num_peers is None or peers == (), "please specify either a number of peers, but not both"
```
mryab, line 191:
```suggestion
        assert num_peers is None or peers == (), "please specify either a number of peers or the list of peers, but not both"
```

hivemind/dht/__init__.py
```diff
@@ -181,6 +180,53 @@ def shutdown(self) -> None:
...
+                address = strip_port(endpoint)
+                if chosen_address is not None and address != chosen_address:
+                    logger.warning("At least two peers returned different visible addresses for this node:"
+                                   f"{address} and {chosen_address} (keeping the former one)")
+                chosen_address = chosen_address or address
+
+            if chosen_address is None:
+                logger.warning(f"None of the selected peers responded with an address ({peers})")
+
+        if not chosen_address and node.protocol.node_info.endpoint:
```
mryab, line 221: Should we make any requests if node.protocol.node_info.endpoint is not None? Maybe it's better to return this value right away (otherwise, the priority is not clear)

hivemind/dht/__init__.py
```diff
@@ -181,6 +180,53 @@ def shutdown(self) -> None:
...
+    async def _get_visible_address(self, node: DHTNode, num_peers: Optional[int], peers: Sequence[Endpoint],
+                                   future: Optional[MPFuture]):
+        if not peers and not node.protocol.node_info.endpoint:
+            peers_and_endpoints = node.protocol.routing_table.get_nearest_neighbors(
+                DHTID.generate(), num_peers or 1, exclude=node.node_id)
+            peers = tuple(endpoint for node_id, endpoint in peers_and_endpoints)
+
+        chosen_address = None
+        if peers:
+            endpoint_variants: Sequence[str] = await asyncio.gather(*(
```
mryab, line 206:
```suggestion
            possible_endpoints: Sequence[str] = await asyncio.gather(*(
```

mryab: "variant" usually means "slight difference"

mryab (changes requested):

mryab (approved):

PR #138, opened 2021-01-14
Load checkpoints on server start

mryab:


hivemind/dht/__init__.py
```diff
@@ -155,6 +155,7 @@ async def _run():
                 method, args, kwargs = await loop.run_in_executor(pipe_awaiter, self._pipe.recv)
                 asyncio.create_task(getattr(self, method)(node, *args, **kwargs))
 
+        logger.info(f"Running DHT node on port {self.port}, initial peers = {self.initial_peers}")
```
justheuristic, line 158: please move this inside _run AFTER the node is created. Otherwise the port is undefined.

mryab: Moved the logger call back

justheuristic (approved):
LGTM

PR #142, opened 2021-01-23
fix problem with NamedTuple inheritance in python3.9

xtinkt:


justheuristic:
Thanks!

Could you please do three more things:
- specify frozen=True - so that we can be sure HeapEntries are immutable
- add python3.9 tests via circleci (if you don't know how to do that, please ping me, i'll guide yo through it)
- bump version (PR rules)

justheuristic (approved):
LGTM

PR #144, opened 2021-02-19
Fix flaky test_remote_module_call

MaximKsh:
increase tolerance to 1e-5

### DIFF
--- a/tests/test_moe.py
+++ b/tests/test_moe.py
@@ -125,6 +125,7 @@ def test_beam_search_correctness():
         assert np.allclose(true_best_scores, our_best_scores)
 
 
+
 @pytest.mark.forked
 def test_determinism():
     atol = 1e-5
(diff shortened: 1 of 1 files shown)


tests/test_moe.py
```diff
@@ -88,7 +88,7 @@ def test_remote_module_call():
         out3 = real_expert(dummy_x)
         assert out3.shape == (3, 1024)
         out3_again = real_expert(dummy_x[1:])
-        assert torch.allclose(out3_again, out3[1:], atol=1e-6, rtol=0)
+        assert torch.allclose(out3_again, out3[1:], atol=1e-5, rtol=0)
```
justheuristic, line 91: [confirmed, this check had a small probability to fail with atol=1e-6]

justheuristic (approved):

PR #145, opened 2021-02-20
Delay importing grpcio-tools till after its installed

yhn112:
A quick fix for the problem with installation via `pip install .`

Now, `grpcio-tools` will be installed before the whole installation process, and will be imported only after being installed.

(Plz, check that I'm not an idiot somewhere here)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,6 @@ import glob
 import os
 import re
 
-import grpc_tools.protoc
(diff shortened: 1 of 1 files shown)


justheuristic (approved):
Seems legit
/* i reproduced this in a clean environment, pip install -e . now works with no deps */

PR #153, opened 2021-02-26
Compiling libp2p daemon on setup

deniskamazur:


justheuristic:
image
Initial code review:
* Please implement safe go version comparison via packaging.version
* If possible, replace os.system calls to curl/tar with urllib and tarfile python modules
* os.system('go install ...
  *  If this command fails, make sure to interrupt installation
  *  It would be great to capture logs from this call and print them to main stdout/err

justheuristic:
[Merging into libp2p branch]

PR #154, opened 2021-02-27
Load state from peers in DecentralizedAverager

justheuristic:
* [x] implemented DecentralizedAverager.load_state_from_peers that attempts to load the training state from another averager
   * The donor averager is chosen in the order from latest successfully updated to earliest
   * The definition of state can be extended by the user (by inheriting from DecentralizedAverager)
* [x] calling `__del__` on a partially created MPFuture will no longer cause error (edge case from albert)
image

* [x] DecentralizedAverager now supports manually getting/setting current group bits (used as dht key)

### DIFF
--- a/hivemind/client/averaging/__init__.py
+++ b/hivemind/client/averaging/__init__.py
@@ -290,7 +290,7 @@ class DecentralizedAverager(mp.Process, averaging_pb2_grpc.DecentralizedAveragin
         chunk_size_bytes = self.matchmaking_kwargs.get('chunk_size_bytes', 2 ** 16)
 
         for tensor in self._averaged_tensors:
-            for part in split_for_streaming(serialize_torch_tensor(tensor), chunk_size_bytes)
+            for part in split_for_streaming(serialize_torch_tensor(tensor), chunk_size_bytes):
                 yield averaging_pb2.DownloadData(tensor_part=part)
 
     def try_load_parameters(self):
--- a/hivemind/client/averaging/__init__.py
+++ b/hivemind/client/averaging/__init__.py
@@ -350,21 +350,17 @@ class DecentralizedAverager(mp.Process, averaging_pb2_grpc.DecentralizedAveragin
 
     async def _load_state_from_peers(self, future: MPFuture):
         key_manager = self._matchmaking.group_key_manager
-        peers_info, _ = self.dht.get(f"{key_manager.prefix}.all_averagers", latest=True) or ({}, None)
-        metadata = None
-        if not isinstance(peers_info, dict):
-            logger.info("Averager could not load state from peers: peer dict is corrupted.")
+        peer_priority, _ = self.dht.get(f"{key_manager.prefix}.all_averagers", latest=True) or ({}, None)
+        peer_priority = {peer: float(info.value) for peer, info in peer_priority.items()
+                         if isinstance(info, ValueWithExpiration) and isinstance(info.value, (float, int))}
+
+        if not isinstance(peer_priority, dict) or len(peer_priority) == 0:
+            logger.info(f"Averager could not load state from peers: peer dict is absent or corrupted {peer_priority}.")
             future.set_result(None)
             return
 
-        def get_priority(peer):
-            info = peers_info[peer]
-            if not isinstance(info, ValueWithExpiration) or not isinstance(info.value, (float, int)):
-                logger.warning(f"Skipping averager {peer} - bad peer info {info.value} (expected a number).")
-                return -float('inf')
-            return float(info.value)  # prefer peers with latest update time
-
-        for peer in sorted(peers_info.keys(), key=get_priority, reverse=True):
+        metadata = None
+        for peer in sorted(peer_priority.keys(), key=peer_priority.get, reverse=True):
             if peer != self.endpoint:
                 logger.info(f"Downloading parameters from peer {peer}")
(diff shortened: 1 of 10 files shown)


hivemind/client/averaging/__init__.py
```diff
@@ -279,6 +289,134 @@ async def rpc_aggregate_part(self, stream: AsyncIterator[averaging_pb2.Averaging
...
+        chunk_size_bytes = self.matchmaking_kwargs.get('chunk_size_bytes', 2 ** 16)
+        metadata, tensors = await self._get_current_state_from_host_process()
+
+        for tensor in tensors:
+            for part in split_for_streaming(serialize_torch_tensor(tensor), chunk_size_bytes):
+                if metadata is not None:
+                    yield averaging_pb2.DownloadData(tensor_part=part, metadata=metadata)
+                    metadata = None
+                else:
+                    yield averaging_pb2.DownloadData(tensor_part=part)
```
mryab, line 313: Looks kinda hacky if your goal is to send the metadata only with the first part of the first tensor. Can't we devise a better protocol here? (for instance, send only the metadata first, then send only the tensor parts)

hivemind/client/averaging/__init__.py
```diff
@@ -279,6 +289,134 @@ async def rpc_aggregate_part(self, stream: AsyncIterator[averaging_pb2.Averaging
...
+            asyncio.create_task(asyncio.wait_for(self.dht.store(
+                download_key, subkey=self.endpoint, value=self.last_updated,
+                expiration_time=get_dht_time() + self._matchmaking.averaging_expiration, return_future=True),
+                timeout=self._matchmaking.averaging_expiration))
+            await asyncio.sleep(self._matchmaking.averaging_expiration)
+
+    async def rpc_download_state(self, request: averaging_pb2.DownloadRequest, context: grpc.ServicerContext
+                                 ) -> AsyncIterator[averaging_pb2.DownloadData]:
+        """ a newcomer requests us to send over our current trainer state for his initialization """
+        chunk_size_bytes = self.matchmaking_kwargs.get('chunk_size_bytes', 2 ** 16)
```
mryab, line 304: Move 2 ** 16 to a module-level constant

hivemind/client/averaging/__init__.py
```diff
@@ -279,6 +289,134 @@ async def rpc_aggregate_part(self, stream: AsyncIterator[averaging_pb2.Averaging
...
+            logger.info("Averager could not load state from peers: peer dict is corrupted.")
+            future.set_result(None)
+            return
+
+        def get_priority(peer):
+            info = peers_info[peer]
+            if not isinstance(info, ValueWithExpiration) or not isinstance(info.value, (float, int)):
+                logger.warning(f"Skipping averager {peer} - bad peer info {info.value} (expected a number).")
+                return -float('inf')
+            return float(info.value)  # prefer peers with latest update time
```
mryab, line 366: Can we rewrite this function not to use external variables and to move it outside the method? For example, with functools.partial, if you intend to use it as a key. It's harder to reason about closures than about non-nested functions, and in this case it's not that necessary

hivemind/client/averaging/__init__.py
```diff
@@ -279,6 +289,134 @@ async def rpc_aggregate_part(self, stream: AsyncIterator[averaging_pb2.Averaging
...
+                stream = None
+                try:
+                    leader_stub = ChannelCache.get_stub(peer, averaging_pb2_grpc.DecentralizedAveragingStub, aio=True)
+                    stream = leader_stub.rpc_download_state(averaging_pb2.DownloadRequest())
+                    current_tensor_parts, tensors = [], []
+                    async for message in stream:
+                        if message.metadata:
+                            metadata = pickle.loads(message.metadata)
+                        if message.tensor_part.dtype and current_tensor_parts:
+                            tensors.append(deserialize_torch_tensor(combine_from_streaming(current_tensor_parts)))
```
mryab, line 380: Needs a comment that `message.tensor_part.dtype` indicates the start of a new tensor

hivemind/client/averaging/__init__.py
```diff
@@ -279,6 +289,134 @@ async def rpc_aggregate_part(self, stream: AsyncIterator[averaging_pb2.Averaging
...
+                            current_tensor_parts = []
+                        current_tensor_parts.append(message.tensor_part)
+                    if current_tensor_parts:
+                        tensors.append(deserialize_torch_tensor(combine_from_streaming(current_tensor_parts)))
+                    future.set_result((metadata, tensors))
+                    self.last_updated = get_dht_time()
+                    break
```
mryab, line 387: `return` instead of `break` on success?

mryab (changes requested):
Also bump the version

hivemind/utils/mpfuture.py
```diff
@@ -161,4 +161,5 @@ def __await__(self):
 
     def __del__(self):
         self._shutdown_trigger.set_result(True)
-        self.connection.close()
+        if hasattr(self, 'connection'):
+            self.connection.close()
```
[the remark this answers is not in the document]
justheuristic: The issue arises when connection destructor is called in parallel to its creation. While this does not *normally* happen, it may sometimes occur in DHT and averager when handling unresponsive peers

mryab (approved):

Issue #155, opened 2021-02-28
Attack vector: Unpickling data from peers may lead to remote code execution


borzunov:
The PR #154 uses the `pickle` module to deserialize data from other peers. Since `pickle` allows to serialize Python code and user-defined classes with arbitrary behavior, this may lead to remote code execution ([example](https://davidhamann.de/2020/04/05/exploiting-python-pickle/)).

The issue may be fixed by switching to a safe data serializer like MessagePack.

justheuristic:
The specific line that causes the issue:
https://github.com/learning-at-home/hivemind/blob/master/hivemind/client/averaging/__init__.py#L379

justheuristic:
@borzunov Should we close this?

borzunov:
I think yes :)

PR #160, opened 2021-02-28
Remove pickle.loads in averager

justheuristic:
Addresses #155 : require pre-serialized (bytes) metadata in get_current_state, switch to MSGPackSerializer in the application code

### DIFF
--- a/hivemind/client/averaging/__init__.py
+++ b/hivemind/client/averaging/__init__.py
@@ -21,7 +21,7 @@ from hivemind.utils.grpc import ChannelCache, GRPC_KEEPALIVE_OPTIONS, \
     serialize_torch_tensor, deserialize_torch_tensor, split_for_streaming, combine_from_streaming
 from hivemind.utils.asyncio import anext, achain, aiter, switch_to_uvloop
 from hivemind.utils.timed_storage import get_dht_time, ValueWithExpiration, DHTExpiration
-from hivemind.utils.serializer import PickleSerializer, MSGPackSerializer
+from hivemind.utils.serializer import MSGPackSerializer
 from hivemind.utils import Endpoint, Port, MPFuture, get_logger
 
 # flavour types
--- a/hivemind/client/averaging/__init__.py
+++ b/hivemind/client/averaging/__init__.py
@@ -42,7 +42,7 @@ class DecentralizedAverager(mp.Process, averaging_pb2_grpc.DecentralizedAveragin
     :param dht: a DHT node that will be used to find groups
     :param start: if True, starts the background process immediately
 
-    :param prefix: a shared prefix for all group keysP
+    :param prefix: a shared prefix for all group keys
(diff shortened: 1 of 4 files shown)


hivemind/client/averaging/__init__.py
```diff
@@ -42,7 +42,7 @@ class DecentralizedAverager(mp.Process, averaging_pb2_grpc.DecentralizedAveragin
     :param dht: a DHT node that will be used to find groups
     :param start: if True, starts the background process immediately
 
-    :param prefix: a shared prefix for all group keys
+    :param prefix: a shared prefix for all group keysP
```
borzunov, line 45: Nit: accidental change?

borzunov, line 45: Nit: accidental change?

borzunov (approved):

hivemind/client/averaging/__init__.py
```diff
@@ -376,7 +384,7 @@ async def _load_state_from_peers(self, future: MPFuture):
                     current_tensor_parts, tensors = [], []
                     async for message in stream:
                         if message.metadata:
-                            metadata = PickleSerializer.loads(message.metadata)
```
mryab, line 379: PickleSerializer is no longer required, so the import can be removed

hivemind/client/averaging/__init__.py
```diff
@@ -348,8 +348,16 @@ def _background_thread_fetch_current_state_if_asked(self):
...
                 continue
 
-    def load_state_from_peers(self, wait=True) -> Optional[Any]:
-        """ Try to download the latest optimizer state one of the existing peer """
+    def load_state_from_peers(self, wait=True) -> Optional[Tuple[bytes, Sequence[torch.Tensor]]]:
+        """
+        Try to download the latest optimizer state one of the existing peer.
+        :returns: on success, return a 2-tuple with (serialized_metadata, tensors), where
+
+        - serialized_metadata is a small bytestring containing **serialized** metadata (e.g. hyperparameters)
```
mryab, line 356: Shouldn't we also deserialize on return? What are the cases where we still need serialized metadata as the output of this method?

mryab (changes requested):

hivemind/client/averaging/__init__.py
justheuristic: As of now, this is left to the user (to provide choose a secure pickler).

tests/test_averaging.py
```diff
@@ -277,7 +278,7 @@ def get_current_state(self):
             """
             nonlocal num_calls, super_metadata, super_tensors
             num_calls += 1
-            return super_metadata, super_tensors
+            return serializer.dumps(super_metadata), super_tensors
```
mryab, line 281: I believe you can use self.serializer here and avoid declaring the serializer at L271?

mryab (commented):
Also bump the version, please

mryab (approved):

PR #161, opened 2021-02-28
feat: add p2p daemon

skobellev:
Add P2P class to manage p2p daemon and ensure it is terminated correctly.


justheuristic:
[note: this will become a root branch for P2Pd functionality]

PR #162, opened 2021-02-28
copytree implementation for py37 compatibility

leshanbog:
Google Colab has python3.7 version.
In python3.7 shutil.copytree has no dirs_exist_ok flag (https://docs.python.org/3/library/shutil.html#shutil.copytree). Here is a python3.7 compatible implementation. Tests for this version are also added to CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,26 @@
 version: 2.1
 
 jobs:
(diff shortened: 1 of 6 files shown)


justheuristic (approved):
Could you please also increment version in hivemind/__init__.py (e.g. 0.9.2)

PR #164, opened 2021-02-28
feat: add p2p daemon

skobellev:
Add P2P class to manage p2p daemon and ensure it is terminated correctly.

### DIFF
--- a/hivemind/p2p/p2p_daemon.py
+++ b/hivemind/p2p/p2p_daemon.py
@@ -1,0 +1,45 @@
+import subprocess
+import typing as tp
+
+
+class P2P(object):
+    """
+    Forks a child process and executes p2pd command with given arguments.
+    Sends SIGKILL to the child in destructor and on exit from contextmanager.
+    """
+
+    LIBP2P_CMD = 'p2pd'
+
+    def __init__(self, *args, **kwargs):
+        self._child = subprocess.Popen(args=self._make_process_args(args, kwargs))
+        try:
+            stdout, stderr = self._child.communicate(timeout=0.2)
+        except subprocess.TimeoutExpired:
+            pass
+        else:
+            raise RuntimeError(f'p2p daemon exited with stderr: {stderr}')
+
+    def __enter__(self):
+        return self._child
+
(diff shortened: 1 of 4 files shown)


hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,45 @@
...
+            raise RuntimeError(f'p2p daemon exited with stderr: {stderr}')
+
+    def __enter__(self):
+        return self._child
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._kill_child()
+
+    def __del__(self):
+        self._kill_child()
```
justheuristic, line 29: [Nitpicking] This will handle del/dereferencing, but not sigkill. Is there perhaps a way to create a non-unix-daemon (aka python daemon) that is guaranteed to die with the host process?

justheuristic (changes requested):
Mostly great. The only request is to move p2p.py into hivemind/dht. The rest is "nice to have", but we can handle that later.

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,45 @@
...
+import typing as tp
+
+
+class P2P(object):
+    """
+    Forks a child process and executes p2pd command with given arguments.
+    Sends SIGKILL to the child in destructor and on exit from contextmanager.
+    """
+
+    LIBP2P_CMD = 'p2pd'
```
justheuristic, line 11: @deniskamazur sanity check: can we guarantee that the p2pd you compile in #153 will be always accessible as P2Pd?

deniskamazur: Yes, as long as the `main` package stays in the p2pd directory.
That's what `go help install` says:
`Install compiles and installs the packages named by the import paths...```

deniskamazur: It actually might be a good idea to download a specific version of libp2p-daemon rather than the latest version of the master branch to make sure refactorings of their codebase don't break our builds.

justheuristic (approved):
upd - let's merge it as is, the rest can be fixed in subsequent PRs

PR #166, opened 2021-03-01
Fix the remaining tests for py37

justheuristic:
* [x] DecentralizedAverager is now compatible with python37's acyncio exception
    * the problem was: grpc.aio with python37 raised concurrent.futures.CancelledError in some cases;
    * we relied on isinstance(asyncio.CancelledError, Exception) == False
    * but isinstance(concurrent.futures.CancelledError, Exception) == True
* [x] DecentralizedAverager now shuts down if dereferenced in the main process
    * [x] though it won't shutdown if dereferenced in forks for obvious reasons
* [x] HIVEMIND_THREADS now actually works
* [x] test_averaging now shuts down dht and averager instances to avoid leaking processes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,5 +65,6 @@ jobs:
 workflows:
   main:
     jobs:
+      - build-and-test-py37
       - build-and-test-py38
       - build-and-test-py39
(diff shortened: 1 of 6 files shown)


mryab (changes requested):

mryab (approved):
Version bump?

leshanbog (approved):

PR #171, opened 2021-03-04
Handle edge cases in DecentralizedAverager

justheuristic:
__Method:__ we ran ALBERT training with 5 peers and shut down a random process _during_ averager.step

__Expected behavior:__ surviving averagers should always retry if error was caused by another peer
image

Found and fixed two cases that did not trigger retries in averager.step():
* if an averager perishes while running AllReduce, it triggers one of 3 gRPC errors in groupmates
* if an averager receives incomplete stream of data in rpc_aggregate_part, it causes RuntimeError in deserialization

Also:
* fixed typing and docstrings left over from #168 
* added tests for spit_for_streaming / combine_from_streaming, including corrupted streams

Note: the screenshot above contains improperly formatted logs (see {e}), this is now fixed.

### DIFF
--- a/hivemind/client/averaging/allreduce.py
+++ b/hivemind/client/averaging/allreduce.py
@@ -153,7 +153,12 @@ class AllReduceRunner(AllReduceProtocol, averaging_pb2_grpc.DecentralizedAveragi
                                      f" instead of {averaging_pb2.MessageCode.Name(averaging_pb2.AVERAGED_PART)},"
                                      f" allreduce failed")
 
-        averaged_part = deserialize_torch_tensor(combine_from_streaming([message.tensor_part for message in outputs]))
+        try:
+            averaged_part = deserialize_torch_tensor(combine_from_streaming(
+                [message.tensor_part for message in outputs]))
+        except RuntimeError as e:
+            raise AllreduceException(f"Could not deserialize averaged part from {peer_endpoint}: {e}")
+
         self.register_averaged_part(peer_endpoint, averaged_part)
         return averaged_part
 
@@ -182,7 +187,11 @@ class AllReduceRunner(AllReduceProtocol, averaging_pb2_grpc.DecentralizedAveragi
     async def accumulate_part_streaming(self, source: Endpoint, stream_messages: Iterable[runtime_pb2.Tensor]
                                         ) -> Iterable[runtime_pb2.Tensor]:
         """ accumulate_part using streams of serialized tensors. Used to prevent duplicate work in serialization """
-        tensor_part: torch.Tensor = deserialize_torch_tensor(combine_from_streaming(stream_messages))
+        try:
+            tensor_part: torch.Tensor = deserialize_torch_tensor(combine_from_streaming(stream_messages))
+        except RuntimeError as e:
+            raise AllreduceException(f"Could not deserialize tensor part from {source} for streaming {e}")
+
         averaged_part = await self.accumulate_part(source, tensor_part)
         if not self.averaged_part_stream.done():
             serialized_tensor = serialize_torch_tensor(averaged_part, self.compression_type, allow_inplace=False)
--- a/hivemind/client/averaging/allreduce.py
+++ b/hivemind/client/averaging/allreduce.py
@@ -188,7 +188,7 @@ class AllReduceRunner(AllReduceProtocol, averaging_pb2_grpc.DecentralizedAveragi
                                         ) -> Iterable[runtime_pb2.Tensor]:
         """ accumulate_part using streams of serialized tensors. Used to prevent duplicate work in serialization """
         try:
-            tensor_part: torch.Tensor = deserialize_torch_tensor(combine_from_streaming(stream_messages))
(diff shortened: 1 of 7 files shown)


hivemind/client/averaging/allreduce.py
```diff
@@ -182,7 +187,11 @@ async def run(self) -> Sequence[torch.Tensor]:
     async def accumulate_part_streaming(self, source: Endpoint, stream_messages: Iterable[runtime_pb2.Tensor]
                                         ) -> Iterable[runtime_pb2.Tensor]:
         """ accumulate_part using streams of serialized tensors. Used to prevent duplicate work in serialization """
-        tensor_part: torch.Tensor = deserialize_torch_tensor(combine_from_streaming(stream_messages))
+        try:
+            tensor_part: torch.Tensor = deserialize_torch_tensor(combine_from_streaming(stream_messages))
```
mryab, line 191: Redundant type annotation

tests/test_util_modules.py
```diff
@@ -192,6 +192,42 @@ def test_serialize_tensor():
     restored = hivemind.combine_from_streaming(chunks)
     assert torch.allclose(hivemind.deserialize_torch_tensor(restored), tensor)
 
+
+def test_split_parts():
+    tensor = torch.randn(910, 512)
+    serialized_tensor_part = hivemind.utils.serialize_torch_tensor(tensor, allow_inplace=False)
+    chunks1 = list(hivemind.utils.split_for_streaming(serialized_tensor_part, 16384))
+    assert len(chunks1) == 114
```
mryab, line 200: We'd better explicitly compute these magic numbers
Also: L203, L211

hivemind/client/averaging/allreduce.py
```diff
@@ -182,7 +187,11 @@ async def run(self) -> Sequence[torch.Tensor]:
...
+        except RuntimeError as e:
+            raise AllreduceException(f"Could not deserialize tensor part from {source} for streaming {e}")
```
mryab, line 193: Ideally, we should narrow down the scope of this exception:
1) Do we want to catch errors that happen only because of streaming? If so, enclose `combine_from_streaming` inside the try-except block
2) Do we want to catch any kind of specifically deserialization errors? If so, move this check inside `deserialize_torch_tensor`. If you still want an AllreduceException here, you can then catch a specific SerializerException

mryab (changes requested):

hivemind/client/averaging/allreduce.py
justheuristic: The problem is that errors in combine_from_streaming only manifest in deserialize_torch_tensor .

tests/test_util_modules.py
```diff
@@ -192,6 +193,42 @@ def test_serialize_tensor():
     restored = hivemind.combine_from_streaming(chunks)
     assert torch.allclose(hivemind.deserialize_torch_tensor(restored), tensor)
 
+
+def test_split_parts():
+    tensor = torch.randn(910, 512)
+    serialized_tensor_part = hivemind.utils.serialize_torch_tensor(tensor, allow_inplace=False)
+    chunks1 = list(hivemind.utils.split_for_streaming(serialized_tensor_part, 16384))
+    assert len(chunks1) == int(np.ceil(tensor.numel() * 4 / 16384))
```
mryab, line 201:
```suggestion
    assert len(chunks1) == int(np.ceil(tensor.numel() * tensor.element_size() / 16384))
```

tests/test_util_modules.py
```diff
@@ -192,6 +193,42 @@ def test_serialize_tensor():
...
+
+    chunks2 = list(hivemind.utils.split_for_streaming(serialized_tensor_part, 10_000))
+    assert len(chunks2) == int(np.ceil(tensor.numel() * 4 / 10_000))
```
mryab, line 204:
```suggestion
    assert len(chunks2) == int(np.ceil(tensor.numel() * tensor.element_size() / 10_000))
```

tests/test_util_modules.py
```diff
@@ -192,6 +193,42 @@ def test_serialize_tensor():
...
+        assert torch.allclose(tensor, hivemind.deserialize_torch_tensor(combined), rtol=1e-5, atol=1e-8)
+
+    assert torch.allclose(tensor, hivemind.deserialize_torch_tensor(combined4), rtol=1e-3, atol=1e-3)
+
+    combined_incomplete = hivemind.utils.combine_from_streaming(chunks4[:5])
+    combined_incomplete2 = hivemind.utils.combine_from_streaming(chunks4[:1])
+    combined_incomplete3 = hivemind.utils.combine_from_streaming(chunks4[:-1])
+    for combined in combined_incomplete, combined_incomplete2, combined_incomplete3:
+        with pytest.raises(RuntimeError):
+            _ = hivemind.deserialize_torch_tensor(combined)
```
mryab, line 228: Is assignment necessary?

mryab (approved):

PR #172, opened 2021-03-04
Serialize DHTID source with msgpack

leshanbog:
Using msgpack to create bytes from DHT keys to avoid pickle compatibility issue.

### DIFF
--- a/hivemind/utils/serializer.py
+++ b/hivemind/utils/serializer.py
@@ -1,5 +1,4 @@
 """ A unified interface for several common serialization methods """
-import pickle
 from io import BytesIO
 from typing import Dict, Any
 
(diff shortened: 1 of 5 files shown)


justheuristic:
@leshanbog
note: you could add a special serialization rule for python tuples, similarly to this syntax:
https://github.com/learning-at-home/hivemind/blob/master/hivemind/dht/storage.py#L9-L10

justheuristic (approved):

tests/test_routing.py
```diff
@@ -5,7 +5,7 @@
 
 from hivemind import LOCALHOST
 from hivemind.dht.routing import RoutingTable, DHTID
-from hivemind.utils.serializer import PickleSerializer
+from hivemind.utils.serializer import MSGPackSerializer
```
mryab, line 8: Import not required

hivemind/utils/serializer.py
```diff
@@ -64,12 +42,18 @@ def _encode_ext_types(cls, obj):
         type_code = cls._ExtTypes.get(type(obj))
         if type_code is not None:
             return msgpack.ExtType(type_code, obj.packb())
+        elif isinstance(obj, tuple):
+            data = msgpack.packb(list(obj), strict_types=True, use_bin_type=True, default=cls._encode_ext_types)
+            return msgpack.ExtType(cls._MsgpackExtTypeCodeTuple, data)
```
mryab, line 47: We'd better leave a comment to explain why tuples need to be handled separately

mryab (changes requested):

mryab (approved):

Issue #159, opened 2021-02-28
hivemind-server unavailable when installing from PyPI


mryab:
**Describe the bug**
When installing hivemind from PyPI, we can't use the `hivemind-server` CLI command, as it tries to import from the folder `scripts`:

```
$ hivemind-server
Traceback (most recent call last):
  File "/home/mryab/miniconda3/envs/hivemind_env_test/bin/hivemind-server", line 5, in <module>
    from scripts.run_server import main
ModuleNotFoundError: No module named 'scripts'
```

We should probably change the directory name the way it's done in [transformers](https://github.com/huggingface/transformers/blob/master/setup.py#L297) or [fairseq](https://github.com/pytorch/fairseq/blob/master/setup.py#L219-L226).

**To Reproduce**
Install hivemind from PyPI and run `hivemind-server`

**Environment**
* Python 3.8.8
* hivemind 0.9.0

PR #173, opened 2021-03-04
Move CLI server launch script to hivemind/hivemind_cli

mryab:
As discussed in the issue linked below, currently the server script doesn't launch when installed from PyPI or in a non-editable mode. To fix this, I moved the CLI script inside the hivemind source directory the same way it's done in other libraries.

Fixes https://github.com/learning-at-home/hivemind/issues/159.


justheuristic:
Everything appears to be in perfect order

justheuristic (approved):

PR #176, opened 2021-03-09
Support client-only participants in AllReduceProtocol

### LINKED ISSUE #147: Support client-only participants in AllReduceProtocol
Here's [AllReduceProtocol's inner logic](https://github.com/learning-at-home/hivemind/blob/master/tests/test_averaging.py#L139-L172): . It implements Butterfly All-Reduce

Note the `part_sizes` - these are fractions of the averaged vector assigned to each respective peer.

__Request:__

* [ ] if support averaging with client-mode peers:
 * Client-mode peer always has `part_sizes[i]==0`
 * Such peers can call aggregate_part from others, but **no one should call aggregate_part on the client-mode peer**

* [ ] implement a basic test case similar to [this](https://github.com/learning-at-home/hivemind/blob/master/tests/test_averaging.py#L139-L172), but with 1-2 client-only peers

* [ ] update AllreduceRunner to work with the updated AllreduceProtocol

foksly:
Resolves #147


tests/test_averaging.py
```diff
@@ -150,15 +150,15 @@ def test_allgather():
 @pytest.mark.asyncio
 async def test_allreduce_protocol():
```
foksly, line 151: I decided not to make a separate test since there will be a lot of copy paste. Is it ok or there are some arguments for making a separate one?

hivemind/client/averaging/allreduce.py
```diff
@@ -30,6 +30,7 @@ def __init__(self, *, group_id: GroupID, tensors: Sequence[torch.Tensor], endpoi
         assert endpoint in ordered_group_endpoints, "endpoint is not a part of the group"
         self.group_id, self.endpoint = group_id, endpoint
```
foksly, line 31: What if we specify client-only participants in separate parameter `client_mode_endpoints` hence we have no need in `part_sizes[i]==0`?

hivemind/client/averaging/allreduce.py
```diff
@@ -59,6 +62,7 @@ async def accumulate_part(self, source: Endpoint, remote_part: torch.Tensor) ->
         assert not self.future.done(), f"already finished allreduce: {self.future}"
         assert source in self.local_tensor_parts, "unexpected source, not a part of current group"
         assert source not in self.accumulated_from, "duplicate source, already received that part"
+        assert not self.endpoint in self.client_mode_endpoints, f"{self.endpoint} is in client mode"
```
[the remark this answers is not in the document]
foksly: If we want to disallow calling `accumulate_part` from a client-only participants should I add a separate test checking a failure?

hivemind/client/averaging/allreduce.py
justheuristic: Yep, that would make sense. But let us not demolish the test for part_sizes[i]==0 - it is an edge case  that may one day help us find a bigger bug.

hivemind/client/averaging/allreduce.py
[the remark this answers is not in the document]
justheuristic: @AllreduceException - In future it would be great to introduce a separate exception for this. As of now, however, the AllreduceException is used for errors that are "normal", e.g. caused by a network failure.
In turn, errors that indicate hivemind implementation problems are raised as AssertionError/ValueError

@separate test - that would be great, but not absolutely required

tests/test_averaging.py
justheuristic: I am okay with whichever version you prefer here.
@mryab , do you have any preferences?

tests/test_averaging.py
```diff
@@ -120,7 +120,7 @@ def test_allreduce_grid():
 @pytest.mark.forked
 def test_allgather():
     dht = hivemind.DHT(start=True, endpoint=f'{hivemind.LOCALHOST}:*')
-    averagers = [hivemind.DecentralizedAverager(torch.ones(1), dht=dht, target_group_size=4, averaging_expiration=15,
+    averagers = [hivemind.DecentralizedAverager([torch.ones(100)], dht=dht, target_group_size=4, averaging_expiration=15,
```
foksly, line 123: fix back

hivemind/client/averaging/allreduce.py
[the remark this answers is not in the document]
mryab: Shouldn't accumulate_part be simply a no-op for client-only participants? I guess the cleanest way to implement averaging would be to have the same execution flow for all workers regardless of their participation status. 

So ideally if you want to join the training process with zero bandwidth, all you have to do is set your bandwidth to 0 — the code remains the same. We may short-circuit that and avoid calling unnecessary functions under the hood, but from the user point of view the API has to be the same

tests/test_averaging.py
mryab: One test is fine, I guess (as long as we check that the colab peer gets the average value)

mryab (approved):

hivemind/client/averaging/allreduce.py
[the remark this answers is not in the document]
justheuristic: [based on our previous conversation]
Averager syntax should definitely be the same for the user, but accumulate_part is not exposed to the user and calling it will incur at least 1 RTT delay.

As a recommendation - let's keep the assertion in accumulate_part and make sure that the rest of our DecentralizedAverager respects it.

hivemind/client/averaging/matchmaking.py
```diff
@@ -165,7 +166,7 @@ async def request_join_group(self, leader: Endpoint, expiration_time: DHTExpirat
                 leader_stub = ChannelCache.get_stub(leader, averaging_pb2_grpc.DecentralizedAveragingStub, aio=True)
                 call = leader_stub.rpc_join_group(averaging_pb2.JoinRequest(
                     endpoint=self.endpoint, schema_hash=self.schema_hash, expiration=expiration_time,
-                    throughput=self.throughput if self.throughput is not None else -1.0,
+                    throughput=0.0 if self.client_mode else (self.throughput if self.throughput is not None else -1.0),
```
mryab, line 169: The triple if condition seems too complex (especially for a single line), can't we break this down or reduce the number of cases?

hivemind/client/averaging/__init__.py
```diff
@@ -157,18 +159,24 @@ def run(self):
...
+            if self.listen:
+                server = grpc.aio.server(**self.kwargs, options=GRPC_KEEPALIVE_OPTIONS)
+                averaging_pb2_grpc.add_DecentralizedAveragingServicer_to_server(self, server)
+                found_port = server.add_insecure_port(self.listen_on)
+                assert found_port != 0, f"Failed to listen to {self.listen_on}"
+                self._port.value = found_port
+                await server.start()
+                asyncio.create_task(self._declare_for_download_periodically())
+            else:
+                logger.info(f"[experimental] The averager running in client mode, please report any bugs.")
```
mryab, line 172:
```suggestion
                logger.info(f"The averager running in an experimental client mode, please report any bugs.")
```

mryab (changes requested):

tests/test_averaging.py
```diff
@@ -117,10 +117,29 @@ def test_allreduce_grid():
     dht.shutdown()
 
 
+@pytest.mark.forked
+def test_allreduce_client_ugly_please_replace():
```
mryab, line 121: Change the name with something more adequate (better yet, make sure that the test is complete before merging)

justheuristic:
based on @mryab comments:
- it would be great to create uuid once at __init__ instead of in @property endpoint
- it would be great to change Endpoint for listen=False peers to return None and manually switch to uuid

justheuristic (approved):
please bump version as well

mryab (approved):
Bump the minor version and rebase on master

PR #182, opened 2021-03-14
Add quantile compression

Vsevolod-pl:


hivemind/utils/grpc.py
```diff
@@ -160,6 +161,45 @@ def top(self) -> ValueError:
         raise ValueError(f"Please use {self.__class__.__name__}.get_stub to get or create stubs")
 
 
+def quantile_encode_torch_qq(weight, n_bits: int):
```
justheuristic, line 164: probably better to rename it into something like "quantile_encode_torch_approx"

hivemind/utils/grpc.py
```diff
@@ -160,6 +161,45 @@ def top(self) -> ValueError:
...
+    n_bins = 2 ** n_bits
+    borders = torch.as_tensor(quantile_qq_approximation(weight.numpy(), n_bins + 1)[:-1])
+    quant_weight = torch.relu_(torch.bucketize(weight, borders) - 1)
+    _sums = torch.zeros(n_bins).scatter_add_(0, quant_weight.flatten(), weight.flatten())
+    lookup = _sums.mul_(n_bins / weight.numel())
```
justheuristic, line 169: note: it may be better to divide not by weight.numel(), but by the actual number of values in each bin

hivemind/utils/grpc.py
```diff
@@ -207,6 +247,18 @@ def serialize_torch_tensor(tensor: torch.Tensor, compression_type=CompressionTyp
...
+        assert tensor.dtype == torch.float32
+
+        quantized, lookup = quantile_encode_torch_qq(tensor.detach(), 8)
+        data = b''.join((lookup.numpy().tobytes(), quantized.numpy().astype(np.uint8).tobytes()))
+
+        proto = runtime_pb2.Tensor(
+            compression=compression_type,
+            buffer=data,
+            size=tensor.shape,
+            dtype='compressed_float32',
```
justheuristic, line 260: Please rename dtype

justheuristic (approved):
[all comments were resolved]

tests/test_util_modules.py
```diff
@@ -131,6 +131,8 @@ def test_vector_compression(size=(128, 128, 64), alpha=5e-08):
     assert error.square().mean() < alpha
     error = deserialize_torch_tensor(serialize_torch_tensor(X, CompressionType.FLOAT16)) - X
     assert error.square().mean() < alpha
+    error = deserialize_torch_tensor(serialize_torch_tensor(X, CompressionType.QUANTILE_8BIT)) - X
+    assert error.square().mean() < 0.0008
```
mryab, line 135: We'd better parametrize this tolerance by a constant (by the way, why not torch.allclose here?)

hivemind/utils/grpc.py
```diff
@@ -10,6 +10,7 @@
 import grpc
 import numpy as np
 import torch
+import hivemind
```
mryab, line 13: Better to use more specific imports, especially when only one function is used. Also, local imports need to be separated by a newline

hivemind/utils/grpc.py
```diff
@@ -207,6 +248,18 @@ def serialize_torch_tensor(tensor: torch.Tensor, compression_type=CompressionTyp
             size=array.shape,
             dtype=array.dtype.name,
             requires_grad=tensor.requires_grad)
+    elif compression_type == CompressionType.QUANTILE_8BIT:
+        assert tensor.dtype == torch.float32
+
+        quantized, lookup = quantile_encode_torch_approx(tensor.detach(), 8)
+        data = b''.join((lookup.numpy().tobytes(), quantized.numpy().astype(np.uint8).tobytes()))
```
mryab, line 255: Casting to ByteTensor before converting to np.array is probably more efficient

hivemind/utils/grpc.py
```diff
@@ -231,6 +284,13 @@ def deserialize_torch_tensor(serialized_tensor: runtime_pb2.Tensor) -> torch.Ten
     elif serialized_tensor.compression == CompressionType.FLOAT16:
         array = np.frombuffer(serialized_tensor.buffer, dtype=np.float16).copy()
         tensor = torch.as_tensor(array, dtype=torch.float32).view(*serialized_tensor.size)
+    elif serialized_tensor.compression == CompressionType.QUANTILE_8BIT:
+        lookup = serialized_tensor.buffer[:256*4]
```
mryab, line 288: We'd better extract this value of 256*4 to a variable or a constant, because right now its meaning is unclear

hivemind/utils/grpc.py
```diff
@@ -160,6 +161,46 @@ def top(self) -> ValueError:
         raise ValueError(f"Please use {self.__class__.__name__}.get_stub to get or create stubs")
 
 
+def quantile_encode_torch_approx(weight, n_bits: int):
```
mryab, line 164: What is the return type of this function? Also, it's probably best to rename the first argument to something more general (like `tensor`)

hivemind/utils/grpc.py
```diff
@@ -160,6 +161,46 @@ def top(self) -> ValueError:
...
+    borders = torch.as_tensor(quantile_qq_approximation(weight.numpy(), n_bins + 1)[1:-1])
+    quant_weight = torch.clamp_(torch.bucketize(weight, borders), 0, n_bins - 1)
+    bin_sums = torch.zeros(n_bins).scatter_add_(0, quant_weight.flatten(), weight.flatten())
+    bin_counts = torch.clamp_min_(torch.bincount(quant_weight.flatten(), minlength=n_bins), 1)
+    lookup = bin_sums / bin_counts
+    return quant_weight, lookup
+
+
+def quantile_qq_approximation(array, n_quantiles, min_chunk_size: int = 10 ** 5, **kwargs):
+    """ Estimate uniform quantiles of data using quantile-of-quantiles. Runs in parallel. """
```
mryab, line 175: Might be just my ignorance, but what are _uniform_ quantiles and how are they different from regular quantiles?

hivemind/utils/grpc.py
```diff
@@ -160,6 +161,46 @@ def top(self) -> ValueError:
...
+    n_bins = 2 ** n_bits
+    borders = torch.as_tensor(quantile_qq_approximation(weight.numpy(), n_bins + 1)[1:-1])
+    quant_weight = torch.clamp_(torch.bucketize(weight, borders), 0, n_bins - 1)
+    bin_sums = torch.zeros(n_bins).scatter_add_(0, quant_weight.flatten(), weight.flatten())
+    bin_counts = torch.clamp_min_(torch.bincount(quant_weight.flatten(), minlength=n_bins), 1)
+    lookup = bin_sums / bin_counts
+    return quant_weight, lookup
+
+
+def quantile_qq_approximation(array, n_quantiles, min_chunk_size: int = 10 ** 5, **kwargs):
```
mryab, line 174: Do we want to leave the possibility to pass arbitrary kwargs to run_in_background? Right now it seems that this isn't used by quantile_encode_torch_approx

mryab (changes requested):
1. I'd like to see the benchmarks for this compression method (at least on synthetic data) and comparison with float16/meanstd compression in terms of speed and round-trip 
2. Right now, it seems that we have two different types of functionality in utils/grpc.py: related to compression and serialization and related to gRPC (caching etc.). What do you think about extracting the compression/serialization-related code into a separate module?

hivemind/utils/grpc.py
justheuristic: it is not, .numpy() is a no-op for cpu tensors

hivemind/utils/grpc.py
justheuristic: something like QUANTILE_8BIT_LOOKUP_SIZE?

hivemind/utils/grpc.py
justheuristic: this technique specifically refers to estimating n+1 quantiles in a uniform symmetric grid in [0, 1/n, 2/n, ..., n/n]

hivemind/utils/grpc.py
justheuristic: I would argue that this should not be decided in this PR. Please add it to #98

justheuristic:
> 1. I'd like to see the benchmarks for this compression method (at least on synthetic data) and comparison with float16/meanstd compression in terms of speed and round-trip
> 2. Right now, it seems that we have two different types of functionality in utils/grpc.py: related to compression and serialization and related to gRPC (caching etc.). What do you think about extracting the compression/serialization-related code into a separate module?

1. agreed, and let's discuss it asap
2. agreed, but let's delay this to #98 to avoid merge conflicts with @mponty

hivemind/utils/grpc.py
mryab: Not sure why this shouldn't be decided in this PR, because this function is introduced specifically for quantile compression

hivemind/utils/grpc.py
```diff
@@ -30,6 +32,11 @@
     ('grpc.http2.min_ping_interval_without_data_ms', 10 * 1000),
 )
 
+SIZE_OF_FLOAT32 = 4
+SIZE_OF_FLOAT16 = 2
+NUM_BITS_QUANTILE_COMPRESSION = 8
```
mryab, line 37:
```suggestion
NUM_BYTES_FLOAT32 = 4
NUM_BYTES_FLOAT16 = 2
NUM_BITS_QUANTILE_COMPRESSION = 8
```

hivemind/utils/grpc.py
```diff
@@ -30,6 +32,11 @@
...
+NUM_QUANTILES_QUANTILE_COMPRESSION = 2 ** NUM_BITS_QUANTILE_COMPRESSION
```
mryab, line 38:
```suggestion
NUM_COMPRESSION_QUANTILES = 2 ** NUM_BITS_QUANTILE_COMPRESSION
```

hivemind/utils/grpc.py
```diff
@@ -160,6 +167,46 @@ def top(self) -> ValueError:
         raise ValueError(f"Please use {self.__class__.__name__}.get_stub to get or create stubs")
 
 
+def quantile_encode_torch_approx(tensor: torch.Tensor, n_bits: int) -> Tuple[torch.Tensor, torch.Tensor]:
```
mryab, line 170:
```suggestion
def quantile_encode_approx(tensor: torch.Tensor, n_bits: int) -> Tuple[torch.Tensor, torch.Tensor]:
```

hivemind/utils/grpc.py
```diff
@@ -207,6 +254,18 @@ def serialize_torch_tensor(tensor: torch.Tensor, compression_type=CompressionTyp
             size=array.shape,
             dtype=array.dtype.name,
             requires_grad=tensor.requires_grad)
+    elif compression_type == CompressionType.QUANTILE_8BIT:
```
mryab, line 257: The compression type is named as 8BIT, but in fact we can easily change the number of bits. Shouldn't we make the name more general?

mryab (changes requested):

hivemind/utils/grpc.py
justheuristic: As a counterargument: this specific compression is 8-bit and our protocol does not support parameterized compressions yet.

tests/test_util_modules.py
justheuristic: @allclose @mryab 
Thing is, allclose measures a significantly different statistic. In this test, we check for mean squared deviation, while allclose would react to the __worst__ case, even if it is just one number in a million.

mryab (approved):

Vsevolod-pl:
Benchmark results:
Compression type: NONE, time: 0.9539689302444458
Compression type: MEANSTD_LAST_AXIS_FLOAT16, time: 0.063877534866333
Compression type: FLOAT16, time: 0.030304114023844402
Compression type: QUANTILE_8BIT, time: 0.18161289691925048

PR #186, opened 2021-03-17
Fix LibP2P-Daemon installation in setup.py

deniskamazur:
* Rename `ProtoCompileInstall` and `ProtoCompileDevelop` to `Install` and `Develop`
* Install LibP2P-Daemon on setup install and setup develop
* Install Golang in Circle CI builds
* Add p2pd binary to gitignore

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,20 +7,17 @@ import urllib.request
 import tarfile
 import tempfile
 import hashlib
-import logging
 
 from packaging import version
 from pkg_resources import parse_requirements
 from setuptools import setup, find_packages
 from setuptools.command.develop import develop
 from setuptools.command.install import install
-from hivemind.utils.logging import get_logger
 
 P2PD_VERSION = 'v0.3.1'
 P2PD_CHECKSUM = '5094d094740f4e375afe80a5683b1bb2'
 
 here = os.path.abspath(os.path.dirname(__file__))
-logger = get_logger(__name__)
 
 
 def md5(fname, chunk_size=4096):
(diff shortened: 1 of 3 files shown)


justheuristic (approved):
Good point, let's use binaries in install mode, we just need to figure out how best to store them

justheuristic:
:note from @mryab :
- check version when downloading (i.e. not just latest)
- do not re-download p2pd if control sum matches the reference

setup.py
```diff
@@ -49,46 +50,54 @@ def proto_compile(output_path):
...
-            status = os.system(f'go build -o {os.path.join(here, "hivemind/hivemind_cli", "p2pd")}')
-            if status:
-                raise RuntimeError('Failed to build or install libp2p-daemon:'\
-                                   f' exited with status code :{status}')
-
-
-class ProtoCompileInstall(install):
+def install_libp2p_daemon(build=False):
+    # verify golang installation if build flag is specified
+    if build:
```
mryab, line 55: In essence, you have two functions with entirely different code. It's best to separate them into two functions with different names

mryab (changes requested):

setup.py
```diff
@@ -30,6 +35,14 @@ def __exit__(self, etype, value, traceback):
         os.chdir(self.savedPath)
 
 
+def md5(fname):
+    hash_md5 = hashlib.md5()
+    with open(fname, "rb") as f:
+        for chunk in iter(lambda: f.read(4096), b""):
```
mryab, line 41: Move 4096 to a constant or at least an argument with default value

mryab (changes requested):

mryab (approved):

setup.py
```diff
@@ -67,38 +79,44 @@ def install_libp2p_daemon():
...
+                raise RuntimeError('Failed to build or install libp2p-daemon:'
                                    f' exited with status code :{status}')
 
 
-class ProtoCompileInstall(install):
+def libp2p_download_install():
+    install_path = os.path.join(here, 'hivemind/hivemind_cli/')
+    binary_path = os.path.join(install_path, 'p2pd')
+    if 'p2pd' not in os.listdir(install_path) or md5(binary_path) != P2PD_CHECKSUM:
+        print('Downloading Peer to Peer Daemon')
```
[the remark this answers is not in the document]
mryab: Will fix that when merging to master

Issue #93, opened 2020-08-29
Discussion: p2p security


justheuristic:
Right now, hivemind works with default tcp protocol with no security. If we are to run on a global scheme, we need to:

__1. make sure peers do not risk their personal data by running hivemind nodes__
This is the boring, but necessary one: we need to make sure we use up-to-date security protocols in both hivemind.dht and hivemind.client/server interaction

__2. figure out how to make hivemind __model__ resistant to malicious peers__
This part is more tricky, there are several attack vectors:
2a. send NaN or wrong gradients to the expert in order to jeopardize its parameters
2b. overwrite existing keys in DHT with wrong information to prevent other peers from finding experts

There are intuitive ways to resist both of these vectors by e.g. dropping outlier gradients in 2a, but security usually works better with expertise, not intuition.

We would appreciate any advice on potential attack vectors and ways to mitigate them.

borzunov:
We have outlined a draft with a number of attack vectors for hivemind and possible defense strategies. This list is not guaranteed to be complete.

## 1. Remote Code Execution vulnerabilities

These are arguably the most critical vulnerabilities, since an attacker is able not only to interfere with an ongoing deep learning experiment, but to take control of participant's computers (e.g. turn them into botnet nodes). This makes participation in an open experiment unsafe.

### a. Using unsafe serializers (#155)

Deserializers like `pickle` and everything that built on top of it (like `torch.load`) can load and execute arbitrary attacker's code ([example](https://davidhamann.de/2020/04/05/exploiting-python-pickle/)).

__Possible solution:__

- Use restricted and safe serialization protocols like MessagePack.

## 2. DHT vulnerabilities

### a. Attacks on DHT operation

Attacker's nodes may send fake contents, omit writings, etc.

__Possible solution:__

- Since DHT is a well-known concept used in various open p2p protocols (like BitTorrent), it is likely that there are standard techniques to make DHT more robust to such attacks (e.g. detect and ban malicious nodes), that we need to study and implement.

### b. Saving records with fake user ID

An attacker may rewrite DHT records corresponding to other node user IDs (e.g. to ruin their statistics or results).

__Possible solution:__

- We can turn the user IDs to RSA open keys, then require users to sign any DHT request that includes a user ID with their closed key, so an attacker won't be able to fake the ID.
- To protect from an attacker replaying previous requests of another node, we should ensure that all writes with the outdated or equal expiration time are ignored.

### c. Saving records with invalid expiration time

[...]

PR #187, opened 2021-03-19
Add RSA signature protection for DHT records

borzunov:
This PR introduces a notion of **protected DHT records** whose key/subkey contains substring `[owner:ssh-rsa ...]` (the format can be changed) with an RSA public key of the owner. Changes to such records always must be signed with the corresponding private key (so only the owner can change them). This protects from malicious nodes trying to spoil the DHT contents (see [the list of attack vectors](https://github.com/learning-at-home/hivemind/issues/93#issuecomment-787993481)).

### DIFF
--- a/hivemind/dht/crypto.py
+++ b/hivemind/dht/crypto.py
@@ -1,65 +1,13 @@
 import base64
 import dataclasses
 import re
-from typing import Optional
 
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import padding, rsa
 
+from hivemind.dht.validation import DHTRecord, RecordValidatorBase
 from hivemind.utils.serializer import MSGPackSerializer
-
-
-@dataclasses.dataclass(init=True, repr=True, frozen=True)
-class DHTRecord:
-    key: bytes
-    subkey: bytes
-    value: bytes
-    expiration_time: float
-
-
-class RecordValidatorBase:
-    """
-    Record validators are a generic mechanism for checking the DHT records including:
-      - Enforcing a data schema (e.g. checking content types)
-      - Enforcing security requirements (e.g. allowing only the owner to update the record)
-    """
-
-    def validate(self, record: DHTRecord) -> None:
-        """
-        This method should:
-          - Return None if validation is successful
-          - Raise ValueError with the reason if it is not
-
-        validate() is called when another DHT peer:
-          - Asks us to store the record
-          - Returns the record by our request
-        """
-
-        raise NotImplementedError()
-
-    def sign_value(self, record: DHTRecord) -> bytes:
-        """
-        Should return `record.value` extended with the record's signature.
-
-        Note: there's no need to overwrite this method if a validator doesn't use a signature.
-
-        sign_value() is called after the application asks the DHT to store the record.
-        """
-
-        return record.value
-
-    def strip_value(self, record: DHTRecord) -> bytes:
-        """
-        Should return `record.value` stripped of the record's signature.
-        strip_value() is only called if validate() was successful.
-
-        Note: there's no need to overwrite this method if a validator doesn't use a signature.
-
-        strip_value() is called before the DHT returns the record by the application's request.
-        """
-
-        return record.value
 
 
 class RSASignatureValidator(RecordValidatorBase):
--- a/hivemind/dht/crypto.py
+++ b/hivemind/dht/crypto.py
@@ -12,7 +12,7 @@ from hivemind.utils.serializer import MSGPackSerializer
 @dataclass(init=True, repr=True, frozen=True)
 class ProtectedRecord:
     key: bytes
-    subkey: Optional[bytes]
+    subkey: bytes
     value: bytes
     expiration_time: float
 
@@ -60,8 +60,6 @@ class RSASignatureValidator(RecordValidatorBase):
         return self._our_marker
 
     def validate(self, record: ProtectedRecord, signature: bytes) -> None:
-        self._marker_re.findall(record.key)
-
         public_keys = self._marker_re.findall(record.key)
         if record.subkey is not None:
             public_keys += self._marker_re.findall(record.subkey)
@@ -81,9 +79,9 @@ class RSASignatureValidator(RecordValidatorBase):
         except InvalidSignature:
             raise ValueError('Invalid signature')
 
-    def sign(self, record: ProtectedRecord) -> Optional[bytes]:
+    def sign(self, record: ProtectedRecord) -> bytes:
         if self._our_marker not in record.key and self._our_marker not in record.subkey:
-            return None
+            return b''
 
         return self._private_key.sign(self._serialize_record(record),
                                       self._padding, self._hash_algorithm)
--- a/hivemind/dht/crypto.py
+++ b/hivemind/dht/crypto.py
@@ -12,13 +12,20 @@ from hivemind.utils.serializer import MSGPackSerializer
 @dataclass(init=True, repr=True, frozen=True)
 class ProtectedRecord:
     key: bytes
-    subkey: bytes
+    subkey: Optional[bytes]
     value: bytes
     expiration_time: float
 
 
 class RecordValidatorBase:
-    def validate(self, record: ProtectedRecord) -> bool:
+    def validate(self, record: ProtectedRecord) -> None:
+        """
+        If validation is successful, returns None.
+        If it is not, raises ValueError with the reason.
+
+        This method must not raise exceptions that are not inherited from ValueError for any input.
+        """
+
         raise NotImplementedError()
 
     def sign(self, record: ProtectedRecord) -> bytes:
@@ -62,16 +69,17 @@ class RSASignatureValidator(RecordValidatorBase):
             return
 
         if signature is None:
-            raise InvalidSignature('Signature is required but not present')
+            raise ValueError('Signature is required but not present')
         if len(set(public_keys)) > 1:
-            raise InvalidSignature("Key and subkey can't contain different public keys")
-        try:
-            public_key = serialization.load_ssh_public_key(public_keys[0])
-        except ValueError as e:
-            raise InvalidSignature(f'Failed to deserialize public key: {e}')
+            raise ValueError("Key and subkey can't contain different public keys")
+        public_key = serialization.load_ssh_public_key(public_keys[0])
 
-        public_key.verify(signature, self._serialize_record(record),
-                          self._padding, self._hash_algorithm)
+        try:
+            public_key.verify(signature, self._serialize_record(record),
+                              self._padding, self._hash_algorithm)
+            # verify() returns None iff the signature is correct
+        except InvalidSignature:
+            raise ValueError('Invalid signature')
 
     def sign(self, record: ProtectedRecord) -> Optional[bytes]:
         if self._our_marker not in record.key and self._our_marker not in record.subkey:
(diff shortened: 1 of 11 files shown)


hivemind/dht/protocol.py
```diff
@@ -221,9 +235,16 @@ async def rpc_store(self, request: dht_pb2.StoreRequest, context: grpc.ServicerC
             asyncio.create_task(self.rpc_ping(dht_pb2.PingRequest(peer=request.peer), context))
         assert len(request.keys) == len(request.values) == len(request.expiration_time) == len(request.in_cache)
         response = dht_pb2.StoreResponse(store_ok=[], peer=self.node_info)
-        keys = map(DHTID.from_bytes, request.keys)
-        for key_id, tag, value_bytes, expiration_time, in_cache in zip(
+        for key, tag, value_bytes, expiration_time, in_cache in zip(
                 keys, request.subkeys, request.values, request.expiration_time, request.in_cache):
+            if self.record_validator is not None:
+                try:
+                    self.record_validator.validate(ProtectedRecord(*record), signature)
```
justheuristic, line 242: Note: RecordValidator.validate returns bool, but this line ignores the bool value.

Request: if not record_validator.validate(...), response.store_ok.append(False); continue

hivemind/utils/crypto.py
```diff
@@ -0,0 +1,84 @@
+import re
```
justheuristic, line 1: Question ( @borzunov , @mryab ): should we store crypto here or in hivemind/dht/crypto.py?

In other words, shall we need the same type of validator outside DHT?

hivemind/utils/crypto.py
```diff
@@ -0,0 +1,84 @@
...
+
+from hivemind.utils.serializer import MSGPackSerializer
+
+
+@dataclass(init=True, repr=True, frozen=True)
+class ProtectedRecord:
+    key: bytes
+    subkey: bytes
+    value: bytes
+    expiration_time: float
```
justheuristic, line 17: Question: do we need to add sender endpoint here?

hivemind/utils/crypto.py
```diff
@@ -0,0 +1,84 @@
...
+        if signature is None:
+            raise InvalidSignature('Signature is required but not present')
+        if len(set(public_keys)) > 1:
+            raise InvalidSignature("Key and subkey can't contain different public keys")
+        try:
+            public_key = serialization.load_ssh_public_key(public_keys[0])
+        except ValueError as e:
+            raise InvalidSignature(f'Failed to deserialize public key: {e}')
+
+        public_key.verify(signature, self._serialize_record(record),
```
leshanbog, line 73: Don't we need to return result of verify()?

borzunov: [RSAPublicKey.verify](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa.html#cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey.verify) returns None if the signature is valid and throws `InvalidSignature` if it is not.

I will fix the code to make this behavior more clear for a reader.

hivemind/utils/crypto.py
borzunov: I don't see security benefits of adding it.

Also, it would require a sender to track its own up-to-date endpoint information, which is not trivial (e.g. it may not know its own public IP outside NAT or it may not spot change of it, which happens in mobile networks).

hivemind/utils/crypto.py
borzunov: I don't mind moving it to `hivemind/dht/crypto.py` :)

hivemind/dht/protocol.py
borzunov: As I wrote below, the `RSASignatureValidator` implementation actually returns None when validation is successful and throws the exception if it is not.

I'd like to keep this behavior, since there are many possible reasons for a validation failure (invalid public key, unsupported signature format, incorrect signature) and it would be useful to know the exact reason for future debugging (we can just return it in code or even log a warning with it).

Unfortunately, I forgot to update `RecordValidatorBase.validate` signature to represent this logic, so I'll fix its signature.

hivemind/utils/crypto.py
borzunov: Done

hivemind/utils/crypto.py
borzunov: Done

hivemind/dht/protocol.py
borzunov: Done

borzunov:
For `RSASignatureValidator` (backed by the `cryptography` module) and a record of size > 1 KiB, `sign_value()`, `validate()`, and `strip_value()` takes < 1 ms each:

```python
In [21]: validator = RSASignatureValidator()

In [23]: record = DHTRecord(b'key', b'subkey' + validator.ownership_marker, b'value' + b'X' * 1000, get_dht_time())

In [30]: %timeit protected_record = dataclasses.replace(record, value=validator.sign_value(record))
852 µs ± 8.94 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [35]: %timeit validator.validate(protected_record)
180 µs ± 3.06 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [36]: %timeit record = dataclasses.replace(protected_record, value=validator.strip_value(protected_record))
10.9 µs ± 105 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

So, using the validator won't interfere with the asyncio event loop.

Generating a private-public key pair (happens once before starting the node) takes ~101 ms:

```python
In [38]: %timeit validator = RSASignatureValidator()
101 ms ± 10.9 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

hivemind/dht/crypto.py
```diff
@@ -0,0 +1,134 @@
...
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import padding, rsa
+
+from hivemind.utils.serializer import MSGPackSerializer
+
+
+@dataclasses.dataclass(init=True, repr=True, frozen=True)
+class DHTRecord:
+    key: bytes
+    subkey: bytes
```
mryab, line 16: AFAIK, this field is optional

hivemind/dht/crypto.py
```diff
@@ -0,0 +1,134 @@
...
+    value: bytes
+    expiration_time: float
+
+
+class RecordValidatorBase:
```
borzunov, line 21: Утащить в dht/validation.py

borzunov, line 21: abcmeta

hivemind/dht/crypto.py
```diff
@@ -0,0 +1,134 @@
...
+    Record validators are a generic mechanism for checking the DHT records including:
+      - Enforcing a data schema (e.g. checking content types)
+      - Enforcing security requirements (e.g. allowing only the owner to update the record)
+    """
+
+    def validate(self, record: DHTRecord) -> None:
+        """
+        This method should:
+          - Return None if validation is successful
+          - Raise ValueError with the reason if it is not
```
borzunov, line 32: inherited from ValueError

hivemind/dht/crypto.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class RecordValidatorBase:
+    """
+    Record validators are a generic mechanism for checking the DHT records including:
+      - Enforcing a data schema (e.g. checking content types)
+      - Enforcing security requirements (e.g. allowing only the owner to update the record)
+    """
+
+    def validate(self, record: DHTRecord) -> None:
```
borzunov, line 28: Return bool and log errors immediately

hivemind/dht/protocol.py
```diff
@@ -190,7 +195,8 @@ async def call_store(self, peer: Endpoint, keys: Sequence[DHTID],
 
         in_cache = in_cache if in_cache is not None else [False] * len(keys)  # default value (None)
         in_cache = [in_cache] * len(keys) if isinstance(in_cache, bool) else in_cache  # single bool
-        keys, subkeys, values, expiration_time, in_cache = map(list, [keys, subkeys, values, expiration_time, in_cache])
```
borzunov, line 193: Remove from diff

hivemind/dht/node.py
```diff
@@ -637,10 +653,24 @@ def finish_search(self):
         elif not self.found_something:
             self.future.set_result(None)
         elif isinstance(self.binary_value, BinaryDHTValue):
-            self.future.set_result(ValueWithExpiration(self.serializer.loads(self.binary_value), self.expiration_time))
+            value_bytes = self.binary_value
+            if self.record_validator is not None:
+                # FIXME: Fill other fields (some validators may need them)
+                record = DHTRecord(None, None, value_bytes, None)
```
borzunov, line 659: do FIXME

mryab (changes requested):

hivemind/dht/crypto.py
borzunov: It is not optional here, at the protocol layer. In case of sending a non-dictionary value or a whole dictionary, the subkey contains a [special `bytes` value](https://github.com/learning-at-home/hivemind/blob/master/hivemind/dht/protocol.py#L26).

tests/test_dht_node.py
```diff
@@ -454,3 +455,33 @@ async def test_dhtnode_edge_cases():
...
+
+    assert await bob.store(key, b'updated_true_value', hivemind.get_dht_time() + 10, subkey=subkey)
+    assert (await alice.get(key, latest=True)).value[subkey].value == b'updated_true_value'
+
+    await bob.shutdown()  # Bob has shut down, now Mallory is the single peer of Alice
+
+    store_ok = await mallory.store(key, b'updated_fake_value',
+                                   hivemind.get_dht_time() + 10, subkey=subkey)
+    assert not store_ok
+    assert (await alice.get(key, latest=True)).value[subkey].value == b'updated_true_value'
```
mryab, line 487: I still think we need the following test: writing in a protected key with a different signature should be possible after the previous record expires

mryab (changes requested):

hivemind/dht/crypto.py
```diff
@@ -0,0 +1,88 @@
...
+        self._marker_re = re.compile(re.escape(marker_format).replace(b'_key_', rb'(.+?)'))
+
+        self._signature_format = signature_format
+        self._signature_re = re.compile(re.escape(signature_format).replace(b'_value_', rb'(.+?)'))
+
+        self._private_key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
+
+        serialized_public_key = self._private_key.public_key().public_bytes(
+            encoding=serialization.Encoding.OpenSSH, format=serialization.PublicFormat.OpenSSH)
+        self._our_marker = marker_format.replace(b'_key_', serialized_public_key)
```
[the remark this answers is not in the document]
borzunov: Renamed it to `self._ownership_marker`, is it okay?

tests/test_dht_node.py
justheuristic: [based on discussion with @borzunov ]
Alas, this functionality is not supported and we will likely have to do without it.

The validation of (key, subkey, value, expiration) tuple does not depend on whether or not there is a previous value under that key.

Unfortunately, conditioning the validator on previous DHT keys would not be secure since
- not all DHT peers that are supposed to have a given value will actually have it at any given moment in time. An attacker can use this to bypass the protection.
- DHT peers sometimes need to pass information to new nodes [e.g. welcome_if_new](https://github.com/learning-at-home/hivemind/blob/master/hivemind/dht/protocol.py#L337).

Fortunately, it seems that there is a workaround that will allow us to secure all hivemind components with the current validator interface.
- Averager/metadata: all secret keys are unique and tied to peer identity. No two peers will even need (or be allowed to write) to the same (key, subkey) pair
- Experts / beam search: an authority (server or blockchain) will lease a separate security key for each expert that gives the use permission to declare this expert. This security key will be tied to (expert uuid, peer_id/endpoint, max expiration).

mryab (approved):

justheuristic (approved):

PR #190, opened 2021-03-21
Fix scalar deserialization

leshanbog:
Optimizers may have scalars in their state dict. So, we need a proper handling of such cases.

### DIFF
--- a/hivemind/utils/grpc.py
+++ b/hivemind/utils/grpc.py
@@ -215,9 +215,17 @@ def serialize_torch_tensor(tensor: torch.Tensor, compression_type=CompressionTyp
 
 def deserialize_torch_tensor(serialized_tensor: runtime_pb2.Tensor) -> torch.Tensor:
     # TODO avoid copying the array (need to silence pytorch warning, because array is not writable)
+
+    def construct_torch_tensor(serialized_tensor, array, dtype=None):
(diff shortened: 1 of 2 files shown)


hivemind/utils/grpc.py
```diff
@@ -215,9 +215,17 @@ def serialize_torch_tensor(tensor: torch.Tensor, compression_type=CompressionTyp
 
 def deserialize_torch_tensor(serialized_tensor: runtime_pb2.Tensor) -> torch.Tensor:
     # TODO avoid copying the array (need to silence pytorch warning, because array is not writable)
+
+    def construct_torch_tensor(serialized_tensor, array, dtype=None):
```
mryab, line 219: 1. Better to move the function from outer scope
2. Better to make `array` the first argument and to pass only the tensor size as input

hivemind/utils/grpc.py
```diff
@@ -215,9 +215,17 @@ def serialize_torch_tensor(tensor: torch.Tensor, compression_type=CompressionTyp
...
+        """ Helper conversion function that handles some edge case """
```
mryab, line 220: Specify the exact edge case, please

mryab (changes requested):

mryab (approved):

PR #193, opened 2021-03-24
Py libp2p bindings

MaximKsh:


hivemind/proto/crypto.proto
```diff
@@ -0,0 +1,20 @@
...
+
+message PublicKey {
+  required KeyType key_type = 1;
+  required bytes data = 2;
+}
+
+message PrivateKey {
+  required KeyType key_type = 1;
+  required bytes data = 2;
+}
```
justheuristic, line 20: \n

hivemind/p2p/p2p_daemon_bindings/utils.py
```diff
@@ -0,0 +1,81 @@
...
+    stream.write(msg_bytes)
+
+
+async def read_pbmsg_safe(stream: asyncio.StreamReader, pbmsg: PBMessage) -> None:
+    len_msg_bytes = await read_unsigned_varint(stream)
+    msg_bytes = await stream.readexactly(len_msg_bytes)
+    pbmsg.ParseFromString(msg_bytes)
+
+
+def get_unused_tcp_port() -> int:
```
justheuristic, line 77: in future, let's migrate to hivemind.utils.find_open_port (does exactly the same as this function)

justheuristic (approved):
LGTM

My only real concern is that the user-facing methods of p2pclient (listen/identify/stream_open/...) should be given docstrings and included into readthedocs. However, this can be done later upon merge into master

PR #196, opened 2021-03-28
Support learning rate schedulers in ExpertBackend

mryab:
Besides this, also added empty __init__.py for package discovery by setuptools and extracted two functions to get expert stats/full data and fixed a bug with BatchTensorDescriptor (we shouldn't pass compression to make_empty).

To reproduce the bug in master:
```
>>> import hivemind
>>> sample_input = hivemind.server.layers.name_to_input['ffn'](4, 16)
>>> args_schema = (hivemind.BatchTensorDescriptor.from_tensor(sample_input),)
>>> args_schema[0].make_empty(1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mryab/hivemind/hivemind/utils/tensor_descr.py", line 61, in make_empty
    return super().make_empty(size=(batch_size, *self.shape[1:]), **kwargs)
  File "/home/mryab/hivemind/hivemind/utils/tensor_descr.py", line 40, in make_empty
    return torch.empty(**properties)
TypeError: empty() received an invalid combination of arguments - got (compression=int, pin_memory=bool, dtype=torch.dtype, layout=torch.layout, size=tuple, device=torch.device, requires_grad=bool, ), but expected one of:
 * (tuple of ints size, *, tuple of names names, torch.memory_format memory_format, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
 * (tuple of ints size, *, torch.memory_format memory_format, Tensor out, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
```


hivemind/hivemind_cli/run_server.py
```diff
@@ -28,13 +29,20 @@ def main():
...
                         help='server will use this many processes to handle incoming requests')
     parser.add_argument('--max_batch_size', type=int, default=16384, required=False,
                         help='The total number of examples in the same batch will not exceed this value')
     parser.add_argument('--device', type=str, default=None, required=False,
                         help='all experts will use this device in torch notation; default: cuda if available else cpu')
+
     parser.add_argument('--optimizer', type=str, default='adam', required=False, help='adam, sgd or none')
+    parser.add_argument('--scheduler_override', type=str, choices=schedule_name_to_scheduler.keys(), required=False,
+                        help='LR scheduler type to use instead of default one for expert_cls')
+    parser.add_argument('--num-warmup-steps',type=int,required=False,help='the number of warmup steps for LR schedule')
```
justheuristic, line 43:
```suggestion
    parser.add_argument('--num-warmup-steps', type=int, required=False, help='the number of warmup steps for LR schedule')
```

hivemind/hivemind_cli/run_server.py
```diff
@@ -28,13 +29,20 @@ def main():
...
+    parser.add_argument('--num-training-steps',type=int,required=False,help='the number of total steps for LR schedule')
```
justheuristic, line 44:
```suggestion
    parser.add_argument('--num-training-steps', type=int, required=False, help='the total number of steps for LR schedule')
```

hivemind/hivemind_cli/run_server.py
```diff
@@ -28,13 +29,20 @@ def main():
...
     parser.add_argument('--num_handlers', type=int, default=None, required=False,
                         help='server will use this many processes to handle incoming requests')
     parser.add_argument('--max_batch_size', type=int, default=16384, required=False,
                         help='The total number of examples in the same batch will not exceed this value')
     parser.add_argument('--device', type=str, default=None, required=False,
                         help='all experts will use this device in torch notation; default: cuda if available else cpu')
+
     parser.add_argument('--optimizer', type=str, default='adam', required=False, help='adam, sgd or none')
+    parser.add_argument('--scheduler_override', type=str, choices=schedule_name_to_scheduler.keys(), required=False,
+                        help='LR scheduler type to use instead of default one for expert_cls')
```
justheuristic, line 42:
```suggestion
                        help='LR scheduler type to use instead of the default one for expert_cls')
```

hivemind/server/expert_backend.py
```diff
@@ -1,14 +1,14 @@
...
 
 from hivemind.server.task_pool import TaskPool
-from hivemind.utils import nested_flatten, nested_pack, nested_compare, nested_map, \
-    BatchTensorDescriptor, DUMMY_BATCH_SIZE
+from hivemind.utils import nested_flatten, nested_pack, nested_compare, nested_map, BatchTensorDescriptor, \
+    DUMMY_BATCH_SIZE
 
 
-class ExpertBackend(nn.Module):
+class ExpertBackend:
```
justheuristic, line 11: please explain the motivation behind removing nn.module dependence

[the reason it was there: for convenience in backend.cuda() / backend.parameters()]

hivemind/server/expert_backend.py
```diff
@@ -26,20 +26,31 @@ class ExpertBackend(nn.Module):
...
         See hivemind.utils.custom_layers.DeterministicDropout for an example
 
-    :param opt: torch optimizer to be applied on every backward call
+    :param optimizer: torch optimizer to be applied on every backward call
+    :param scheduler: a function to create the learning rate scheduler for the expert
     :param args_schema: description of positional arguments to expert.forward, list of BatchTensorProto
     :param kwargs_schema: description of keyword arguments to expert.forward, dict of BatchTensorProto
     :param outputs_schema: description of outputs from expert.forward, nested structure of BatchTensorProto
+    :param num_warmup_steps: the number of warmup steps for LR schedule
+    :param num_training_steps: the number of total steps for LR schedule
```
justheuristic, line 35:
```suggestion
    :param num_training_steps: the total number of steps for LR schedule
```

hivemind/server/expert_backend.py
```diff
@@ -121,18 +135,54 @@ def backward(self, *inputs: torch.Tensor) -> Tuple[torch.Tensor, ...]:
...
                      for x in nested_flatten((args, kwargs)))
 
-    def apply_gradients(self) -> None:
+    def apply_gradients(self, batch_size) -> None:
         """
         Train the expert for one step. This method is called by ``ExpertBackend.backward`` after computing gradients.
         """
-        self.opt.step()
-        self.opt.zero_grad()
+        # TODO support gradient accumulation
```
justheuristic, line 147: [suggestion] please remove this TODO and create an issue before merging

hivemind/server/layers/__init__.py
```diff
@@ -10,7 +11,12 @@
 
 name_to_input = {'ffn': lambda batch_size, hid_dim: torch.empty((batch_size, hid_dim)),
                  'transformer': lambda batch_size, hid_dim:
-                 (torch.empty((batch_size, 128, hid_dim)), torch.empty((batch_size, hid_dim), dtype=torch.bool)),
+                 (torch.empty((batch_size, 128, hid_dim)), torch.empty((batch_size, 128), dtype=torch.bool)),
                  'nop': lambda batch_size, hid_dim: torch.empty((batch_size, hid_dim)),
                  'det_dropout': lambda batch_size, hid_dim:
                  (torch.empty((batch_size, hid_dim)), torch.randint(0, 1, (batch_size, hid_dim)))}
+
+expert_name_to_scheduler = {'ffn': get_linear_schedule_with_warmup, 'transformer': get_linear_schedule_with_warmup,
```
justheuristic, line 19: arguably the default should be none - that way we do not run a risk of accidentally using a schedule (and the config will always contain the schedule explicitly.

In contrast, if we keep this as a default, we will have to break compatibility whenever we decide upon a new better schedule

justheuristic (approved):
My only real concern is that we should, arguably, default to 'none' schedule instead of a custom decay scheme.

This has two motivations:
- it will be harder to unwittingly use a schedule if one doesn't want one
- if we find a better schedule in the future, we won't have to break compatibility

hivemind/server/expert_backend.py
mryab: It was never a proper nn.Module to begin with (no `forward`, for instance), and after adding the schedule I concluded that it might be misleading to still treat it as a module, since `backend.state_dict()` will not contain the entire state to be saved

hivemind/server/expert_backend.py
mryab: Created https://github.com/learning-at-home/hivemind/issues/198

hivemind/server/layers/__init__.py
mryab: Replaced and removed expert_name_to_scheduler

hivemind/server/expert_backend.py
justheuristic: [agreed]

hivemind/server/__init__.py
```diff
@@ -70,9 +70,10 @@ def __init__(
 
     @staticmethod
     def create(listen_on='0.0.0.0:*', num_experts: int = None, expert_uids: str = None, expert_pattern: str = None,
-               expert_cls='ffn', hidden_dim=1024, optim_cls=torch.optim.Adam, num_handlers=None, max_batch_size=4096,
-               device=None, no_dht=False, initial_peers=(), dht_port=None, checkpoint_dir: Optional[Path] = None,
-               load_experts=False, compression=CompressionType.NONE, *, start: bool, **kwargs) -> Server:
+               expert_cls='ffn', hidden_dim=1024, optim_cls=torch.optim.Adam, scheduler: str = 'none',
+               num_warmup_steps=None, num_training_steps=None, num_handlers=None, max_batch_size=4096, device=None,
+               no_dht=False, initial_peers=(), dht_port=None, checkpoint_dir: Optional[Path] = None,
+               load_expert_states=False, compression=CompressionType.NONE, *, start: bool, **kwargs) -> Server:
```
romakail, line 76: Can we check the necessity to load the expert only with the checkpoint_dir parameter?  If it,s None with initialize it from the scratch, if it's a path - we load state from this path.

hivemind/server/__init__.py
```diff
@@ -241,7 +252,7 @@ def shutdown(self):
 def background_server(*args, shutdown_timeout=5, **kwargs) -> Tuple[hivemind.Endpoint, hivemind.Endpoint]:
     """ A context manager that creates server in a background thread, awaits .ready on entry and shutdowns on exit """
     pipe, runners_pipe = mp.Pipe(duplex=True)
-    runner = mp.get_context("spawn").Process(target=_server_runner, args=(runners_pipe, *args), kwargs=kwargs)
+    runner = mp.Process(target=_server_runner, args=(runners_pipe, *args), kwargs=kwargs)
```
romakail, line 255: This might be the solution to probem me and JustHeuristic faced while implementing custom experts.

hivemind/server/expert_backend.py
```diff
@@ -121,18 +135,53 @@ def backward(self, *inputs: torch.Tensor) -> Tuple[torch.Tensor, ...]:
...
+        return full_state
+
+    def load_full_state(self, state_dict: Dict):
+        self.update_count = state_dict['stats']['updates']
+        self.examples_processed = state_dict['stats']['examples_processed']
+
+        self.expert.load_state_dict(state_dict['model'])
+        self.optimizer.load_state_dict(state_dict['optimizer'])
+        if self.scheduler is not None:
+            self.scheduler.load_state_dict(state_dict['scheduler'])
```
romakail, line 184: I think that existence of keys "optimizer", "stats" and "scheduler" in the state_dict should be checked before loading_state. This might be helpful if the user has already trained his own model and wants to load it as an expert.

romakail (commented):
- Maybe some replacement of backend.parameters() is worth implementing. An iterator over weights might be useful for example to calculate them.

hivemind/server/__init__.py
mryab: Done

hivemind/server/expert_backend.py
mryab: Done

justheuristic (approved):
[i have nothing else to say]
LGTM

romakail (approved):
I have no more questions

PR #207, opened 2021-04-05
Better exception handling in Runtime

mryab:


justheuristic (approved):
:+1:

Note: can we move the if not set_result logic from MPFuture to Server?

Issue #184, opened 2021-03-17
Support communication-efficient optimizers


ploshkin:
`<edit>`
The goal of this project is to implement pre-configured decentralized optimizers such as [DADAM](https://arxiv.org/abs/1901.09109) and decentralized [Local AMSGrad](https://dl.acm.org/doi/10.1145/3412815.3416891). We intend to create a general interface that runs adaptive optimizers with periodic averaging of gradients, parameters and optimizer statistics on different time-frames]

The final version should support:
* [ ] wrapping any adam-like optimizers such as Adam, AMSGrad, LAMB(FusedLAMB), Radam, etc
* [ ] full torch.optim.Optimizer interface including state_dict/load_state_dict
   * [ ] for parameter groups, either support it or explicitly throw an error
* [ ] averaging optimizer statistics in background while the model is training, incorporating local updates after the averaging is done
* [ ] have a default confguration that converges :)
  * [ ] test convergence locally with a toy task
  * [ ] test in standard ALBERT-base setup for wikitext103
  * [ ] test with ALBERT-large on OpenBookCorpus

`</edit>`

Code draft:

```
# usage:
class LAMBWithAveraging(FusedLAMB, AdaptiveOptimizer):
    def get_averaged_statistics() -> Sequence[Tensor]
       return TODO()

opt = DecentralizedOptimizer(model, base_optimizer=LAMBWithAveraging(model.parameters()),
                             prefix='MY_UNIQUE_STRING')


### code

class AdaptiveOptimizer(abc.Mixin):
   def get_averaged_statistics(self):
     raise NotImplementedError()

class DecentralizedOptimizer(abc.meta):
    def __init__(self, parameters, base_optimizer, dht, prefix: str, ..., ):
       assert isinstance(base_optimizer, AdaptiveOptimizer), "TODO see example here"

       self.parameters = list(parameters)  # on GPU
       self.base_optimizer = base_optimizer
       self.global_step = ...
       
       zeros_like_grads = [torch.zeros_like(x.detach().cpu()) for x in self.parameters]
       self.grad_averager = DecentralizedAverager(
           averaged_tensors= zeros_like_grads  # average gradients only
            prefix=prefix+'_grad_averager', **some_kwargs
           )
       
       parameters = [x.cpu().detach().float() self.parameters()]
       averaged_statistics = [x.cpu().detach().float() for x in base_optimizer.get_averaged_statistics()]
       
       self.almost_v_averager = DecentralizedAverager(
[...]

PR #210, opened 2021-04-06
Implement basic DecentralizedOptimizer

justheuristic:
This PR implements a simple DecentralizedOptimizer that averages model parameters with peers. It is meant as a base implementation that will will be extended with #184 and a collaborative-training-like optimizer in subsequent PRs.


hivemind/client/averaging/__init__.py
```diff
@@ -196,14 +197,14 @@ def run_in_background(self, await_ready=True, timeout=None):
...
-        if self.is_alive():
+        if self._parent_pid != os.getpid() or self.is_alive():
             self._pipe.send(('_SHUTDOWN', None))
             self.terminate()
         else:
             logger.warning("DHT shutdown has no effect: the process is not alive")
 
     def __del__(self):
-        if self.is_alive():
+        if self._parent_pid != os.getpid() or self.is_alive():
```
justheuristic, line 207: This check fixes an edge case where deleting muiltiple averagers at once may trigger AssertionError "Can only check is_alive for child processes"

justheuristic:
TODOS:
* add a more advanced test (aka benchmark) in a PR with full optimizer
* update sphinx docs
* [refactoring] maybe push load_state_from_peers from averager into DecentralizedOptimizer

tests/test_training.py
```diff
@@ -36,3 +37,31 @@ def test_training(max_steps: int = 100, threshold: float = 0.9):
                 break
 
         assert accuracy >= threshold, f"too small accuracy: {accuracy}"
+
+
+@pytest.mark.forked
+def test_decentralized_optimizer_step():
```
yhn112, line 43: I would like to have a more complex test here, but sure, we don't need it immediately.

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+
+class DecentralizedOptimizer(DecentralizedOptimizerBase):
+    """
+    A simple optimizer that trains a shared model by averaging model parameters with peers in the background.
+
+    :param opt: a pytorch optimizer configured to update model parameters.
+    :param dht: a running hivemind DHT daemon connected to other peers
+    :param prefix: all DHT keys that point to optimization metadata will have this prefix
+    :param target_group_size: maximum group size for averaging (see DecentralizedAverager)
+    :param kwargs: additional parameters passed to DecentralizedAverager
```
mryab, line 22: We need an explicit argument for update frequency in the __init__ for this class, so that it isn't necessary to look for the frequency in other docstrings

tests/test_training.py
```diff
@@ -36,3 +37,31 @@ def test_training(max_steps: int = 100, threshold: float = 0.9):
...
+
+    param1 = torch.nn.Parameter(torch.zeros(32, 32), requires_grad=True)
+    opt1 = DecentralizedOptimizer(opt=torch.optim.SGD([param1], lr=0.1),
+                                  dht=DHT(initial_peers=initial_peers, start=True),
+                                  prefix='foo', target_group_size=2, verbose=True)
+
+    param2 = torch.nn.Parameter(torch.ones(32, 32), requires_grad=True)
+    opt2 = DecentralizedOptimizer(opt=torch.optim.SGD([param2], lr=0.05),
+                                  dht=DHT(initial_peers=initial_peers, start=True),
+                                  prefix='foo', target_group_size=2, verbose=True)
```
mryab, line 55: Also, pass the update frequency to these optimizers to ensure consistency with `time.sleep` below

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+            if group_info is not None:
+                with torch.no_grad(), averager.get_tensors() as averaged_tensors:
+                    for local_tensor, averaged_tensor in zip(local_tensors, averaged_tensors):
+                        local_tensor[...] = averaged_tensor.to(dtype=local_tensor.dtype)
+                if verbose:
+                    logger.info(f"Finished averaging round in with {len(group_info)} peers.")
+            else:
+                logger.info(f"Averaging round failed: could not find group.")
+        except Exception as e:
+            logger.info(f"Averaging round failed: caught {e}.")
```
mryab, line 88: `logger.error`?

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+            group_info = averager.step(**kwargs)
+
+            if group_info is not None:
+                with torch.no_grad(), averager.get_tensors() as averaged_tensors:
+                    for local_tensor, averaged_tensor in zip(local_tensors, averaged_tensors):
+                        local_tensor[...] = averaged_tensor.to(dtype=local_tensor.dtype)
+                if verbose:
+                    logger.info(f"Finished averaging round in with {len(group_info)} peers.")
+            else:
+                logger.info(f"Averaging round failed: could not find group.")
```
mryab, line 86: `logger.warning`?

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+
+        try:
+            if verbose:
+                logger.info(f"Starting a new averaging round with current parameters.")
+            group_info = averager.step(**kwargs)
+
+            if group_info is not None:
+                with torch.no_grad(), averager.get_tensors() as averaged_tensors:
+                    for local_tensor, averaged_tensor in zip(local_tensors, averaged_tensors):
+                        local_tensor[...] = averaged_tensor.to(dtype=local_tensor.dtype)
```
mryab, line 82: It's best to explicitly write `copy_()` for the local tensor

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+        update_event.wait()
+        update_event.clear()
+        if stop_event.is_set():
+            break
+
+        with torch.no_grad(), averager.get_tensors() as averaged_tensors:
+            local_tensors = tuple(p for group in opt.param_groups for p in group['params'])
+            assert len(local_tensors) == len(averaged_tensors), "The number of optimized parameters should not change."
+            for local_tensor, averaged_tensor in zip(local_tensors, averaged_tensors):
+                averaged_tensor[...] = local_tensor.cpu().float()
```
mryab, line 72: 1) It's best to explicitly write copy_() for the local tensor
2) What if the dtype of averaged_tensor isn't float?

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+    def __del__(self):
+        self.stop_event.set()
+        while self.background_averaging_thread.is_alive():
+            self.update_event.set()
+
+
+def _average_parameters_in_background(update_event: Event, stop_event: Event, averager: DecentralizedAverager,
+                                      opt: torch.optim.Optimizer, verbose: bool, **kwargs):
+    """ Iteratively find groups of peers, average parameters with these peers and update local model parameters. """
+    while True:
```
mryab, line 62:
```suggestion
    while not stop_event.is_set():
```

mryab: This way, you can get rid of `break` below

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+        update_event.wait()
+        update_event.clear()
+        if stop_event.is_set():
+            break
+
+        with torch.no_grad(), averager.get_tensors() as averaged_tensors:
```
mryab, line 68: I think this entire function can be wrapped in torch.no_grad

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+                                                  kwargs=dict(verbose=verbose, timeout=timeout))
+        self.background_averaging_thread.start()
+
+    def step(self, closure: Optional[Callable[[], Any]] = None):
+        self.update_event.set()
+        with self.lock_parameters:
+            return self.opt.step(closure)
+
+    def zero_grad(self, set_to_none: Optional[bool] = False):
+        return self.opt.zero_grad()
```
mryab, line 51: `set_to_none` isn't passed to the original optimizer

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+    :param prefix: all DHT keys that point to optimization metadata will have this prefix
+    :param target_group_size: maximum group size for averaging (see DecentralizedAverager)
+    :param kwargs: additional parameters passed to DecentralizedAverager
+
+    :note: using DecentralizedOptimizer with adaptive learning rates may result in poor convergence due to
+      out-of-sync adaptive learning rates (such as adam second momentum or schedule step). Please ensure that these
+      statistics are synchronized or use a more advanced DecentralizedOptimizer version, if applicable.
+    :note: the base optimizer cannot add param groups after the DecentralizedOptimizer is created
+    """
+    averager_step: Optional[MPFuture] = None
```
mryab, line 29: Unneeded variable?

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+
+from hivemind.dht import DHT
+from hivemind.client.averaging import DecentralizedAverager
+from hivemind.client.optim.base import DecentralizedOptimizerBase
+from hivemind.utils import MPFuture, get_logger
+
+logger = get_logger(__name__)
+
+
+class DecentralizedOptimizer(DecentralizedOptimizerBase):
```
mryab, line 14: `ParameterAveragingOptimizer`? (just to clarify that this optimizer only does parameter averaging)

mryab: Alternatively, we can have one DecentralizedOptimizer class which averages parameters,gradients or statistics depending on use case. It shouldn't be too large, given that most of the work is done in DecentralizedAverager

mryab (changes requested):

hivemind/client/optim/simple.py
```diff
@@ -0,0 +1,88 @@
...
+            return self.opt.step(closure)
+
+    def zero_grad(self, set_to_none: Optional[bool] = False):
+        return self.opt.zero_grad()
+
+    def __del__(self):
+        self.stop_event.set()
+        while self.background_averaging_thread.is_alive():
+            self.update_event.set()
+
```
nevec, line 57: I'd like to have an explicit method for shutting down the averager

nevec (approved):

tests/test_training.py
justheuristic: Resolution: will add a more complex test for a more advanced optimizer in a subsequent PR.

hivemind/client/optim/base.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def add_param_group(self, param_group: dict) -> None:
+        raise ValueError(f"{self.__class__.__name__} does not support calling add_param_group after creation."
+                         f"Please provide all parameter groups at init.")
+
+    def state_dict(self) -> dict:
+        return self.opt.state_dict()
+
+    def load_state_dict(self, state_dict: dict) -> None:
+        return self.opt.load_state_dict(state_dict)
```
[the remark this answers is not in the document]
justheuristic: it was my mistake that load_state_dict returned None. In actuality, some optimizers report statuses like "all keys matched"

hivemind/client/optim/simple.py
justheuristic: Solution: implemented averaging_period, measured in seconds. Here's how it works:

If the optimizer is active, the background thread will await until get_dht_time is divisible by averaging_period, then begin averaging.

This trick ensures that all peers begin averaging at almost the same time. Otherwise, having a long averaging_period will result in a situation where only a small fraction of peers is active at any given point in time and no peer is able to find group (to say nothing about following the moshpit protocol)

tests/test_training.py
justheuristic: Alas, the time.sleep is meant not for consistency with averaging period, but simply to make sure that the slow CI executor was able to... well, do the averaging :)

hivemind/client/optim/simple.py
justheuristic: Fixed

hivemind/client/optim/simple.py
justheuristic: Alas, it aint that simple:
* It is important that `__del__` handler will terminate the thread immediately without running an extra round of averaging
* We could move update_event.wait/clear to the end of the cycle, but we must wait for the first update to run the averaging

Which one do you prefer?
- move update_event.wait()/clear() to the end of the cycle, add an extra wait()/clear() before the cycle
- check for stop_event twice - in the `while` condition _and_ directly after wait()/clear()
- keep the code as is?

hivemind/client/optim/simple.py
justheuristic: nice catch!

hivemind/client/optim/simple.py
justheuristic: Renamed (first option)

While most of the work is indeed handled by the DecentralizedAverager, i am unsure which user inter would be preferable for the generalized optimizer (with local gradients, statistics, etc).

Let us handle that in subsequent pull requests.

hivemind/client/optim/simple.py
justheuristic: thanks!

hivemind/client/optim/simple.py
justheuristic: agreed

hivemind/client/optim/simple.py
justheuristic: done

hivemind/client/optim/simple.py
justheuristic: Also: added a DecentralizedSGD ([term usage example](https://arxiv.org/pdf/2003.10422.pdf)) wrapper for convenience -- mainly not to scare people off with ParameterAveragingOptimizer. To the best of my knowledge, SGD is the only mainstream optimizer from torch.optim that is guaranteed to converge with parameter averaging.

Note: i intend to use DecentralizedSGD for resnet50 example after some performance tweaks.

mryab (approved):

PR #211, opened 2021-04-07
Replace FeedforwardBlock with a correct implementation

mryab:
Previously, we had a larger version of the FFN layer that had more parameters and ReLU as an activation. This PR replaces the default FFN layer with an implementation used in ALBERT by the [transformers](https://github.com/huggingface/transformers/blob/master/src/transformers/models/albert/modeling_albert.py#L374-L398) library as a reasonable default.

### DIFF
--- a/hivemind/server/layers/common.py
+++ b/hivemind/server/layers/common.py
@@ -13,12 +13,13 @@ class FeedforwardBlock(nn.Module):
         super().__init__()
         self.ffn = nn.Linear(hid_dim, 4 * hid_dim)
         self.ffn_output = nn.Linear(4 * hid_dim, hid_dim)
+        self.layer_norm = nn.LayerNorm(hid_dim, eps=1e-12)
 
     def forward(self, x):
         ffn_output = self.ffn(x)
         ffn_output = gelu_fast(ffn_output)
         ffn_output = self.ffn_output(ffn_output)
-        return ffn_output
+        return self.layer_norm(x + ffn_output)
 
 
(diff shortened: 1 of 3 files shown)


hivemind/server/layers/common.py
```diff
@@ -2,21 +2,23 @@
...
-        self.layers = nn.Sequential(
-            nn.Linear(hid_dim, 4 * hid_dim),
-            nn.LayerNorm(4 * hid_dim),
-            nn.ReLU(inplace=True),
-            nn.Linear(4 * hid_dim, 4 * hid_dim),
-            nn.LayerNorm(4 * hid_dim),
-            nn.ReLU(inplace=True),
-            nn.Linear(4 * hid_dim, hid_dim),
-        )
+        self.ffn = nn.Linear(hid_dim, 4 * hid_dim)
```
justheuristic, line 14: Consideration: what do you think about using pre-layer-norm here by default?

justheuristic (approved):
Request: changing FeedforwardBlock will alter the benchmarking performance reported [here](https://learning-at-home.readthedocs.io/en/latest/user/benchmarks.html#server-throughput)

image

Can you please update the performance stats there?

mryab:
Re: benchmarks, we'll update the entire docs section when most of the pending server-side features are implemented and merged to master

hivemind/server/layers/common.py
mryab: Added Post-LN for consistency with BERT etc.

Issue #216, opened 2021-04-09
Training SWAV with Decentralized SGD


justheuristic:
Let's add a tutorial for training SWAV/ResNet50 with Decentralized SGD

The intent is to use DecentralizedSGD optimizer with [vissl](https://github.com/facebookresearch/vissl) library for swav.

Here's a basic tutorial for training simclr in vissl: https://colab.research.google.com/drive/1Rt3Plt3ph84i1A-eolLFafybwjrBFxYe?usp=sharing .

The engineering is up to you, but it appears that the two hardest tasks will be to
- modify [vissl trainer](https://dl.fbaipublicfiles.com/vissl/tutorials/run_distributed_engines.py) to use DecentralizedSGD instead of fixed parallelism
- tune [DecentralizedSGD](https://github.com/learning-at-home/hivemind/blob/master/hivemind/client/optim/simple.py) performance for swav (see below)

The main request for DecentralizedSGD is to implement a training regime where it would be able to run averaging all the time with latest parameters. The issue with the current implementation is that DecentralizedSGD will spend up to half of the time looking for groups and when it will end up averaging model parameters, these parameters will be an older snapshot since before the averager began looking for group. Finally, when the averager [actually updates model parameters](https://github.com/learning-at-home/hivemind/blob/master/hivemind/client/optim/simple.py#L95-L97), these updates will disregard any local changes to model parameters made during averaging.

Here's a few ideas on how to improve DecentralizedSGD:
- [ ] after an evaraging round, instead of [overwriting](https://github.com/learning-at-home/hivemind/blob/master/hivemind/client/optim/simple.py#L95-L97), it would be better to compute __weight = weight + averaged_weight - weight_before_averaging_step__. This will prevent DecentralizedSGD from disregarding local updates concurrent with averager.step.
- [ ] in DecentralizedAverager, we can implement a callback that allows the user to update the model parameters right before the beginning of AllReduce (i.e. after the group is formed). This should significantly reduce the staleness of averaged parameters.
- [ ] in DecentralizedSGD, we can modify the code for calling averager step to allow for concurrent matchmaking and allreduce. In other words, once the averager has found one group, let him immediately look for the next group while still running allreduce.

Implementing this into an example will require the following steps:
- [ ] create a root folder, e.g. `./hivemind/examples/swav`, containing...
- [ ] modified training runner that uses DecentralizedSGD
[...]

PR #218, opened 2021-04-12
Prevent DecentralizedSGD from accidentally skipping a fraction of training batches

ploshkin:
Relates: #216


justheuristic (approved):
My only request: before merging, please change the PR name to reflect the content of this PR

e.g. "prevent Decentralized SGD from accidentally skipping a fraction of training batches"

PR #227, opened 2021-04-16
Add DHT schema validator

borzunov:
This PR introduces the DHT schema validator that allows an application to restrict the DHT content format using the [Cerberus](https://docs.python-cerberus.org/en/stable/) module.


justheuristic (approved):
Well done!
/* i have no meaningful change requests */

hivemind/dht/node.py
```diff
@@ -368,7 +368,7 @@ def _update_cache_on_store(self, key_id: DHTID, subkeys: List[Subkey], binary_va
             self.protocol.cache.store(key_id, stored_value_bytes, stored_expiration)
         elif not store_succeeded and not is_dictionary:  # store rejected, check if local cache is also obsolete
             rejected_value, rejected_expiration = max(zip(binary_values, expirations), key=lambda p: p[1])
-            if (self.protocol.cache.get(key_id)[1] or float("inf")) <= rejected_expiration:  # cache would be rejected
+            if (self.protocol.cache.get(key_id) or (None, float("inf")))[1] <= rejected_expiration:  # cache would be rejected
```
justheuristic, line 371: Great catch!

I am a bit confused as for what could cause this error to occur.

AFAIU, the key is written into the cache in .store_many before this method is called, and the error occurs the value expires *before* the store is finished -- or is removed due to cache size limit.

Is this what triggered the error in your case?

borzunov: Thanks!

The error was triggered when a node has refused to store a record that has not passed the validator (in this case, `store_succeeded = False` as well).

borzunov:
I've updated the node to refuse to store invalid records locally.

This didn't make sense for the signature validator (locally signed records are always valid) but is relevant for the schema validator (the application's peer may try storing an invalid record in its own DHT because of a bug).

I will re-request the review because of these non-trivial changes.

hivemind/dht/node.py
```diff
@@ -368,11 +376,13 @@ def _update_cache_on_store(self, key_id: DHTID, subkeys: List[Subkey], binary_va
...
                 self._schedule_for_refresh(key_id, refresh_time=get_dht_time())  # fetch new key in background (asap)
         elif is_dictionary and key_id in self.protocol.cache:  # there can be other keys and we should update
-            for subkey, stored_value_bytes, expiration_time in zip(subkeys, binary_values, expirations):
-                self.protocol.cache.store_subkey(key_id, subkey, stored_value_bytes, expiration_time)
+            for subkey, stored_value_bytes, expiration_time, valid in zip(
+                    subkeys, binary_values, expirations, store_ok):
+                if valid:
+                    self.protocol.cache.store_subkey(key_id, subkey, stored_value_bytes, expiration_time)
```
justheuristic, line 385:
```suggestion
            for subkey, stored_value_bytes, expiration_time, accepted in zip(
                    subkeys, binary_values, expirations, store_ok):
                if accepted:
                    self.protocol.cache.store_subkey(key_id, subkey, stored_value_bytes, expiration_time)
```

[Optional] i would argue that valid=False does not necessarily mean that the record is invalid: peers may have refused it for a number of reasons, such as having a record with newer expiration time.

tests/test_dht_schema.py
```diff
@@ -0,0 +1,59 @@
...
+@pytest.mark.forked
+@pytest.mark.asyncio
+async def test_schema_validator():
+    validator = SchemaValidator({
+        b'experiment_name': {'type': 'binary'},
+        b'n_batches': {'type': 'dict',
+                       'keysrules': {'type': 'binary'},
+                       'valuesrules': {'type': 'integer', 'min': 0}},
+        b'signed_data': {'type': 'dict',
+                         'keysrules': {'type': 'binary', 'regex': rb'^.*\[owner:.+\].*$'}},
```
mryab, line 17: It might be possible to avoid these custom formats (which might not be familiar to users without Cerberus experience) with [pydantic](https://github.com/samuelcolvin/pydantic/) and native annotations

hivemind/dht/schema.py
```diff
@@ -0,0 +1,62 @@
...
+                  for key, value in schema.items()}
+
+        self._validator = self._ExtendedCerberusValidator(schema)
+        self._validator.allow_unknown = True
+
+    def validate(self, record: DHTRecord) -> bool:
+        value = DHTProtocol.serializer.loads(record.value)
+        if record.subkey not in DHTProtocol.RESERVED_SUBKEYS:
+            subkey = DHTProtocol.serializer.loads(record.subkey)
+            document = {record.key: {subkey: value}}
```
mryab, line 54: Feel obliged to note that `document` is not used in our codebase as a synonym for DHT records. I'd probably use something like `deserialized_record`, though I agree that it's a matter of taste

hivemind/dht/node.py
mryab: Personally, I think that this way of obtaining the expiration time and intentionally failing the check otherwise (with `"inf"`) is kinda hacky and error-prone, which allowed the bug to appear in the first place. 

What do you think about rewriting this in a more clear and explicit manner, such as 
```
cached_value = self.protocol.cache.get(key_id)
if cached_value is not None and cached_value.expiration_time <= rejected_expiration: # cache would be rejected
```

mryab (changes requested):

hivemind/dht/node.py
borzunov: Sure!

Personally, I'd argue for prohibiting the `value or default` expressions in all new code since this is known to be a error-prone anti-pattern.

borzunov:
I've replaced Cerberus to Pydantic and will re-request the review because of that.

hivemind/dht/schema.py
```diff
@@ -0,0 +1,123 @@
...
+            readable_record = {self._alias_to_name.get(key_alias, key_alias):
+                               deserialized_record[key_alias]}
+            logger.warning(f"Record {readable_record} doesn't match the schema: {e}")
+            return False
+
+        parsed_value = parsed_record.dict(by_alias=True)[key_alias]
+        if parsed_value != deserialized_record[key_alias]:
+            logger.warning(
+                f"Value {deserialized_record[key_alias]} needed type conversions to match "
+                f" the schema: {parsed_value}. The type conversions are not allowed")
```
mryab, line 90:
```suggestion
                f" the schema: {parsed_value}. Type conversions are not allowed")
```

mryab (approved):

justheuristic (approved):

PR #228, opened 2021-04-16
Add Switch Transformers-like RemoteMixtureOfExperts

mryab:


hivemind/client/switch_moe.py
```diff
@@ -0,0 +1,190 @@
...
+    :param k_best: average this many highest-scoring experts to compute activations
+    :param k_min: make sure at least this many experts returned output (i.e. didn't fail)
+    :param timeout_after_k_min: wait for this many seconds after k_min experts returned results.
+    :param detect_anomalies: whether to check input/output tensors for NaN and infinity values
+     Any expert that didn't manage to return output after that delay is considered unavailable
+    """
+
+    def __init__(self, *, grid_size: Tuple[int, ...], utilization_alpha: float = 0.9, grid_dropout: float = 1.0,
+                 jitter_eps: float = 1e-2, **kwargs):
+        super().__init__(grid_size=grid_size, **kwargs)
```
justheuristic, line 48: - perhaps we should hard-code k_best=1 k_min=0 here

justheuristic (approved):
Great job!

Summary of our conversation:
- there's still one test (the one you introduced) that needs fixing
- would be great to ensure that moe returns zeros when the expert did not return result. Maybe add this as an option (no_response=zeros|raise|nan, optional warnings, up to you)
- would be nice to hard-code k_min/k_best for switch MoE
- please update the docstrings for switch
- please include switch MoE into sphinx RTFD
- there will be a ton of performance tuning of beam search for k=1 in future PRs

hivemind/client/switch_moe.py
```diff
@@ -0,0 +1,177 @@
...
+        grid_indices = torch.zeros([len(flat_experts), len(grid_probs)], dtype=torch.int64)
+        for i, expert in enumerate(flat_experts):
+            expert_indices = expert.uid[len(self.beam_search.uid_prefix):]
+            expert_indices = list(map(int, expert_indices.split(UID_DELIMITER)))
+            grid_indices[i] = torch.as_tensor(expert_indices, dtype=grid_indices.dtype)
+
+        scores_per_dim = [
+            dim_scores[flat_batch_indices, dim_indices] if len(flat_batch_indices) else torch.zeros(0)
+            for dim_scores, dim_indices in zip(grid_probs, grid_indices.T)]
+        flat_scores = torch.prod(torch.stack(scores_per_dim, dim=0), dim=0)
```
mryab, line 173: This is the only difference from the base class, but as of now I have no clear idea of how to abstract that away

hivemind/client/switch_moe.py
```diff
@@ -0,0 +1,177 @@
...
+            expert_indices = expert.uid[len(self.beam_search.uid_prefix):]
+            expert_indices = list(map(int, expert_indices.split(UID_DELIMITER)))
+
+            for dim_index, dim_utilization in zip(expert_indices, batch_utilization):
+                dim_utilization[dim_index] += 1
+
+        return torch.cat([
+            torch.nn.functional.normalize(dim_utilization, p=1, dim=0)
+            for dim_utilization in batch_utilization
+        ])
```
mryab, line 143: Probably suboptimal, any advice on speeding that up would be welcome

justheuristic: We can probably use .scatter_add_ to avoid loops. Mind if i create an issue?

justheuristic (approved):
LGTM

PR #229, opened 2021-04-16
Statistics averaging

nevec:
This PR extends `TrainingAverager` with optimizer's stats averaging. It should serve as base for subsequent decentralized adaptive optims implementations.

- implement averaging feature
- add basic test
- move hivemind.client.optim into separate module
- fix typo, which has lead to infinite recursion


hivemind/client/__init__.py
```diff
@@ -1,4 +1,4 @@
 from hivemind.client.expert import RemoteExpert
 from hivemind.client.moe import RemoteMixtureOfExperts
 from hivemind.client.averaging import DecentralizedAverager
```
justheuristic, line 3: suggestion: import TrainingAverager on this line as well?

hivemind/client/averaging/training.py
```diff
@@ -46,7 +49,7 @@ def __init__(self, opt: torch.optim.Optimizer, *, average_parameters: bool, aver
     def step(self, wait: bool = True, **kwargs):
         """ Average optimizer weights and gradients with peers. """
         if not wait:
-            return run_in_background(self.step, wait=False, **kwargs)
+            return run_in_background(self.step, wait=True, **kwargs)
```
justheuristic, line 52: good catch!

justheuristic (approved):

tests/test_averaging.py
```diff
@@ -377,3 +377,48 @@ def test_getset_bits():
...
+        (x1 - a).pow(2).sum().backward()
+        (x2 - a).pow(2).sum().backward()
+        opt1.step()
+        opt2.step()
+
+        with torch.no_grad():
+            x_avg = 0.5 * (x1 + x2)
+            grad_avg = 0.5 * (x1.grad + x2.grad)
+            stats_avg = 0.5 * (opt1.state[x1]["exp_avg_sq"] + opt2.state[x2]["exp_avg_sq"])
+
```
borzunov, line 413:
```suggestion

    # We set wait=False to test run_in_background() usage
```

nit: Using `wait=False` and then waiting for the result looked surprising to me. I'd suggest clarifying this with the comment.

tests/test_averaging.py
```diff
@@ -377,3 +377,48 @@ def test_getset_bits():
...
+        f1 = averager1.step(wait=False)
```
borzunov, line 414:
```suggestion
        # We set wait=False to test hivemind.utils.run_in_background() usage
        f1 = averager1.step(wait=False)
```

nit: Using `wait=False` and then waiting for the result looked surprising to me. I'd suggest clarifying this with the comment.

borzunov (approved):
I approve the PR, but ask you to consider the minor change I've suggested :)

tests/test_averaging.py
nevec: Actually the main purpose  of `wait=False` is to prevent deadlock, when averager1 waits for averager2 to join. 
Fix: write this explicitly in comment

borzunov: Oh, sure, missed that! Thanks for the explanation :)

PR #230, opened 2021-04-18
Add version-consistent result rounding to load_balance_peers

justheuristic:
__Problem:__
In current version, calling load_balance_peers may result in a nasty bug: two peers with different numpy/scipy versions (or even different builds of the same version, such as mkl/nomkl) can make incompatible decisions on how to round load balancing outputs.

Here's an example from my local laptiop (numpy+atlas)
image

Here's a colab instance with __the same version of numpy+scipy, but different build and python version__
image

When training collaboratively, all AWS peers would split vector parts as:
`(4461869, 4461869, 4461868, 4461868, 0, 0, 0, 0)`

... while a some desktop peers decided:
`(4461869, 4461868, 4461868, 4461869, 0, 0, 0, 0)`

As a result, AllReduce failed with INTERNAL_ERROR due to incompatible part sizes.

__Solution:__
- rounded load balancing results up to 10^-9
- hard-code a solver method in case scipy will ever change the default solver (they did so in the past)
- added a callback for leader to solve linprog for the group


justheuristic:
[closed: this does not fully solve our problem; added issue #231 ]

mryab (approved):

PR #165, opened 2021-03-01
NAT traversal with libp2p

deniskamazur:
Our current strategy is to use [go-libp2p-daemon](https://github.com/libp2p/go-libp2p-daemon) to connect between users
* __Compile libp2p daemon__
   * [x] Build libp2p daemon on setup
      * [x] Ensure that we build a specific version of go-libp2p-daemon
   * [x] Make sure p2pd is included with pypi package
   * [x] Find out the optimal parameters for nat traversal

* __Implement a P2Pd controller (codename hivemind.P2P)__
   * [ ] Make sure it always kills the p2pd process on termination (both with del and with kill -9)
   * [x] add_rpc_servicer (simlar to GRPC)
       * [x] Ensure that it supports asyncio
       * [ ] Ensure that it is process-safe
   * [x] Add unit tests for hivemind.P2P
   * [x] Figure out the user experience (e.g. spawn P2P when DHT is created)
   * [x] Figure out the codebase layout (e.g. hivemind.dht.p2p)
      * Considerations:
          * Move synchronization and get_dht_time to P2P? (rationale: support clock correction)
          * Move dht's validate=True to P2P?
          * In future, can we merge hivemind.DHT into libp2p's DHT? (expiration, conflict resolution, atomic subkey addition, performance)

* __Convert `hivemind.dht.DHT`__
   * [x] support the equivalent of gRPC.unary_unary RPC - get one message, return one message
   * [x] hivemind.dht.DHTProtocol to p2pd
   * [ ] ensure that we can deal with stale connections
   * [x] make sure it passes `tests/test_dht_node.py`
   * [x] make sure it passes `tests/test_dht_experts.py`
   * [ ] tune performance for `tests/benchmark_dht` with 1000 nodes and 16384 experts

* __Convert `hivemind.client.averaging.DecentralizedAverager` to p2pd__
   * [ ] figure out how to implement rpc_join_group:
       * maybe support the equivalent of gRPC.stream_stream RPC (for averager)
   * [ ] ensure that large messages do not cause reduced throughput
       * [ ] test if split_into_parts is still required. If not, remove it
   * [ ] make sure it passes `tests/test_averaging.py`
   * [ ] tune performance with `tests/benchmark_averaging.py`

* __Convert `hivemind.dht.Server`__
   * [ ] find some way to attach several processes to one RPC (as in server/connection_handler.py)
   * [ ] make sure it passes `tests/test_moe.py`
   * [ ] make sure it passes `tests/test_training.py`
[...]

PR #238, opened 2021-04-20
Add initial support for connecting via libp2p

MaximKsh:


tests/test_p2p_daemon.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+@pytest.mark.asyncio
+async def test_daemon_killed_on_del():
+    p2p_daemon = await P2P.create()
+
+    child_pid = p2p_daemon._child.pid
+    assert is_process_running(child_pid)
+
+    p2p_daemon.__del__()
```
justheuristic, line 42: [note]
make sure we do not call __del__ directly to avoid double calls 
(how __del__ works: https://stackoverflow.com/a/1481512 )

hivemind/client/averaging/matchmaking.py
```diff
@@ -462,5 +462,13 @@ async def _declare_averager_periodically(self, key_manager: GroupKeyManager):
                                                        looking_for_group=False)
 
 
+def compute_schema_hash(tensors: Sequence[torch.Tensor]) -> bytes:
```
borzunov, line 465: In the master branch, `compute_schema_hash` is already defined in `hivemind/client/averaging/__init__.py`. Please either remove this change or remove the previous definition from `__init__.py`.

setup.py
```diff
@@ -2,12 +2,31 @@
 import glob
 import os
 import re
+import subprocess
+import urllib.request
+import tarfile
+import tempfile
+import hashlib
```
borzunov, line 9: nit: Please sort the imports

setup.py
```diff
@@ -28,20 +47,62 @@ def proto_compile(output_path):
...
 
 
-class ProtoCompileInstall(install):
+def libp2p_build_install():
+    try:
+        proc = subprocess.Popen(['go', 'version'],
+                                stdout=subprocess.PIPE)
+        result, _ = proc.communicate()
+        result = result.decode('ascii', 'replace')
+        _, _, v, _ = result.split(' ')
```
borzunov, line 56: This method of parsing output is not very readable and reliable. I suggest to use the regular expression like `^go version go([\d.]+)` instead.

setup.py
```diff
@@ -28,20 +47,62 @@ def proto_compile(output_path):
...
+        if version.parse(v) < version.parse("1.13"):
+            raise EnvironmentError(f'newer version of go required: must be >= 1.13, found {version}')
+
+    except FileNotFoundError:
+        raise FileNotFoundError('could not find golang installation')
+
+    with tempfile.TemporaryDirectory() as tempdir:
+        url = f'https://github.com/learning-at-home/go-libp2p-daemon/archive/refs/tags/{P2PD_VERSION}.tar.gz'
+        dest = os.path.join(tempdir, 'libp2p-daemon.tar.gz')
+        urllib.request.urlretrieve(url, os.path.join(tempdir, dest))
```
borzunov, line 68: Bug: `tempdir` is appended twice.

hivemind/p2p/p2p_daemon_bindings/keys.py
```diff
@@ -0,0 +1,91 @@
...
+@unique
+class KeyType(Enum):
+    RSA = 0
+    Ed25519 = 1
+    Secp256k1 = 2
+    ECDSA = 3
+    ECC_P256 = 4
+
+
+class Key(ABC):
```
borzunov, line 17: Ideally, it would be great to replace these classes with the interfaces of the `cryptography` module (since we already use it for RSA signatures of DHT records).

However, this may require a lot of work, so I'm okay if you won't do that in this PR.

hivemind/p2p/p2p_daemon_bindings/p2pclient.py
```diff
@@ -0,0 +1,75 @@
+from typing import AsyncIterator, Iterable, Sequence, Tuple
+
+import asyncio
+from hivemind.p2p.p2p_daemon_bindings.control import ControlClient, DaemonConnector, StreamHandler
+from contextlib import asynccontextmanager
+from multiaddr import Multiaddr
+from hivemind.p2p.p2p_daemon_bindings.datastructures import PeerInfo, StreamInfo, ID
```
borzunov, line 7:
```suggestion
import asyncio
from contextlib import asynccontextmanager
from typing import AsyncIterator, Iterable, Sequence, Tuple

from multiaddr import Multiaddr

from hivemind.p2p.p2p_daemon_bindings.control import ControlClient, DaemonConnector, StreamHandler
from hivemind.p2p.p2p_daemon_bindings.datastructures import PeerInfo, StreamInfo, ID
```

nit: I suggest to sort imports and divide it to stdlib/third-party/project-wide groups.

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,290 @@
...
+import subprocess
+import typing as tp
+
+import google.protobuf
+from multiaddr import Multiaddr
+import hivemind.p2p.p2p_daemon_bindings.p2pclient as p2pclient
+from hivemind.p2p.p2p_daemon_bindings.datastructures import ID, StreamInfo
+
+from hivemind.utils.networking import find_open_port
+from hivemind.utils.logging import get_logger
```
borzunov, line 14:
```suggestion
from multiaddr import Multiaddr

from hivemind.p2p.p2p_daemon_bindings import p2pclient
from hivemind.p2p.p2p_daemon_bindings.datastructures import ID, StreamInfo
from hivemind.utils.networking import find_open_port
from hivemind.utils.logging import get_logger
```

nit: I suggest to simplify imports and divide it to stdlib/third-party/project-wide groups.

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,186 @@
+import hashlib
+from typing import Union, List, Sequence, Any
+
+import base58
+import multihash
+
+from multiaddr import Multiaddr, protocols
+from hivemind.proto import p2pd_pb2
+
+from hivemind.p2p.p2p_daemon_bindings.keys import PublicKey
```
borzunov, line 10:
```suggestion
import multihash
from multiaddr import Multiaddr, protocols

from hivemind.p2p.p2p_daemon_bindings.keys import PublicKey
from hivemind.proto import p2pd_pb2
```

nit: I suggest to sort imports and divide it to stdlib/third-party/project-wide groups.

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,186 @@
...
+
+        def update(self, input: bytes) -> None:
+            self._digest += input
+
+        def digest(self) -> bytes:
+            return self._digest
+
+    multihash.FuncReg.register(
+        IDENTITY_MULTIHASH_CODE, "identity", hash_new=lambda: IdentityHash()
+    )
```
borzunov, line 36:
```suggestion
    multihash.FuncReg.register(
        IDENTITY_MULTIHASH_CODE, "identity", hash_new=IdentityHash)
```

nit: `lambda` seems redundant.

hivemind/proto/p2pd.proto
```diff
@@ -0,0 +1,158 @@
+syntax = "proto2";
```
borzunov, line 1: I suggest to make a docstring clarifying that this code comes from the [repo](https://github.com/mhchia/py-libp2p-daemon-bindings), mentioning the license and author name.

hivemind/p2p/p2p_daemon_bindings/p2pclient.py
```diff
@@ -0,0 +1,75 @@
+from typing import AsyncIterator, Iterable, Sequence, Tuple
```
borzunov, line 1: I suggest to make a docstring clarifying that this code comes from the [repo](https://github.com/mhchia/py-libp2p-daemon-bindings), mentioning the license and author name.

Same for every file in this directory.

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,290 @@
...
+            quic=quic, tls=tls, connManager=conn_manager,
+            dhtClient=dht_client, natPortMap=nat_port_map,
+            autonat=auto_nat, b=bootstrap, **kwargs)
+        self._assign_daemon_ports(host_port, daemon_listen_port)
+        for try_count in range(self.NUM_RETRIES):
+            try:
+                self._initialize(proc_args)
+                await self._identify_client(P2P.RETRY_DELAY * (2 ** try_count))
+            except Exception as e:
+                logger.debug(f"Failed to initialize p2p daemon: {e}", RuntimeWarning)
```
borzunov, line 70: This will fail with `TypeError: not all arguments converted during string formatting` since the `RuntimeWarning` argument is not used.

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,290 @@
...
+            return await P2P.receive_data(reader)
+        finally:
+            writer.close()
+
+    def __del__(self):
+        self._kill_child()
+
+    @property
+    def is_alive(self):
+        return self._child.is_alive
```
borzunov, line 271: AFAIU, `self._child` does not have the `is_alive` property (since it is a result of `subprocess.Popen`).

borzunov (changes requested):
Thanks for the contribution!

I didn't spend much time reading `p2p_daemon_bindings` and tests yet, but I've read everything else.

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,290 @@
...
+    @staticmethod
+    async def send_protobuf(protobuf, out_proto_type, writer):
+        if type(protobuf) != out_proto_type:
+            error = TypeError('Unary handler returned protobuf of wrong type.')
+            await P2P.send_raw_data(pickle.dumps(error), writer)
+            raise error
+        await P2P.send_raw_data(protobuf.SerializeToString(), writer)
+
+    @staticmethod
+    async def receive_exactly(reader, n_bytes, max_bytes=1 << 16):
```
borzunov, line 141: Is `reader` an asyncio stream now? If it is, we can use the native readexactly() method instead.

tests/test_p2p_daemon.py
```diff
@@ -0,0 +1,343 @@
...
+    echo "{1}"
+else
+    echo "{2}"
+fi
+'''
+
+
+def is_process_running(pid: int) -> bool:
+    cmd = CHECK_PID_CMD.format(pid, RUNNING, NOT_RUNNING)
+    return subprocess.check_output(cmd, shell=True).decode('utf-8').strip() == RUNNING
```
borzunov, line 29: In Bash, `if` just checks whether the command's return code is zero, so you can just call `ps -p {pid}` here and check its return status (without using the constant with the Bash code).

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,326 @@
...
+        self._assign_daemon_ports(host_port, daemon_listen_port)
+        self._client_listen_port = find_open_port()
+        self._client = p2pclient.Client(
+            Multiaddr(f'/ip4/127.0.0.1/tcp/{self._daemon_listen_port}'),
+            Multiaddr(f'/ip4/127.0.0.1/tcp/{self._client_listen_port}'))
+        await self._identify_client(0)
+        return self
+
+    async def wait_peers_at_least(self, peers_num, attempts=3):
+        while attempts:
```
borzunov, line 104: nit: Maybe change to `for _ in range(attempts)` to improve readabilty?

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,326 @@
...
+        self._alive = True
+        self._assign_daemon_ports(host_port, daemon_listen_port)
+        self._client_listen_port = find_open_port()
+        self._client = p2pclient.Client(
+            Multiaddr(f'/ip4/127.0.0.1/tcp/{self._daemon_listen_port}'),
+            Multiaddr(f'/ip4/127.0.0.1/tcp/{self._client_listen_port}'))
+        await self._identify_client(0)
+        return self
+
+    async def wait_peers_at_least(self, peers_num, attempts=3):
```
borzunov, line 103: I think the code `await client.wait_peers_at_least(n)` may be misleading: it is similar to the call to [asyncio.wait_for](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for) and I would expect it to wait for `n` seconds (while actually it waits for `n` peers).

I'd suggest to rename the method to `wait_for_at_least_n_peers` (the line `await client.wait_for_at_least_n_peers(n)` would not be misleading).

tests/test_p2p_daemon.py
```diff
@@ -0,0 +1,343 @@
...
+
+    await p2p_daemon.shutdown()
+    assert not is_process_running(child_pid)
+
+
+def handle_square(x):
+    return x ** 2
+
+
+def handle_add(args):
```
borzunov, line 93: This can be replaced to the built-in function `sum`.

borzunov (changes requested):

hivemind/proto/crypto.proto
```diff
@@ -0,0 +1,24 @@
+//Originally taken from: https://github.com/mhchia/py-libp2p-daemon-bindings
+//Licence: MIT
+//Author: Kevin Mai-Husan Chia
```
borzunov, line 3: Can you please clarify that this is taken from [py-libp2p](https://github.com/libp2p/py-libp2p/blob/master/libp2p/crypto/pb/crypto.proto) (not the p2p bindings themselves)?

This is not only to acknowledge the authors, but to make it clear where the code comes from and why it's written this way.

hivemind/p2p/p2p_daemon_bindings/keys.py
```diff
@@ -0,0 +1,97 @@
+"""
+Originally taken from: https://github.com/mhchia/py-libp2p-daemon-bindings
```
borzunov, line 2: Can you please specify that this is taken from [py-libp2p](https://github.com/libp2p/py-libp2p/blob/master/libp2p/crypto/keys.py) (not the p2p bindings themselves)?

This is not only to acknowledge the authors, but to make it clear where the code comes from and why it's written this way.

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,326 @@
...
+            try:
+                await self._listen_task
+            except asyncio.CancelledError:
+                self._listen_task = None
+                self._server_stopped.clear()
+
+    async def add_stream_handler(self, name, handle):
+        if self._listen_task is None:
+            self.start_listening()
+        await self._client.stream_handler(name, P2P._handle_stream(handle))
```
borzunov, line 259: Unlike the unary handlers, the stream handlers use `pickle` for serializing inputs/outputs and lead to the risk of remote code execution (see #155).

While an application is not vulnerable until a stream handler is added (and the DHT-over-libp2p implementation in #165 does not use it), I still think that the presence of such a convenient API may seduce a programmer to use it (and accidentally allow their experiment to be turned into a botnet).

Therefore, I suggest to either remove `P2P.add_stream_handler`, `P2P. _handle_stream`, and other methods using `pickle` from this PR or rewrite them with `MSGPackSerializer`.

borzunov:
I don't have any more change requests for this PR and invite @justheuristic and @mryab to go on with reviewing it :)

MaximKsh:
Fix latest comments. MessagePack serialization doesn't work with numpy arrays so I rewrite tests to torch tensors and hivemind/utils/compression.py

tests/test_p2p_daemon.py
```diff
@@ -0,0 +1,407 @@
...
+RUNNING = 'running'
+NOT_RUNNING = 'not running'
+CHECK_PID_CMD = '''
+if ps -p {0} > /dev/null;
+then
+    echo "{1}"
+else
+    echo "{2}"
+fi
+'''
```
borzunov, line 28: These 3 constants are not used anymore.

hivemind/p2p/p2p_daemon.py
```diff
@@ -148,13 +147,14 @@ async def send_raw_data(byte_str, writer):
...
-        await P2P.send_raw_data(pickle.dumps(data), writer)
+        raw_data = MSGPackSerializer.dumps(data)
+        await P2P.send_raw_data(raw_data, writer)
 
     @staticmethod
     async def send_protobuf(protobuf, out_proto_type, writer):
         if type(protobuf) != out_proto_type:
             error = TypeError('Unary handler returned protobuf of wrong type.')
-            await P2P.send_raw_data(pickle.dumps(error), writer)
+            await P2P.send_raw_data(MSGPackSerializer.dumps(error), writer)
```
borzunov, line 157: `MSGPackSerializer` won't be able to serialize the TypeError.

tests/test_p2p_daemon.py
```diff
@@ -97,6 +100,27 @@ def handle_add(args):
     return result
 
 
+def handle_square_torch(x):
+    tensor = runtime_pb2.Tensor()
+    tensor.ParseFromString(x)
+    tensor = deserialize_torch_tensor(tensor)
```
borzunov, line 95: It would be convenient to create functions that deserialize the tensor from the raw bytes (`x` here) and serialize it from the raw bytes. You may put them to `hivemind/utils/compression.py` as well.

tests/test_p2p_daemon.py
```diff
@@ -220,7 +244,7 @@ def server_target(handler_name, server_side, client_side, response_received):
 @pytest.mark.asyncio
 async def test_call_peer_different_processes():
     handler_name = "square"
-    test_input = np.random.randn(2, 3)
+    test_input = 2
```
borzunov, line 236: Does it works on tensors (without this change)?

borzunov (changes requested):
See the comments on the `MSGPackSerializer` usage above.

tests/test_p2p_daemon.py
MaximKsh: Maybe do it apart? It doesn't affect behavior

borzunov: I'm okay with doing it in a later PR :)

hivemind/p2p/p2p_daemon.py
```diff
@@ -38,6 +39,9 @@ class P2P(object):
     RETRY_DELAY = 0.4
     HEADER_LEN = 8
     BYTEORDER = 'big'
+    PB_HEADER_LEN = 1
+    RESULT_MESSAGE = int(0).to_bytes(PB_HEADER_LEN, BYTEORDER)
+    ERROR_MESSAGE = int(1).to_bytes(PB_HEADER_LEN, BYTEORDER)
```
borzunov, line 44:
```suggestion
    RESULT_MESSAGE = b'\x00'
    ERROR_MESSAGE = b'\x01'
```

hivemind/proto/p2pd.proto
```diff
@@ -160,3 +160,7 @@ message PSResponse {
   repeated string topics = 1;
   repeated bytes peerIDs = 2;
 }
+
+message P2PRPCError {
```
borzunov, line 164: nit: I suggest to rename it to just `RPCError` since everything in this file is related to p2p :)

hivemind/p2p/p2p_daemon.py
```diff
@@ -146,27 +150,41 @@ async def send_raw_data(byte_str, writer):
...
     @staticmethod
-    async def receive_raw_data(reader: asyncio.StreamReader):
-        header = await reader.readexactly(P2P.HEADER_LEN)
+    async def send_protobuf_with_error(protobuf, out_proto_type, writer):
+        if type(protobuf) != out_proto_type:
+            raise TypeError('Unary handler returned protobuf of wrong type.')
+        if out_proto_type == p2pd_pb2.P2PRPCError:
+            await P2P.send_raw_data(P2P.ERROR_MESSAGE, writer)
+        else:
+            await P2P.send_raw_data(P2P.RESULT_MESSAGE, writer)
```
borzunov, line 170:
```suggestion
            writer.write(P2P.ERROR_MESSAGE)
        else:
            writer.write(P2P.RESULT_MESSAGE)
```

I'd suggest to send one byte instead of 9 bytes (`HEADER_LEN` + the type byte) here.

hivemind/p2p/p2p_daemon.py
```diff
@@ -146,27 +150,41 @@ async def send_raw_data(byte_str, writer):
...
+
+        await P2P.send_raw_data(protobuf.SerializeToString(), writer)
+
+    @staticmethod
+    async def send_error_protobuf(protobuf, out_proto_type, writer):
```
borzunov, line 175: This method is not used.

hivemind/p2p/p2p_daemon.py
```diff
@@ -175,20 +193,32 @@ async def receive_protobuf(in_proto_type, reader):
...
+            return None, protobuf
+        else:
+            raise TypeError('invalid protobuf message type')
+
     @staticmethod
     def _handle_stream(handle):
         async def do_handle_stream(stream_info, reader, writer):
             try:
-                request = await P2P.receive_data(reader)
+                request = await P2P.receive_raw_data(reader) # receive raw data
```
borzunov, line 214: nit: The comment duplicating the code should be removed.

tests/test_p2p_daemon.py
```diff
@@ -110,6 +114,13 @@ def handle_add_torch(args):
     return serialize_torch_tensor(result).SerializeToString()
 
 
+def handle_add_torch_with_exc(args):
+    try:
+        return handle_add_torch(args)
+    except:
```
borzunov, line 120:
```suggestion
    except Exception:
```

Catching everything is a bad practice in Python (e.g. it may disrupt the work of Ctrl+C which raises `KeyboardInterrupt`).

hivemind/p2p/p2p_daemon.py
```diff
@@ -175,20 +193,32 @@ async def receive_protobuf(in_proto_type, reader):
         protobuf.ParseFromString(await P2P.receive_raw_data(reader))
         return protobuf
 
+    @staticmethod
+    async def receive_protobuf_with_error(in_proto_type, reader):
```
borzunov, line 197:
```suggestion
    async def receive_protobuf_or_error(in_proto_type, reader):
```

The current name may be misleading since the received protobuf does not always contain an error.

hivemind/p2p/p2p_daemon.py
```diff
@@ -146,27 +150,41 @@ async def send_raw_data(byte_str, writer):
...
-            error = TypeError('Unary handler returned protobuf of wrong type.')
-            await P2P.send_raw_data(MSGPackSerializer.dumps(error), writer)
-            raise error
+            raise TypeError('Unary handler returned protobuf of wrong type.')
         await P2P.send_raw_data(protobuf.SerializeToString(), writer)
 
     @staticmethod
-    async def receive_raw_data(reader: asyncio.StreamReader):
-        header = await reader.readexactly(P2P.HEADER_LEN)
+    async def send_protobuf_with_error(protobuf, out_proto_type, writer):
```
borzunov, line 164:
```suggestion
    async def send_protobuf_or_error(protobuf, out_proto_type, writer):
```

The current name may be misleading since the protobuf does not always contain an error.

borzunov (changes requested):
See the minor comments above.

hivemind/p2p/p2p_daemon.py
MaximKsh: From my point of view it's better to keep send_raw_data as the only one place where `writer.write` is called.

borzunov (approved):

justheuristic (approved):

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+            Multiaddr(f'/ip4/127.0.0.1/tcp/{self._daemon_listen_port}'),
+            Multiaddr(f'/ip4/127.0.0.1/tcp/{self._client_listen_port}'))
+
+    async def _identify_client(self, delay):
+        await asyncio.sleep(delay)
+        encoded = await self._client.identify()
+        self.id = encoded[0].to_base58()
+
+    def _assign_daemon_ports(self, host_port=None, daemon_listen_port=None):
+        self._host_port, self._daemon_listen_port = host_port, daemon_listen_port
```
mryab, line 139: Assign the attributes to the class only at the end of the method

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+        if host_port is None:
+            self._host_port = find_open_port()
+        if daemon_listen_port is None:
+            self._daemon_listen_port = find_open_port()
```
mryab, line 143: Do you need this if you're checking for equality later and `host_port` can't be None at that point?

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+            while self._daemon_listen_port == self._host_port:
+                self._daemon_listen_port = find_open_port()
+
+    @staticmethod
+    async def send_raw_data(byte_str, writer):
+        request = len(byte_str).to_bytes(P2P.HEADER_LEN, P2P.BYTEORDER) + byte_str
+        writer.write(request)
+
+    @staticmethod
+    async def send_message_pack(data, writer):
```
mryab, line 153:
```suggestion
    async def send_msgpack(data, writer):
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+                reader: asyncio.StreamReader,
+                writer: asyncio.StreamWriter) -> None:
+            try:
+                try:
+                    request = await P2P.receive_protobuf(in_proto_type, reader)
+                except asyncio.IncompleteReadError:
+                    logger.debug("Incomplete read while receiving request from peer")
+                    return
+                except google.protobuf.message.DecodeError as error:
+                    logger.warning(repr(error))
```
mryab, line 239: Use logger.exception for this

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+            try:
+                await self._listen_task
+            except asyncio.CancelledError:
+                self._listen_task = None
+                self._server_stopped.clear()
+
+    async def add_stream_handler(self, name, handle):
+        if self._listen_task is None:
+            self.start_listening()
+        await self._client.stream_handler(name, P2P._handle_stream(handle))
```
mryab, line 285:
```suggestion
        await self._client.stream_handler(name, self._handle_stream(handle))
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+from hivemind.utils.logging import get_logger
+from hivemind.utils.networking import find_open_port
+
+logger = get_logger(__name__)
+
+
+@dataclasses.dataclass(frozen=False)
+class P2PContext(object):
+    ours_id: str
+    ours_port: int
```
mryab, line 24:
```suggestion
    id: str
    port: int
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+        if msg_type == P2P.RESULT_MESSAGE:
+            protobuf = in_proto_type()
+            protobuf.ParseFromString(await P2P.receive_raw_data(reader))
+            return protobuf, None
+        elif msg_type == P2P.ERROR_MESSAGE:
+            protobuf = p2pd_pb2.RPCError()
+            protobuf.ParseFromString(await P2P.receive_raw_data(reader))
+            return None, protobuf
+        else:
+            raise TypeError('invalid protobuf message type')
```
mryab, line 203:
```suggestion
            raise TypeError('Invalid Protobuf message type')
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+            writer.close()
+
+    def __del__(self):
+        self._kill_child()
+
+    @property
+    def is_alive(self):
+        return self._alive
+
+    async def shutdown(self, timeout=None):
```
mryab, line 310: Unused argument

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+
+    @staticmethod
+    def _make_dht_mode(dht_mode):
+        if dht_mode == 'dht':
+            return {'dht': 1}
+        if dht_mode == 'dht_server':
+            return {'dhtServer': 1}
+        if dht_mode == 'dht_client':
+            return {'dhtClient': 1}
+        return {'dht': 0}
```
mryab, line 344: This function may be replaced with a constant dictionary lookup

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+
+    @staticmethod
+    def _make_force_reachability(force_reachability):
+        if force_reachability == 'public':
+            return {'forceReachabilityPublic': 1}
+        if force_reachability == 'private':
+            return {'forceReachabilityPrivate': 1}
+        return {}
```
mryab, line 352: Also create a lookup table instead of if-else sequence

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+
+import logging
+from typing import AsyncIterator, Awaitable, Callable, Dict, Iterable, Sequence, Tuple
+
+import asyncio
+from contextlib import asynccontextmanager
+from multiaddr import Multiaddr, protocols
+from hivemind.p2p.p2p_daemon_bindings.datastructures import PeerInfo, StreamInfo, ID
+from hivemind.proto import p2pd_pb2 as p2pd_pb
+from hivemind.p2p.p2p_daemon_bindings.utils import DispatchFailure, read_pbmsg_safe, write_pbmsg, raise_if_failed
```
mryab, line 15: Please reorder the imports as follows:
1) builtins
2) third-party
3) hivemind

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+        raise ValueError(
+            f"connection protocol should be only one protocol out of {supported_protos}"
+            f", maddr={maddr}"
+        )
+    return tuple(proto_cand)[0]
+
+
+class DaemonConnector:
+    control_maddr: Multiaddr
+    logger = logging.getLogger("p2pclient.DaemonConnector")
```
mryab, line 42: Switch to our utils.logging.get_logger

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+import asyncio
+from contextlib import asynccontextmanager
+from multiaddr import Multiaddr, protocols
+from hivemind.p2p.p2p_daemon_bindings.datastructures import PeerInfo, StreamInfo, ID
+from hivemind.proto import p2pd_pb2 as p2pd_pb
+from hivemind.p2p.p2p_daemon_bindings.utils import DispatchFailure, read_pbmsg_safe, write_pbmsg, raise_if_failed
+
+StreamHandler = Callable[[StreamInfo, asyncio.StreamReader, asyncio.StreamWriter], Awaitable[None]]
+
+_supported_conn_protocols = (
```
mryab, line 19: Mark this variable as a constant by uppercasing

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+)
+
+
+def parse_conn_protocol(maddr: Multiaddr) -> int:
+    proto_codes = set(proto.code for proto in maddr.protocols())
+    proto_cand = proto_codes.intersection(_supported_conn_protocols)
+    if len(proto_cand) != 1:
+        supported_protos = (
+            protocols.protocol_with_code(proto) for proto in _supported_conn_protocols
+        )
```
mryab, line 32: Can this be extracted to a constant outside the function?

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+        raise ValueError(
+            f"connection protocol should be only one protocol out of {supported_protos}"
+            f", maddr={maddr}"
+        )
+    return tuple(proto_cand)[0]
+
+
+class DaemonConnector:
+    control_maddr: Multiaddr
```
mryab, line 41: Can this attribute be moved inside `__init__`?

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+                "DaemonConnector %s opens connection to %s", self, self.control_maddr
+            )
+            return await asyncio.open_unix_connection(control_path)
+        elif proto_code == protocols.P_IP4:
+            host = self.control_maddr.value_for_protocol(protocols.P_IP4)
+            port = int(self.control_maddr.value_for_protocol(protocols.P_TCP))
+            return await asyncio.open_connection(host, port)
+        else:
+            raise ValueError(
+                f"protocol not supported: protocol={protocols.protocol_with_code(proto_code)}"
```
mryab, line 64:
```suggestion
                f"Protocol not supported: {protocols.protocol_with_code(proto_code)}"
```

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+    logger = logging.getLogger("p2pclient.DaemonConnector")
+    DEFAULT_CONTROL_MADDR = "/unix/tmp/p2pd.sock"
+
+    def __init__(self, control_maddr: Multiaddr = None) -> None:
+        if control_maddr is None:
+            control_maddr = Multiaddr(self.DEFAULT_CONTROL_MADDR)
+        self.control_maddr = control_maddr
+
+    async def open_connection(self) -> (asyncio.StreamReader, asyncio.StreamWriter):
+        proto_code = parse_conn_protocol(self.control_maddr)
```
mryab, line 51: By the way, can we check that this protocol is supported during init?

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+                f"protocol not supported: protocol={protocols.protocol_with_code(proto_code)}"
+            )
+
+
+class ControlClient:
+    listen_maddr: Multiaddr
+    daemon_connector: DaemonConnector
+    handlers: Dict[str, StreamHandler]
+    logger = logging.getLogger("p2pclient.ControlClient")
+    DEFAULT_LISTEN_MADDR = "/unix/tmp/p2pclient.sock"
```
mryab, line 73: Same thing: move logger and attributes to their respective places

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __init__(
+        self, daemon_connector: DaemonConnector, listen_maddr: Multiaddr = None
+    ) -> None:
```
mryab, line 77: Are we hitting the 120 column limit with this type annotation? Better if it's a single line

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+        if listen_maddr is None:
+            listen_maddr = Multiaddr(self.DEFAULT_LISTEN_MADDR)
```
mryab, line 79: Why not make that a default value for the argument? (assuming it's immutable)

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+        self.listen_maddr = listen_maddr
+        self.daemon_connector = daemon_connector
+        self.handlers = {}
+
+    async def _handler(self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter):
+        pb_stream_info = p2pd_pb.StreamInfo()  # type: ignore
+        await read_pbmsg_safe(reader, pb_stream_info)
+        stream_info = StreamInfo.from_pb(pb_stream_info)
+        self.logger.info("New incoming stream: %s", stream_info)
```
mryab, line 88: Maybe switch to debug here?

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -0,0 +1,217 @@
...
+                f"protocol not supported: protocol={protocols.protocol_with_code(proto_code)}"
+            )
+
+        async with server:
+            self.logger.info(
+                "DaemonConnector %s starts listening to %s", self, self.listen_maddr
+            )
+            yield self
+
+        self.logger.info("DaemonConnector %s closed", self)
```
mryab, line 118: Replace with f-strings for consistency with the rest of the repo

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+# NOTE: enabling to be interoperable w/ the Go implementation
+ENABLE_INLINING = True
+MAX_INLINE_KEY_LENGTH = 42
+
+IDENTITY_MULTIHASH_CODE = 0x00
+
+if ENABLE_INLINING:
+
+    class IdentityHash:
+        _digest: bytes
```
mryab, line 28: Is this declaration necessary?

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+
+        def digest(self) -> bytes:
+            return self._digest
+
+    multihash.FuncReg.register(
+        IDENTITY_MULTIHASH_CODE, "identity", hash_new=IdentityHash
+    )
+
+
+class ID:
```
mryab, line 44: Better to rename this to something more meaningful, like PeerID

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+    _bytes: bytes
+    _xor_id: int = None
+    _b58_str: str = None
```
mryab, line 47: Declare all attributes and their default values at init

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def __init__(self, peer_id_bytes: bytes) -> None:
+        self._bytes = peer_id_bytes
+
+    @property
+    def xor_id(self) -> int:
+        if not self._xor_id:
```
mryab, line 54: If checking for None, better do so explicitly. Furthermore, you can make a reasonable default value at init to avoid lazy evaluation

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def to_base58(self) -> str:
+        if not self._b58_str:
+            self._b58_str = base58.b58encode(self._bytes).decode()
+        return self._b58_str
+
+    def __repr__(self) -> str:
+        return f"<libp2p.peer.id.ID ({self!s})>"
+
+    __str__ = pretty = to_string = to_base58
```
mryab, line 69: Don't redefine builtins that way if possible, please :) Better to explicitly define a method

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+    def to_bytes(self) -> bytes:
+        return self._bytes
+
+    def to_base58(self) -> str:
+        if not self._b58_str:
+            self._b58_str = base58.b58encode(self._bytes).decode()
+        return self._b58_str
+
+    def __repr__(self) -> str:
+        return f"<libp2p.peer.id.ID ({self!s})>"
```
mryab, line 67:
```suggestion
        return f"<libp2p.peer.id.ID ({self.to_base58()})>"
```

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+        if not self._xor_id:
+            self._xor_id = int(sha256_digest(self._bytes).hex(), 16)
+        return self._xor_id
+
+    def to_bytes(self) -> bytes:
+        return self._bytes
+
+    def to_base58(self) -> str:
+        if not self._b58_str:
+            self._b58_str = base58.b58encode(self._bytes).decode()
```
mryab, line 63: Also avoid lazy evaluation here

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, str):
+            return self.to_base58() == other
+        elif isinstance(other, bytes):
+            return self._bytes == other
+        elif isinstance(other, ID):
+            return self._bytes == other._bytes
+        else:
+            return NotImplemented
```
mryab, line 79: Why not False?

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def __hash__(self) -> int:
+        return hash(self._bytes)
+
+    @classmethod
+    def from_base58(cls, b58_encoded_peer_id_str: str) -> "ID":
```
mryab, line 85:
```suggestion
    def from_base58(cls, base58_id: str) -> "ID":
```

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+        peer_id_bytes = base58.b58decode(b58_encoded_peer_id_str)
+        pid = ID(peer_id_bytes)
+        return pid
```
mryab, line 88:
```suggestion
        return cls(peer_id_bytes)
```

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+def sha256_digest(data: Union[str, bytes]) -> bytes:
+    if isinstance(data, str):
+        data = data.encode("utf8")
+    return hashlib.sha256(data).digest()
+
+
+class StreamInfo:
+    peer_id: ID
+    addr: Multiaddr
+    proto: str
```
mryab, line 109: Unused

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+        self.peer_id = peer_id
+        self.addr = addr
+        self.proto = proto
+
+    def __repr__(self) -> str:
+        return (
+            f"<StreamInfo peer_id={self.peer_id} addr={self.addr} proto={self.proto}>"
+        )
+
+    def to_pb(self) -> p2pd_pb2.StreamInfo:
```
mryab, line 121:
```suggestion
    def to_protobuf(self) -> p2pd_pb2.StreamInfo:
```

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+        pb_msg = p2pd_pb2.StreamInfo(
+            peer=self.peer_id.to_bytes(), addr=self.addr.to_bytes(), proto=self.proto
+        )
+        return pb_msg
+
+    @classmethod
+    def from_pb(cls, pb_msg: p2pd_pb2.StreamInfo) -> "StreamInfo":
```
mryab, line 128:
```suggestion
    def from_protobuf(cls, pb_msg: p2pd_pb2.StreamInfo) -> "StreamInfo":
```

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+        stream_info = cls(
+            peer_id=ID(pb_msg.peer), addr=Multiaddr(pb_msg.addr), proto=pb_msg.proto
+        )
+        return stream_info
+
+
+class PeerInfoLibP2P:
+    peer_id: ID
+    addrs: List[Multiaddr]
```
mryab, line 137:
```suggestion
class PeerInfo:
```

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def __init__(self, peer_id: ID, addrs: Sequence[Multiaddr]) -> None:
```
mryab, line 139: Can't this be replaced with a dataclass?

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+        return (
+            isinstance(other, PeerInfo)
+            and self.peer_id == other.peer_id
+            and self.addrs == other.addrs
+        )
+
+
+def info_from_p2p_addr(addr: Multiaddr) -> PeerInfoLibP2P:
+    if not addr:
+        raise InvalidAddrError("`addr` should not be `None`")
```
mryab, line 153: But it can't be none, since it's not Optional[Multiaddr] :)
Also, check for None explicitly

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+    p2p_part = parts[-1]
+    last_protocol_code = p2p_part.protocols()[0].code
+    if last_protocol_code != protocols.P_P2P:
+        raise InvalidAddrError(
+            f"The last protocol should be `P_P2P` instead of `{last_protocol_code}`"
+        )
+
+    # make sure the /p2p value parses as a peer.ID
+    peer_id_str: str = p2p_part.value_for_protocol(protocols.P_P2P)
+    peer_id: ID = ID.from_base58(peer_id_str)
```
mryab, line 170:
```suggestion
    peer_id = ID.from_base58(peer_id_str)
```

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # we might have received just an / p2p part, which means there's no addr.
+    if len(parts) > 1:
+        addr = Multiaddr.join(*parts[:-1])
+
+    return PeerInfo(peer_id, [addr])
+
+
+class InvalidAddrError(ValueError):
+    pass
```
mryab, line 180: Define this exception before it's used

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+
+class PeerInfo(PeerInfoLibP2P):
+    @classmethod
+    def from_pb(cls, peer_info_pb: p2pd_pb2.PeerInfo) -> PeerInfoLibP2P:
+        peer_id = ID(peer_info_pb.id)
+        addrs = [Multiaddr(addr) for addr in peer_info_pb.addrs]
+        return PeerInfo(peer_id, addrs)
+
+    def __str__(self):
+        return self.peer_id.pretty() + " " + ",".join(str(a) for a in self.addrs)
```
mryab, line 191: Use string interpolation instead of concatenation

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+    return PeerInfo(peer_id, [addr])
+
+
+class InvalidAddrError(ValueError):
+    pass
+
+
+class PeerInfo(PeerInfoLibP2P):
+    @classmethod
+    def from_pb(cls, peer_info_pb: p2pd_pb2.PeerInfo) -> PeerInfoLibP2P:
```
mryab, line 185:
```suggestion
    def from_protobuf(cls, peer_info_pb: p2pd_pb2.PeerInfo) -> PeerInfo:
```

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def __eq__(self, other: Any) -> bool:
+        return (
+            isinstance(other, PeerInfo)
+            and self.peer_id == other.peer_id
+            and self.addrs == other.addrs
+        )
+
+
+def info_from_p2p_addr(addr: Multiaddr) -> PeerInfoLibP2P:
```
mryab, line 151:
```suggestion
def info_from_p2p_addr(addr: Multiaddr) -> PeerInfo:
```

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -0,0 +1,191 @@
...
+        addr = Multiaddr.join(*parts[:-1])
+
+    return PeerInfo(peer_id, [addr])
+
+
+class InvalidAddrError(ValueError):
+    pass
+
+
+class PeerInfo(PeerInfoLibP2P):
```
mryab, line 183: Is PeerInfoLibP2P used anywhere on its own?

mryab (changes requested):

setup.py
```diff
@@ -28,20 +47,64 @@ def proto_compile(output_path):
...
+
+def libp2p_download_install():
+    install_path = os.path.join(here, 'hivemind', 'hivemind_cli')
+    binary_path = os.path.join(install_path, 'p2pd')
+    if 'p2pd' not in os.listdir(install_path) or md5(binary_path) != P2PD_CHECKSUM:
+        print('Downloading Peer to Peer Daemon')
+        url = f'https://github.com/learning-at-home/go-libp2p-daemon/releases/download/{P2PD_VERSION}/p2pd'
+        urllib.request.urlretrieve(url, binary_path)
+        os.chmod(binary_path, 0o777)
+        print(md5(binary_path))
```
mryab, line 89: Is this print statement necessary?

setup.py
```diff
@@ -28,20 +47,64 @@ def proto_compile(output_path):
             file.truncate()
 
 
-class ProtoCompileInstall(install):
+def libp2p_build_install():
+    try:
+        proc = subprocess.Popen(['go', 'version'], stdout=subprocess.PIPE)
```
mryab, line 52: Why not use `subprocess.run` and check its output?

setup.py
```diff
@@ -28,20 +47,64 @@ def proto_compile(output_path):
...
+        tar = tarfile.open(dest, 'r:gz')
+        tar.extractall(tempdir)
+        tar.close()
+
+        result = subprocess.run(['go', 'build', '-o', os.path.join(here, "hivemind", "hivemind_cli", "p2pd")],
+                                cwd=os.path.join(tempdir, f'go-libp2p-daemon-{P2PD_VERSION[1:]}', 'p2pd'))
+
+        if result.returncode:
+            raise RuntimeError('Failed to build or install libp2p-daemon:'
+                               f' exited with status code :{result.returncode}')
```
mryab, line 78:
```suggestion
                               f' exited with status code: {result.returncode}')
```

setup.py
```diff
@@ -28,20 +47,64 @@ def proto_compile(output_path):
...
+        raise FileNotFoundError('could not find golang installation')
+
+    with tempfile.TemporaryDirectory() as tempdir:
+        url = f'https://github.com/learning-at-home/go-libp2p-daemon/archive/refs/tags/{P2PD_VERSION}.tar.gz'
+        dest = os.path.join(tempdir, 'libp2p-daemon.tar.gz')
+        urllib.request.urlretrieve(url, dest)
+
+        tar = tarfile.open(dest, 'r:gz')
+        tar.extractall(tempdir)
+        tar.close()
```
mryab, line 71: Can this be used as a context manager?

setup.py
```diff
@@ -28,20 +47,64 @@ def proto_compile(output_path):
...
+        v = m.group(1)
+
+        if version.parse(v) < version.parse("1.13"):
+            raise EnvironmentError(f'newer version of go required: must be >= 1.13, found {version}')
+
+    except FileNotFoundError:
+        raise FileNotFoundError('could not find golang installation')
+
+    with tempfile.TemporaryDirectory() as tempdir:
+        url = f'https://github.com/learning-at-home/go-libp2p-daemon/archive/refs/tags/{P2PD_VERSION}.tar.gz'
```
mryab, line 65: Make this url a constant as well

setup.py
```diff
@@ -28,20 +47,64 @@ def proto_compile(output_path):
...
+        dest = os.path.join(tempdir, 'libp2p-daemon.tar.gz')
+        urllib.request.urlretrieve(url, dest)
+
+        tar = tarfile.open(dest, 'r:gz')
+        tar.extractall(tempdir)
+        tar.close()
+
+        result = subprocess.run(['go', 'build', '-o', os.path.join(here, "hivemind", "hivemind_cli", "p2pd")],
+                                cwd=os.path.join(tempdir, f'go-libp2p-daemon-{P2PD_VERSION[1:]}', 'p2pd'))
```
mryab, line 74: Also, it's more readable to use shell=True and one statement for the command instead of multiple arguments in a list

mryab (changes requested):

setup.py
borzunov: Please use `f'go build -o {shlex.quote(os.path.join(here, "hivemind", "hivemind_cli", "p2pd"))}'`.

Without the `shlex.quote`, running commands with `shell=True` may lead to a security [vulnerability](https://docs.python.org/3/library/shlex.html#shlex.quote) (maybe not in this case, but it's a good practice to write secure code everywhere since the code from one place may be unexpectedly reused in others).

setup.py
```diff
@@ -28,20 +50,61 @@ def proto_compile(output_path):
...
+        os.chmod(binary_path, 0o777)
+        if md5(binary_path) != P2PD_CHECKSUM:
+            raise RuntimeError(f'Downloaded p2pd binary from {url} does not match with md5 checksum')
+
+
+class Install(install):
     def run(self):
-        proto_compile(os.path.join(self.build_lib, 'hivemind', 'proto'))
+        libp2p_download_install()
+        # proto_compile(os.path.join(self.build_lib, 'hivemind', 'proto'))
```
borzunov, line 97: @deniskamazur Is commenting this intended?

deniskamazur: commented it out for debugging purposes, sorry

tests/test_p2p_daemon_bindings.py
```diff
@@ -0,0 +1,769 @@
...
+import os
+import subprocess
+import time
+import uuid
+from contextlib import asynccontextmanager, AsyncExitStack
+from typing import NamedTuple
+
+from google.protobuf.message import EncodeError
+from multiaddr import Multiaddr, protocols
+
```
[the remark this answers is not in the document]
MaximKsh: isort-ed

hivemind/p2p/p2p_daemon.py
MaximKsh: I rewrote this to assigning later

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+    @staticmethod
+    async def send_message_pack(data, writer):
+        raw_data = MSGPackSerializer.dumps(data)
+        await P2P.send_raw_data(raw_data, writer)
+
+    @staticmethod
+    async def send_protobuf(protobuf, out_proto_type, writer):
+        if type(protobuf) != out_proto_type:
+            raise TypeError('Unary handler returned protobuf of wrong type.')
+        await P2P.send_raw_data(protobuf.SerializeToString(), writer)
```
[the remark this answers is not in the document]
MaximKsh: I think yes, I removed send_protobuf and rename send_protobuf_or_error to send_protobuf and nothing changed

hivemind/p2p/p2p_daemon.py
```diff
@@ -0,0 +1,352 @@
...
+
+    @staticmethod
+    async def receive_message_pack(reader):
+        return MSGPackSerializer.loads(await P2P.receive_raw_data(reader))
+
+    @staticmethod
+    async def receive_protobuf(in_proto_type, reader):
+        protobuf = in_proto_type()
+        protobuf.ParseFromString(await P2P.receive_raw_data(reader))
+        return protobuf
```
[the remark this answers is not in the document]
MaximKsh: same as send

hivemind/p2p/p2p_daemon_bindings/control.py
MaximKsh: isort-ed

hivemind/p2p/p2p_daemon_bindings/control.py
MaximKsh: With default listen_maddr = Multiaddr(DEFAULT_LISTEN_MADDR) it is longer than 120

hivemind/p2p/p2p_daemon_bindings/datastructures.py
MaximKsh: I'm not sure, looks like it switching hashing type and I don't know can we change this or not

hivemind/p2p/p2p_daemon_bindings/datastructures.py
MaximKsh: move explicit assignment to init.

hivemind/p2p/p2p_daemon_bindings/datastructures.py
MaximKsh: move explicit assignment to init.

hivemind/p2p/p2p_daemon_bindings/datastructures.py
MaximKsh: I merged PeerInfo with PeerInfoLibP2P

hivemind/p2p/p2p_daemon_bindings/datastructures.py
MaximKsh: I merged PeerInfo with PeerInfoLibP2P

requirements.txt
```diff
@@ -10,4 +10,6 @@ grpcio>=1.33.2
 grpcio-tools>=1.33.2
 protobuf>=3.12.2
 configargparse>=1.2.3
+multiaddr==0.0.9
+pymultihash==0.8.2
```
mryab, line 14: We usually have restricions of the form >= to support all dependency versions starting from minimally possible, otherwise we'll have lots of unnecessary conflicts with other libraries

tests/test_p2p_daemon_bindings.py
```diff
@@ -0,0 +1,775 @@
...
+    await read_pbmsg_safe(s, pb_msg)
+    assert pb_msg.SerializeToString() == msg_bytes
+
+
+@pytest.mark.parametrize(
+    "pb_msg, msg_bytes",
+    (
+        (
+            p2pd_pb.Response(),
+            b'Z\x08\x00*V\x08\x01\x12R\n"\x12 \x03\x8d\xf5\xd4(/#\xd6\xed\xa5\x1bU\xb8s\x8c\xfa\xad\xfc{\x04\xe3\xecw\xdeK\xc9,\xfe\x9c\x00:\xc8\x12\x02\xa2\x02\x12\x08\x04\x7f\x00\x00\x01\x06\xdea\x12\x08\x04\xc0\xa8\n\x87\x06\xdea\x12\x14)\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x06\xdeb',  # noqa: E501
```
mryab, line 269: Also maybe it's a good idea to have original messages instead of raw bytes

tests/test_p2p_daemon_bindings.py
```diff
@@ -0,0 +1,775 @@
...
+        await asyncio.sleep(0.01)
+
+
+class Daemon:
+    control_maddr = None
+    proc_daemon = None
+    log_filename = ""
+    f_log = None
+    closed = None
+
```
mryab, line 376: Better to move all definitions which are not directly relevant to tests to `tests/test_utils`

tests/test_p2p_daemon_bindings.py
```diff
@@ -0,0 +1,775 @@
...
+    c = ControlClient(daemon_connector=DaemonConnector())
+    assert c.listen_maddr == Multiaddr(ControlClient.DEFAULT_LISTEN_MADDR)
+
+
+@pytest.mark.parametrize(
+    "msg_bytes",
+    (
+        b'\x08\x00"R\n"\x12 F\xec\xd3p0X\xbeT\x95p^\xc8{\xc8\x13\xa3\x9c\x84d\x0b\x1b\xbb\xa0P\x98w\xc1\xb3\x981i\x16\x12\x02\xa2\x02\x12\x08\x04\x7f\x00\x00\x01\x06\xc7\xb6\x12\x08\x04\xc0\xa8\n\x87\x06\xc7\xb6\x12\x14)\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x06\xc7\xb7',  # noqa: E501
+        b'\x08\x00"R\n"\x12 \xd0\xf0 \x9a\xc6v\xa6\xd3;\xcac|\x95\x94\xa0\xe6:\nM\xc53T\x0e\xf0\x89\x8e(\x0c\xb9\xf7\\\xa5\x12\x02\xa2\x02\x12\x08\x04\x7f\x00\x00\x01\x06\xc9%\x12\x08\x04\xc0\xa8\n\x87\x06\xc9%\x12\x14)\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x06\xc9&',  # noqa: E501
+        b'\x08\x00"R\n"\x12 \xc3\xc3\xee\x18i\x8a\xde\x13\xa9y\x905\xeb\xcb\xa4\xd07\x14\xbe\xf4\xf8\x1b\xe8[g94\x94\xe3f\x18\xa9\x12\x02\xa2\x02\x12\x08\x04\x7f\x00\x00\x01\x06\xc9`\x12\x08\x04\xc0\xa8\n\x87\x06\xc9`\x12\x14)\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x06\xc9a',  # noqa: E501
```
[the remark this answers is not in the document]
MaximKsh: This code was taken from libp2p repo, I don't have original objects initialization.

tests/test_p2p_daemon_bindings.py
MaximKsh: This code was taken from libp2p repo, I don't have original objects initialization.

hivemind/proto/crypto.proto
```diff
@@ -0,0 +1,24 @@
+//Originally taken from: https://github.com/libp2p/py-libp2p
+//Licence: MIT
+//Author: Kevin Mai-Husan Chia and others
+
+syntax = "proto2";
```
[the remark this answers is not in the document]
MaximKsh: Switching proto2 -> proto3 and removing required/optional leads to some strange network errors. I tried to debug but couldn't find the problem quickly. I suggest to keep it like this for now

tests/test_utils/__init__.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+async def connect_safe(p2pd_tuple_0, p2pd_tuple_1):
+    peer_id_1, maddrs_1 = await p2pd_tuple_1.identify()
+    await p2pd_tuple_0.connect(peer_id_1, maddrs_1)
+    await try_until_success(
+        functools.partial(
+            _check_connection, p2pd_tuple_0=p2pd_tuple_0, p2pd_tuple_1=p2pd_tuple_1
+        )
+    )
```
mryab, line 192: \n

mryab (approved):

tests/test_utils/__init__.py
justheuristic, line 192:
```suggestion
    )

```

justheuristic: [converted to suggestion below]

Issue #170, opened 2021-03-04
DecentralizedAverager per-tensor compression


justheuristic:
Currently DecentralizedAverager supports one compression option for all tensors.
That is problematic if we want to go beyond float16 compression
* [ ] allow setting different compression schemes for different tensors
* [ ] implement some general "best practice" compression scheme that adapts to tensor dimensions
* [ ] support tensor compression in load_state_from_peers
* [ ] [optional] natively support lower precision in DecentralizedAverager

PR #244, opened 2021-04-24
Compression type aware averaging

mponty:
WIP
Implements https://github.com/learning-at-home/hivemind/issues/170


justheuristic:
integrated into #272

Issue #232, opened 2021-04-19
Update quickstart


justheuristic:
In quickstart.md:
- [x] switch from calling run_server.py directly to `hivemind-server`
- [x] add reference to new tutorials (e.g. albert)

In benchmarking:
- [x] update statistics for `python benchmark_throughput.py --preset default`
- [x] add averaging benchmark
- [x] add compression benchmark

PR #245, opened 2021-04-25
Add Dockerfile, refactor tests

mryab:


docs/user/benchmarks.md
```diff
@@ -5,7 +5,7 @@ hivemind.
 
 ### Server throughput
 
-You can use [this benchmark](https://github.com/learning-at-home/hivemind/blob/master/tests/benchmark_throughput.py) to
+You can use [this benchmark](https://github.com/learning-at-home/hivemind/blob/master/benchmarks/benchmark_throughput.py) to
```
justheuristic, line 8: Thanks, this move was long overdue :)

p.s. I've added a task to re-run the benchmark numbers to https://github.com/learning-at-home/hivemind/issues/232

justheuristic (approved):
request: merge this after stability fix (which is due today), also bump version

.github/workflows/push-docker-image.yml
```diff
@@ -0,0 +1,55 @@
+name: Push to Docker Hub
+
+on:
+  push:
+    branches: [ master ]
+    tags:
+      - "*.*.*"
+  pull_request:
```
borzunov, line 8: AFAIU, the [pull_request](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request) event is triggered when the PR metainfo (comments/labels/assigned people) changes but is not triggered when new commits are pushed.

Is this the intended behavior?

borzunov (changes requested):
See the comment above.

.github/workflows/push-docker-image.yml
mryab: As you can see from the history of this PR, the event is indeed triggered on every push :) The [synchronize](https://github.community/t/what-is-a-pull-request-synchronize-event/14784/4) event is triggered on each push to the branch of PR

borzunov: Got it, thanks for the explanation!

borzunov (approved):

PR #249, opened 2021-04-26
Implement combining validators

borzunov:
This PR proposes a way to combine validators. The API should match the following requirements:

1. An application should be able to add new validators to an existing DHT instance, so that several applications (e.g. the averager and the MoE) may use one DHT instance (passed to them in the class constructors).
2. Some validator classes allow using multiple validator instances (e.g. each application may add a `SchemaValidator` with its own schema), others don't (e.g. only one `RSASignatureValidator` makes sense).
3. The validator classes should be executed in a specific order (e.g. `RSASignatureValidator` should remove the signature before `SchemaValidator` will try to deserialize a record and validate the schema).

The proposed solution is described in this [comment](https://github.com/learning-at-home/hivemind/pull/249#issuecomment-827259341).

### DIFF
--- a/hivemind/dht/__init__.py
+++ b/hivemind/dht/__init__.py
@@ -17,6 +17,7 @@ import asyncio
 import ctypes
 import multiprocessing as mp
 from concurrent.futures import ThreadPoolExecutor
+from functools import partial
 from typing import List, Optional, Sequence, Union, Callable, Awaitable, TypeVar
 
 import hivemind
@@ -195,7 +196,7 @@ class DHT(mp.Process):
                 future.set_exception(e)
 
     def set_validators_if_not_present(self, validators: Dict[str, RecordValidatorBase]) -> None:
-        self.run_coroutine(partial(self._set_validators_if_not_present, validators=validators))
+        self.run_coroutine(partial(DHT._set_validators_if_not_present, validators=validators))
 
     async def _set_validators_if_not_present(
             self, node: DHTNode, validators: Dict[str, RecordValidatorBase]) -> None:
--- a/hivemind/dht/__init__.py
+++ b/hivemind/dht/__init__.py
@@ -18,7 +18,7 @@ import ctypes
 import multiprocessing as mp
 from concurrent.futures import ThreadPoolExecutor
 from functools import partial
-from typing import List, Optional, Sequence, Union, Callable, Awaitable, TypeVar
+from typing import Iterable, List, Optional, Sequence, Union, Callable, Awaitable, TypeVar
 
 import hivemind
 from hivemind.client import RemoteExpert
@@ -51,7 +51,7 @@ class DHT(mp.Process):
 
     def __init__(self, listen_on: Endpoint = "0.0.0.0:*", initial_peers: Sequence[Endpoint] = (), *, start: bool,
                  daemon: bool = True, max_workers: Optional[int] = None, parallel_rpc: Optional[int] = None,
-                 expiration: float = 300, record_validators: List[RecordValidatorBase] = None,
+                 expiration: float = 300, record_validators: Iterable[RecordValidatorBase] = (),
                  **kwargs):
         super().__init__()
         assert not isinstance(initial_peers, str), "please specify a list/tuple of initial peers (even if there's one)"
@@ -195,16 +195,16 @@ class DHT(mp.Process):
             if not future.done():
                 future.set_exception(e)
 
-    def append_validators(self, record_validators: List[RecordValidatorBase]) -> None:
+    def add_validators(self, record_validators: Iterable[RecordValidatorBase]) -> None:
         if not self.ready.is_set():
             raise RuntimeError(
                 "Can't append new validators before the DHT process has started. "
                 "Consider adding them to the initial list via DHT.__init__(record_validators=...)")
 
-        self.run_coroutine(partial(DHT._append_validators, record_validators=record_validators))
+        self.run_coroutine(partial(DHT._add_validators, record_validators=record_validators))
 
-    async def _append_validators(
-            self, node: DHTNode, record_validators: List[RecordValidatorBase]) -> None:
+    async def _add_validators(
+            self, node: DHTNode, record_validators: Iterable[RecordValidatorBase]) -> None:
         node.protocol.record_validator.extend(record_validators)
 
     def get_visible_address(self, num_peers: Optional[int] = None, peers: Sequence[Endpoint] = ()) -> Hostname:
--- a/hivemind/dht/__init__.py
+++ b/hivemind/dht/__init__.py
@@ -23,6 +23,7 @@ import hivemind
 from hivemind.client import RemoteExpert
 from hivemind.dht.node import DHTNode, DHTID, DHTExpiration
 from hivemind.dht.routing import DHTValue, DHTKey, Subkey
+from hivemind.dht.validation import CompositeValidator
 from hivemind.utils.networking import Hostname, Endpoint, strip_port
 from hivemind.utils import MPFuture, get_logger, switch_to_uvloop, ValueWithExpiration, await_cancelled, get_dht_time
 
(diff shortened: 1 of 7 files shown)


justheuristic:
[optional suggestion] the term for adding (with overwriting) keys and values of one dictionary to another is `dict1.update(dict2)`. Perhaps it is more intuitive to rename this method to `.update_validator`. That said, i definitely do not insist and the decision is entirely yours.

borzunov:
Initially, I've decided to name it `extend` to emphasize a different behavior from `update` (the fact that it does not change the existing keys). However, now it seems to be unclear, so I'll rename it to `update_if_not_present`.

justheuristic (approved):
[i have no further suggestions, LGTM]

hivemind/dht/validation.py
```diff
@@ -52,3 +53,30 @@ def strip_value(self, record: DHTRecord) -> bytes:
...
 
         return record.value
+
+
+class CompositeValidator(RecordValidatorBase):
+    def __init__(self, validators: Dict[str, RecordValidatorBase]=None):
+        self._validators = validators if validators is not None else {}
+
+    def set_if_not_present(self, validators: Dict[str, RecordValidatorBase]) -> None:
+        for key, val in validators.items():
```
mryab, line 63: Here and below: `val` is too ambiguous (could be expanded both to value and validator), I'd write `value` explicitly

hivemind/dht/validation.py
```diff
@@ -52,3 +53,30 @@ def strip_value(self, record: DHTRecord) -> bytes:
         """
 
         return record.value
+
+
+class CompositeValidator(RecordValidatorBase):
```
mryab, line 58: TODO write a docstring explaining the mechanics of combinating validators

borzunov:
I've tried to implement a simpler scheme to combine validators. Now the applications may append them to the DHT in any order:

```python
# In the application A's code
dht.append_validators([RSASignatureValidator(), SchemaValidator(SchemaA)])

# In the application B's code
dht.append_validators([SchemaValidator(SchemaB), RSASignatureValidator()])
```

Internally, each validator class implements the `.priority` property and the `.merge_with()` policy, so the `CompositeValidator` may derive the correct order for executing validators and merge validators of different classes correctly (e.g. the OR policy for schemas).

I've also made the SchemaValidator to not reject keys that are not specified in any of the schemas (that became possible due to the OR policy mentioned above).

I will re-request the review because of these significant changes.

hivemind/dht/__init__.py
```diff
@@ -190,6 +195,18 @@ async def _run_coroutine(self, node: DHTNode, coro: Callable[[DHT, DHTNode], Awa
             if not future.done():
                 future.set_exception(e)
 
+    def append_validators(self, record_validators: List[RecordValidatorBase]) -> None:
```
mryab, line 198:
```suggestion
    def add_validators(self, record_validators: List[RecordValidatorBase]) -> None:
```

hivemind/dht/validation.py
```diff
@@ -52,3 +53,71 @@ def strip_value(self, record: DHTRecord) -> bytes:
...
+        Defines the order of applying this validator with respect to other validators.
+
+        The validators are applied:
+          - In order of increasing priority for signing a record
+          - In order of decreasing priority for validating and stripping a record
+        """
+
+        return 0
+
+    def merge_with(self, other: 'RecordValidatorBase') -> bool:
```
mryab, line 69: Shouldn't that be an abstract method?

hivemind/dht/schema.py
```diff
@@ -75,19 +76,26 @@ def validate(self, record: DHTRecord) -> bool:
...
+            logger.warning(
+                f"Record {readable_record} doesn't match the schemas: {validation_errors}")
             return False
 
         parsed_value = parsed_record.dict(by_alias=True)[key_alias]
         if parsed_value != deserialized_record[key_alias]:
             logger.warning(
                 f"Value {deserialized_record[key_alias]} needed type conversions to match "
-                f" the schema: {parsed_value}. Type conversions are not allowed")
+                f"the schema: {parsed_value}. Type conversions are not allowed")
```
mryab, line 98: What if there are two valid schemas, one of which requires conversion and one does not? Is `break` a valid course of action in that case?

hivemind/dht/schema.py
```diff
@@ -100,6 +108,14 @@ def _key_id_to_str(key_id: bytes) -> str:
 
         return binascii.hexlify(key_id).decode()
 
+    def merge_with(self, other: RecordValidatorBase) -> bool:
+        if not isinstance(other, SchemaValidator):
+            return False
+
+        self._alias_to_name.update(other._alias_to_name)
```
mryab, line 115: It is no longer a private field

hivemind/dht/validation.py
```diff
@@ -52,3 +53,71 @@ def strip_value(self, record: DHTRecord) -> bytes:
...
+          - Return False if the merging has not happened. In this case, both `self` and `other`
+            should remain unchanged. The DHT will try merging `other` to another validator or
+            add it as a separate validator (to be applied sequentially).
+        """
+
+        return False
+
+
+class CompositeValidator(RecordValidatorBase):
+    def __init__(self, validators: List[RecordValidatorBase]=None):
```
mryab, line 93: Either you're missing Optional, or the List type restriction is too narrow (why not just Iterable?): this way, you'll be able to use an empty tuple by default

hivemind/dht/validation.py
```diff
@@ -52,3 +53,71 @@ def strip_value(self, record: DHTRecord) -> bytes:
...
+        self._validators = []
+        if validators is not None:
+            self.extend(validators)
+
+    def extend(self, validators: List[RecordValidatorBase]) -> None:
```
mryab, line 98:
```suggestion
    def extend(self, validators: List[RecordValidatorBase]):
```

hivemind/dht/validation.py
```diff
@@ -52,3 +53,71 @@ def strip_value(self, record: DHTRecord) -> bytes:
...
+        for new_validator in validators:
+            for existing_validator in self._validators:
+                if existing_validator.merge_with(new_validator):
+                    break
+            else:
```
mryab, line 103: Can we encounter a situation in which a validator conflicts with existing validators and can't be added at all?

mryab (changes requested):

hivemind/dht/validation.py
borzunov: Sorry, I was relying on the older version of [PEP 484](https://stackoverflow.com/a/62732458) that made arguments with the `None` default `Optional` automatically and did not know this feature is now deprecated :)

hivemind/dht/validation.py
borzunov: I think having this default behavior (a validator refuses to merge with anything and is always applied sequentially) is reasonable for new validators.

hivemind/dht/schema.py
borzunov: In strictly typed languages, accessing a private field of another instance of the same class is valid (e.g. in C++, see "Private member access" [here](https://en.cppreference.com/w/cpp/language/access)), so I think it's reasonable to assume the same in Python.

hivemind/dht/validation.py
borzunov: Why is that?

I thought the [convention](https://github.com/python/typing/issues/65#issuecomment-96904050) is that this means no annotation, and to add one we should explicitly specify `-> None` (like we do in other part of the project, e.g. [here](https://github.com/learning-at-home/hivemind/blob/master/hivemind/dht/__init__.py#L65)).

hivemind/dht/schema.py
borzunov: Fixed this case (it is possible only if two schemas have different annotations for the same key, now we allow both of them).

hivemind/dht/validation.py
borzunov: I have not made it abstract to provide a reasonable default behavior (new validators are applied sequentially by default and refuse to merge with anything).

hivemind/dht/validation.py
borzunov: I can't imagine such a situation.

One possibility is that several validators narrow down the set of possible records to the empty set, but I think this is a problem a user should solve themself (it would be correct from the POV of validator logic).

hivemind/dht/schema.py
```diff
@@ -75,21 +76,28 @@ def validate(self, record: DHTRecord) -> bool:
...
+                validation_errors.append(ValueError(
+                    f"Value {deserialized_record[key_alias]} needed type conversions to match "
+                    f"the schema: {parsed_value}. Type conversions are not allowed"))
+            else:
+                return True
+
+        readable_record = {self._alias_to_name.get(key_alias, key_alias):
+                           deserialized_record[key_alias]}
+        logger.warning(
+            f"Record {readable_record} doesn't match with all schemas: {validation_errors}")
```
mryab, line 99:
```suggestion
            f"Record {readable_record} doesn't match any of the schemas: {validation_errors}")
```

hivemind/dht/schema.py
mryab: Agreed!

hivemind/dht/validation.py
mryab: Got it, thanks!

mryab (approved):

tests/test_dht_validation.py
```diff
@@ -0,0 +1,93 @@
...
+                       expiration_time=hivemind.get_dht_time() + 10)
+
+    signed_record = dataclasses.replace(record, value=validator.sign_value(record))
+    assert signed_record.value.count(b'[signature:') == 0
+    # Expect failed validation since `unknown_key` is not a part of any schema
+    assert not validator.validate(signed_record)
+
+
+@pytest.mark.forked
+def test_dht_set_validators_if_not_present(validators_for_app):
```
mryab, line 68:
```suggestion
def test_dht_add_validators(validators_for_app):
```

borzunov:
@justheuristic has said that I can merge this :)

Issue #219, opened 2021-04-14
Tutorial: ALBERT-large collaborative training


justheuristic:
Let's implement a basic example for collaborative training with ALBERT
- core training code ( @leshanbog )
  - [x] implement basic training scripts (run_first_peer/run_trainer) based on mryab/collaborative-training 
  - [x] achieve exact match with old training code
  - [x] test fault tolerance against common network failures

- update metric logging code (@yhn112 )
   - [x] tune first peer's wandb to avoid crashes (or restart on crashes)
   - [x] use the same DHT key prefix for metrics and averaging (aka self.prefix)
   - [x] make the prototype pep8-compliant

- add basic security layer: (@borzunov )
   - [x] protect value types in:
      - [x] progress
      - [x] averaging
      - [x] metrics
   - [x] ensure that the averager validate reasonable min/max values (i.e. for batches_processed)
   - [x] make sure it supports DataParallel on peers
   - [x] make sure it will save/load scheduler state dict correctly with optimizer

- add full description in README.md

justheuristic:
Current progress by @leshanbog : https://github.com/learning-at-home/hivemind/tree/collaborative_albert_example

yhn112:
Addition to my part:
Make wandb optional
Better console logging on the coordinator
Log information about individual peers' status to wandb. Needed for debug.

justheuristic:
Addition to addition to your part:
- [ ] modify performance and num peers to count all peers that are currently alive, regardless of step

justheuristic:
It appears that everything in the issue description is done

There are also 3 additions/feature-requests aboveby @yhn112 . Mike, do you consider them solved?

yhn112:
Solved

PR #250, opened 2021-04-29
Protect training progress and metrics with signatures and DHT schema validation

borzunov:
This PR follows #219 and implements:

1. Validating DHT schema for training progress and metrics
2. Signing the training progress and the metrics (so a peer can't change the stats of someone else)
3. Using the local public key as a replacement for the peer's UUID
4. Using type-validated data structures for the stats in the code instead of plain lists

I have tested the following:

1. Training (`run_first_peer.py` and `run_trainer.py`) works without errors.
2. When connecting to the DHT from the Python interactive console, the DHT does not accept progress/metrics records that:
    - Doesn't match the schema.
    - Doesn't have a signature or have an invalid one.

Along the way, it improves the validator system:

1. Makes `RSASignatureValidator` and `SchemaValidator` picklable because their instances may be sent to the DHT in another process (this happens when we add validators to the existing DHT process after its creation).
1. Implements caching of the private key while creating `RSASignatureValidator`, so it is not generated again for each DHT-using entity inside one process (saves time because the generation takes ~100 ms).
1. Implements the `BytesWithPublicKey` type for DHT schemas.
1. Adds `prefix` parameter to `SchemaValidator` (if present, adds `prefix + '_'` to all field names) and moves them from the `bytes` to `str` keys (since the rest of code uses `str` keys).
1. Refactors `RSASignatureValidator` (e.g. better naming like `signature_validator.ownership_marker` -> `signature_validator.local_public_key`).

### DIFF
--- a/examples/albert/metrics_utils.py
+++ b/examples/albert/metrics_utils.py
@@ -1,7 +1,7 @@
 from typing import Dict, List, Tuple
 
 from hivemind.dht.crypto import RSASignatureValidator
-from hivemind.dht.schema import SchemaValidator
+from hivemind.dht.schema import BytesWithPublicKey, SchemaValidator
 from hivemind.dht.validation import RecordValidatorBase
 from pydantic import BaseModel, StrictFloat, confloat, conint
 
--- a/examples/albert/metrics_utils.py
+++ b/examples/albert/metrics_utils.py
@@ -1,0 +1,18 @@
+from hivemind.dht.schema import SchemaValidator
+from pydantic import BaseModel, StrictFloat, UUID4, confloat, conint
+
+
+class LocalMetrics(BaseModel):
+    step: conint(ge=0, strict=True)
+    samples_per_second: confloat(ge=0.0, strict=True)
+    samples_accumulated: conint(ge=0, strict=True)
+    loss: StrictFloat
+    mini_steps: conint(ge=0, strict=True)
+
+
+class MetricSchema(BaseModel):
+    metrics: Dict[UUID4, LocalMetrics]
+
+
+def make_schema_validator(experiment_prefix: str) -> SchemaValidator:
+    return SchemaValidator(MetricSchema, prefix=experiment_prefix)
--- a/examples/albert/metrics_utils.py
+++ b/examples/albert/metrics_utils.py
@@ -1,5 +1,9 @@
+from typing import Dict, List, Tuple
+
+from hivemind.dht.crypto import RSASignatureValidator
 from hivemind.dht.schema import SchemaValidator
-from pydantic import BaseModel, StrictFloat, UUID4, confloat, conint
+from hivemind.dht.validation import RecordValidatorBase
+from pydantic import BaseModel, StrictFloat, confloat, conint
 
 
 class LocalMetrics(BaseModel):
@@ -11,8 +15,11 @@ class LocalMetrics(BaseModel):
 
 
 class MetricSchema(BaseModel):
-    metrics: Dict[UUID4, LocalMetrics]
+    metrics: Dict[BytesWithPublicKey, LocalMetrics]
 
 
-def make_schema_validator(experiment_prefix: str) -> SchemaValidator:
-    return SchemaValidator(MetricSchema, prefix=experiment_prefix)
+def make_validators(experiment_prefix: str) -> Tuple[List[RecordValidatorBase], bytes]:
+    signature_validator = RSASignatureValidator()
+    validators = [SchemaValidator(MetricSchema, prefix=experiment_prefix),
+                  signature_validator]
+    return validators, signature_validator.local_public_key
--- a/examples/albert/run_trainer.py
+++ b/examples/albert/run_trainer.py
@@ -2,13 +2,15 @@
 
 import logging
 import os
+import uuid
 from dataclasses import dataclass, field, asdict
 from pathlib import Path
 from typing import Optional, Dict, Any, List
-import uuid
 
-from datasets import load_from_disk
+import hivemind
+import torch
 import transformers
+from datasets import load_from_disk
 from torch.utils.data import DataLoader
 from transformers import (set_seed, HfArgumentParser, TrainingArguments,
                           DataCollatorForLanguageModeling, AlbertTokenizerFast, AlbertConfig, AlbertForPreTraining)
@@ -16,9 +18,9 @@ from transformers.optimization import get_linear_schedule_with_warmup
 from transformers.trainer_utils import is_main_process
 from transformers.trainer import Trainer
 from torch_optimizer import Lamb
-import torch
 
-import hivemind
+import metrics_utils
+
 
 logger = logging.getLogger(__name__)
 LRSchedulerBase = getattr(torch.optim.lr_scheduler, '_LRScheduler', None)
@@ -190,15 +192,18 @@ class CollaborativeCallback(transformers.TrainerCallback):
             if self.collaborative_optimizer.local_step != self.last_reported_collaboration_step:
                 self.last_reported_collaboration_step = self.collaborative_optimizer.local_step
 
-                statistics = [self.collaborative_optimizer.local_step,
-                              self.collaborative_optimizer.performance_ema.samples_per_second,
-                              self.samples,
-                              self.loss,
-                              self.steps]
+                samples_per_second = self.collaborative_optimizer.performance_ema.samples_per_second
+                statistics = metrics_utils.LocalMetrics(
+                    step=self.collaborative_optimizer.local_step,
+                    samples_per_second=samples_per_second,
+                    samples_accumulated=self.samples,
+                    loss=self.loss,
+                    mini_steps=self.steps)
                 self.loss = 0
                 self.steps = 0
-                self.dht.store(self.collaborative_optimizer.prefix + "_metrics", subkey=self.trainer_uuid,
-                               value=statistics, expiration_time=hivemind.get_dht_time() + self.statistics_expiration,
+                self.dht.store(key=self.collaborative_optimizer.prefix + "_metrics",
+                               subkey=self.trainer_uuid, value=statistics.dict(),
+                               expiration_time=hivemind.get_dht_time() + self.statistics_expiration,
                                return_future=True)
         self.samples = self.collaborative_optimizer.local_samples_accumulated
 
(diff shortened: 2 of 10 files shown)


examples/albert/run_trainer.py
```diff
@@ -165,11 +165,12 @@ def get_optimizer_and_scheduler(training_args, model):
 
 class CollaborativeCallback(transformers.TrainerCallback):
     def __init__(self, dht: hivemind.DHT, optimizer: hivemind.CollaborativeOptimizer,
-                 model: torch.nn.Module, trainer_uuid: str, statistics_expiration: float):
+                 model: torch.nn.Module, local_public_key: bytes, statistics_expiration: float):
```
mryab, line 168:
```suggestion
                 model: torch.nn.Module, public_key: bytes, statistics_expiration: float):
```

mryab: In fact, I think it's safe to replace all occurrences of local_public_key with public_key as long as we're talking about attributes of objects (and arguments of their initialization methods)

examples/albert/metrics_utils.py
```diff
@@ -0,0 +1,25 @@
...
+    loss: StrictFloat
+    mini_steps: conint(ge=0, strict=True)
+
+
+class MetricSchema(BaseModel):
+    metrics: Dict[BytesWithPublicKey, LocalMetrics]
+
+
+def make_validators(experiment_prefix: str) -> Tuple[List[RecordValidatorBase], bytes]:
+    signature_validator = RSASignatureValidator()
```
mryab, line 22: Will it generate a new key pair after restarting an experiment on the same peer?

hivemind/dht/crypto.py
```diff
@@ -22,43 +22,55 @@ class RSASignatureValidator(RecordValidatorBase):
     the corresponding private key (so only the owner can change them).
     """
 
-    def __init__(self,
-                 marker_format: bytes=b'[owner:_key_]',
-                 signature_format: bytes=b'[signature:_value_]'):
-        self._marker_re = re.compile(re.escape(marker_format).replace(b'_key_', rb'(.+?)'))
+    PUBLIC_KEY_FORMAT = b'[owner:_key_]'
+    SIGNATURE_FORMAT = b'[signature:_value_]'
```
mryab, line 26: Also, the docstring probably needs to be updated, since the format is constant now

hivemind/dht/crypto.py
```diff
@@ -22,43 +22,55 @@ class RSASignatureValidator(RecordValidatorBase):
...
+    def _init_signature_params(self) -> None:
         self._padding = padding.PSS(mgf=padding.MGF1(hashes.SHA256()),
                                     salt_length=padding.PSS.MAX_LENGTH)
         self._hash_algorithm = hashes.SHA256()
 
     @property
-    def ownership_marker(self) -> bytes:
-        return self._ownership_marker
+    def local_public_key(self) -> bytes:
+        return self._local_public_key
```
mryab, line 57: Maybe just `public_key`

hivemind/optim/collaborative.py
```diff
@@ -37,6 +42,19 @@ def register_step(self, local_step: int):
         self.eta_next_step = float('inf')
 
 
+class LocalTrainingProgress(BaseModel):
```
mryab, line 45:
```suggestion
class TrainingState(BaseModel):
```

hivemind/optim/collaborative.py
```diff
@@ -263,12 +287,18 @@ def report_training_progress(self):
...
-
-            assert self.is_valid_peer_state(local_state_info), local_state_info
-            self.dht.store(self.training_progress_key, subkey=self.averager.endpoint, value=local_state_info,
-                           expiration_time=current_time + self.metadata_expiration, return_future=True)
+                local_state_info = LocalTrainingProgress(
+                    endpoint=self.averager.endpoint,
+                    step=self.local_step,
+                    samples_accumulated=self.local_samples_accumulated,
+                    samples_per_second=self.performance_ema.samples_per_second,
+                    time=get_dht_time(),
```
mryab, line 295: Is it possible to use current_time?

hivemind/optim/collaborative.py
```diff
@@ -296,24 +326,26 @@ def fetch_collaboration_state(self) -> CollaborationState:
...
-            total_samples_per_second += samples_per_second
-            if opt_step == global_optimizer_step:
-                total_samples_accumulated += samples_accumulated
-                estimated_current_samples += samples_accumulated + max(0, current_time - timestep) * samples_per_second
+        for state in valid_peer_states:
+            total_samples_per_second += state.samples_per_second
+            if state.step == global_optimizer_step:
+                total_samples_accumulated += state.samples_accumulated
+                estimated_current_samples += (
+                    state.samples_accumulated +
```
mryab, line 347: Can be contracted to a single line

tests/test_dht_crypto.py
```diff
@@ -41,3 +45,88 @@ def test_rsa_signature_validator():
...
+    first_validator = RSASignatureValidator()
+    second_validator = RSASignatureValidator()
+    assert first_validator.local_public_key == second_validator.local_public_key
+
+    third_validator = RSASignatureValidator(ignore_cached_key=True)
+    assert first_validator.local_public_key != third_validator.local_public_key
+
+
+def test_validator_instance_is_pickable():
+    # Needs to be pickable because the validator instance may be sent between processes
```
mryab, line 60:
```suggestion
def test_validator_instance_is_picklable():
    # Needs to be picklable because the validator instance may be sent between processes
```

mryab: Also check all other occurences

mryab (changes requested):

examples/albert/metrics_utils.py
borzunov: Yes. However, the public key here replaces the trainer UUID that was also randomly regenerated after each restart.

Do you think having a constant key pair may be useful? I imagine cases like implementing a leaderboard, however it may be a better idea to aggregate peers with different key pairs by endpoint/login at this stage (since one user may want to run several trainers).

hivemind/dht/crypto.py
borzunov: I'm afraid that removing `local_` will make naming misleading.

For example, in this class, we would have `self._public_key` storing the _local_ public key and `self._public_key_re` matching _all possible_ public keys, but the naming may suggest that `self._public_key_re` is just a regex to find the _local_ public key as well.

Also, I think it will get worse after implementing #253 that operates with a number of key pairs (`auth_server_public_key`, `service_public_key`, `client_public_key`). Therefore, I'd suggest to clarify whose public key we're storing everywhere.

examples/albert/run_trainer.py
borzunov: See the [discussion](https://github.com/learning-at-home/hivemind/pull/250#discussion_r625110139) below.

hivemind/optim/collaborative.py
```diff
@@ -263,12 +287,18 @@ def report_training_progress(self):
...
-
-            assert self.is_valid_peer_state(local_state_info), local_state_info
-            self.dht.store(self.training_progress_key, subkey=self.averager.endpoint, value=local_state_info,
-                           expiration_time=current_time + self.metadata_expiration, return_future=True)
+                local_state_info = TrainingState(
+                    endpoint=self.averager.endpoint,
+                    step=self.local_step,
+                    samples_accumulated=self.local_samples_accumulated,
+                    samples_per_second=self.performance_ema.samples_per_second,
+                    current_time=get_dht_time(),
```
mryab, line 295: I meant that it's possible to use the value of the variable `current_time` in L289, since it already stores the recent value of `get_dht_time` :) `time` as a class field name was perfectly fine, it's probably best to keep it as is

examples/albert/metrics_utils.py
mryab: I think it's reasonable to assume that a single user shouldn't have to generate new key pairs for each run just for convenience. However, we can probably leave it as is for now if nobody objects

hivemind/dht/crypto.py
```diff
@@ -16,49 +16,61 @@
...
-                 signature_format: bytes=b'[signature:_value_]'):
-        self._marker_re = re.compile(re.escape(marker_format).replace(b'_key_', rb'(.+?)'))
+    PUBLIC_KEY_FORMAT = b'[owner:_key_]'
+    SIGNATURE_FORMAT = b'[signature:_value_]'
 
-        self._signature_format = signature_format
-        self._signature_re = re.compile(re.escape(signature_format).replace(b'_value_', rb'(.+?)'))
+    PUBLIC_KEY_REGEX = re.escape(PUBLIC_KEY_FORMAT).replace(b'_key_', rb'(.+?)')
+    _public_key_re = re.compile(PUBLIC_KEY_REGEX)
+    _signature_re = re.compile(re.escape(SIGNATURE_FORMAT).replace(b'_value_', rb'(.+?)'))
```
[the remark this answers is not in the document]
borzunov: They are, I'll change the names to uppercase.

hivemind/optim/collaborative.py
borzunov: Got it, I've changed the names back :)

mryab (approved):

hivemind/dht/crypto.py
```diff
@@ -16,49 +16,61 @@
...
-            logger.debug(f"Key and subkey can't contain different public keys in {record}")
+            logger.warning(f"Key and subkey can't contain different public keys in {record}")
             return False
         public_key = serialization.load_ssh_public_key(public_keys[0])
 
-        signatures = self._signature_re.findall(record.value)
+        signatures = self._SIGNATURE_RE.findall(record.value)
         if len(signatures) != 1:
-            logger.debug(f"Record should have exactly one signature in {record}")
+            logger.warning(f"Record should have exactly one signature in {record}")
```
justheuristic, line 73: This error can be misleading to the user: if another peer writes invalid data, all other peers will see this error.
The problem is, they have nothing to do with it.

At the same time, i can't imagine a situation where a normal peer in non-debug mode should benefit from this message.

I would suggest reverting this to logger.debug.

justheuristic (approved):

hivemind/dht/crypto.py
borzunov: You're right, I've switched it back to `logger.debug` here and in SchemaValidator.

PR #254, opened 2021-05-06
Fix random freezes in averager.step, improve error handling

justheuristic:
This patch introduces three changes:
- [x] fix a heisenbug where `DecentralizedAverager` would randomly hang on pytorch ops: https://github.com/pytorch/pytorch/issues/17199
- [x] tweak "sanity check failed" clause in `DecentralizedAverager._step` so that it is no longer triggered by sanctioned retries
- [x] tweak `Matchmaking.request_join_group` to handle RPC errors instead of cancelling the entire matchmaking

### DIFF
--- a/hivemind/client/averaging/matchmaking.py
+++ b/hivemind/client/averaging/matchmaking.py
@@ -10,7 +10,7 @@ import concurrent.futures
 import asyncio
 
 import grpc
-from grpc._cython.cygrpc import InternalError
+import grpc._cython.cygrpc
 
 from hivemind.client.averaging.group_info import GroupInfo
 from hivemind.client.averaging.key_manager import GroupKeyManager, GroupKey
@@ -200,7 +200,7 @@ class Matchmaking(averaging_pb2_grpc.DecentralizedAveragingServicer):
             if call is not None:
                 call.cancel()
             return None
-        except (grpc.RpcError, grpc.aio.AioRpcError, InternalError, StopAsyncIteration) as e:
+        except (grpc.RpcError, grpc.aio.AioRpcError, grpc._cython.cygrpc.InternalError, StopAsyncIteration) as e:
             logger.warning(f"{self} - failed to request potential leader {leader}: {e}")
             return None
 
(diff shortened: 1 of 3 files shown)


justheuristic:
So, here's how this might explain the error reported earlier by @nevec : 
- initializing small tensors with ones/zeros -> no hang
- initializing large tensors with ones/zeros -> averager hangs
- initializing large tensors with empty -> no hang
- initializing large tensors with randn -> no hang

Maybe
- large ones/zeros are assigned with multiple threads, which triggers OMP
- small ones/zeros contend with a single thread -> no OMP
- empty needs no item assignment, just malloc -> no OMP
- randn uses RNG which is is inherently sequential -> no OMP

mryab (approved):

hivemind/client/averaging/matchmaking.py
```diff
@@ -199,6 +200,10 @@ async def request_join_group(self, leader: Endpoint, expiration_time: DHTExpirat
             if call is not None:
                 call.cancel()
             return None
+        except (grpc.RpcError, grpc.aio.AioRpcError, InternalError, StopAsyncIteration) as e:
+            logger.warning(f"{self} - failed to request potential leader {leader}: {e}")
```
mryab, line 204:
```suggestion
            logger.exception(f"{self} - failed to request potential leader {leader}: {e}")
```

hivemind/client/averaging/matchmaking.py
```diff
@@ -10,6 +10,7 @@
 import asyncio
 
 import grpc
+from grpc._cython.cygrpc import InternalError
```
borzunov, line 13:
```suggestion
```

nit: `InternalError` looks like a built-in error, so I suggest to remove this line and use the full name `grpc._cython.cygrpc.InternalError` instead.

borzunov (approved):

hivemind/client/averaging/matchmaking.py
justheuristic: fixed it, thanks

PR #259, opened 2021-05-10
Compression type aware avg upd

mponty:
WIP
See https://github.com/mponty/hivemind/commit/34b103f3a15867b6644176b922ad683cc0416b78


justheuristic:
Current state: works for single-tensor lists, but fails for multi-tensor lists

justheuristic:
integrated into #272

PR #261, opened 2021-05-12
Improve error handling, remove deprecated functionality

justheuristic:
- [x] modified .load_state_from_peers to handle network errors and ensure mpfuture is always set
- [x] removed dht.get/declare_experts from library, tests, benchmarks and examples
- [x] removed default `expiration=...` parameter in hivemind.DHT
- [x] removed stale TODOs and notifications
- [x] rollback Runtime.stop to a shutdown pipe -> server no longer hangs on shutdown (also: repaired benchmark_throughput)

__Note on `Runtime.stop`:__ the reason it didn't work was that when you trigger the server to shut down, it was usually awaiting for one of the pools in the selector ([source](https://github.com/learning-at-home/hivemind/blob/master/hivemind/server/runtime.py#L119)). Since new samples would never come, selector would never get triggered, the loop would never check on `self.stop` and server would be left in a deadlock.

For the same reason, `benchmark_throughput` would hang before printing results since #256 .

__Note 2:__ other applications of `.stop` work correctly because they do not await anything inside of `while self.stop.wait(...)` loop.

__master:__ (killed after hanging for several minutes)
image

__after patch:__
image

### DIFF
--- a/hivemind/client/averaging/__init__.py
+++ b/hivemind/client/averaging/__init__.py
@@ -464,7 +464,7 @@ class DecentralizedAverager(mp.Process, averaging_pb2_grpc.DecentralizedAveragin
 
         finally:
             if not future.done():
-                logger.warning("Averager could not load state from peers: all peers failed.")
+                logger.warning("Averager could not load state from peers: all requests have failed.")
                 future.set_result(None)
 
     def get_group_bits(self, wait: bool = True):
--- a/hivemind/client/averaging/__init__.py
+++ b/hivemind/client/averaging/__init__.py
@@ -422,47 +422,49 @@ class DecentralizedAverager(mp.Process, averaging_pb2_grpc.DecentralizedAveragin
         return future.result() if wait else future
 
     async def _load_state_from_peers(self, future: MPFuture):
-        key_manager = self._matchmaking.group_key_manager
-        peer_priority, _ = self.dht.get(f"{key_manager.prefix}.all_averagers", latest=True) or ({}, None)
-        peer_priority = {peer: float(info.value) for peer, info in peer_priority.items()
-                         if isinstance(info, ValueWithExpiration) and isinstance(info.value, (float, int))}
+        try:
+            key_manager = self._matchmaking.group_key_manager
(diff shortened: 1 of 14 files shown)


hivemind/client/averaging/__init__.py
```diff
@@ -183,7 +183,7 @@ async def _run():
                     self._port.value = found_port
                     await server.start()
                 else:
-                    logger.info(f"The averager running in an experimental client mode, please report any bugs.")
+                    logger.debug(f"The averager is running in client mode.")
```
justheuristic, line 186: after 5+ successful runs, i'd say this is no longer experimental

hivemind/server/task_pool.py
```diff
@@ -173,6 +173,7 @@ def _pool_input_loop(self, pending_batches: Dict[Any, List[Task]], *args, **kwar
             batch_inputs = [inp.detach().requires_grad_(inp.requires_grad).share_memory_() for inp in batch_inputs]
 
             logger.debug(f"{self.name}, batch {batch_index}: sending to runtime")
+            print(f"{self.name}, batch {batch_index}: sending to runtime")
```
mryab, line 176: del

hivemind/client/averaging/__init__.py
```diff
@@ -422,47 +422,50 @@ def load_state_from_peers(self, wait=True) -> Optional[Tuple[Any, Sequence[torch
...
-                    logger.info(f"Failed to download state from {peer} - {e}")
-                finally:
-                    if stream is not None:
-                        await stream.code()
+                        logger.info(f"Finished downloading state from {peer}")
+                        future.set_result((metadata, tensors))
+                        self.last_updated = get_dht_time()
+                        return
+                    except BaseException as e:
+                        logger.warning(f"Failed to download state from {peer} - {repr(e)}")
```
mryab, line 460:
```suggestion
                        logger.exception(f"Failed to download state from {peer} - {repr(e)}")
```

mryab (changes requested):

mryab (approved):

PR #265, opened 2021-05-18
Log correct loss in examples/albert/run_first_peer.py

borzunov:
2314e7ebd54852190cbaca867bc58407a6c7596f changed the formula for the current loss, but it was not changed in the `logger.info` call. Therefore, the coordinator logged inadequate loss values.

Screen Shot 2021-05-18 at 6 41 15 PM


justheuristic (approved):
Note: it appears that this change has already been introduced in https://github.com/mryab/collaborative-training/

yhn112 (approved):

yhn112:
> Note: it appears that this change has already been introduced in https://github.com/mryab/collaborative-training/

No problem, merging it still looks possible =)

PR #270, opened 2021-06-11
Disable p2pd compilation by default

yhn112:
After these changes, both `pip install hivemind` and `pip install -e hivemind` will use precompiled binary p2pd, and will not require Go installation.
To run with building it from source, one should use `pip install hivemind --install-option="--buildgo"` (with `-e`, optionally).

Also, fixes a minor bug with the exception in case of no Go installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,7 +105,7 @@ class Install(install):
 
 
 class Develop(develop):
-    user_options = install.user_options + [('buildgo', None, None)]
+    user_options = develop.user_options + [('buildgo', None, None)]
 
     def initialize_options(self):
         super().initialize_options()
--- a/setup.py
+++ b/setup.py
@@ -86,7 +86,7 @@ def libp2p_download_install():
 
 
 class Install(install):
-    user_options = install.user_options + [('buildgo', None, None)]
+    user_options = install.user_options + [('buildgo', None, "Builds p2pd from source")]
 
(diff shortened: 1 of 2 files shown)


justheuristic (approved):
Good job!

deniskamazur (approved):

setup.py
```diff
@@ -90,15 +86,39 @@ def libp2p_download_install():
...
 
 class Install(install):
+    user_options = install.user_options + [('buildgo', None, None)]
+
+    def initialize_options(self):
+        super().initialize_options()
+        self.buildgo = False
+
+    def finalize_options(self):
+        super().finalize_options()
```
mryab, line 96: del

setup.py
```diff
@@ -90,15 +86,39 @@ def libp2p_download_install():
 
 
 class Install(install):
+    user_options = install.user_options + [('buildgo', None, None)]
```
mryab, line 89:
```suggestion
    user_options = install.user_options + [('build_go', None, None)]
```

mryab (changes requested):

setup.py
yhn112: Fixed

setup.py
yhn112: Impossible to apply

mryab (approved):

PR #271, opened 2021-06-13
Measure testing coverage on pull request

yhn112:


.circleci/config.yml
```diff
@@ -77,8 +77,11 @@ jobs:
           command: pip install -e . --install-option="--buildgo"
           name: setup
       - run:
-          command: pytest ./tests
+          command: pytest --cov=hivemind ./tests
           name: tests
+      - run:
+          command: codecov
+          name: codecov
```
mryab, line 84: Better to run this for all versions just in case (codecov should be able to merge the reports)

mryab (changes requested):

.circleci/config.yml
yhn112: 👌

yhn112:
I'm not sure if it's better to disable comments by codecov bot (like the one above) or not.

justheuristic:
> I'm not sure if it's better to disable comments of codecov bot (like the one above) or not.
> The same information is available in the "Checks" tab

[weak suggestion] If you're still undecided, let's disable the comments.

mryab (approved):

Issue #188, opened 2021-03-20
Convert averager rpc_aggregate_part to P2P


justheuristic:
[this PR should be merged into libp2p branch]

Right now, the single heaviest communication pattern in hivemind is averaging model parameters with peers. This is currently done via AllreduceRunner in this RPC:
https://github.com/learning-at-home/hivemind/blob/master/hivemind/client/averaging/allreduce.py#L204

... called via this request method:
https://github.com/learning-at-home/hivemind/blob/master/hivemind/client/averaging/allreduce.py#L137

The main quest is to overhaul this method to work via P2P
- [ ] create a test where two AllreduceRunners are manually created with two connected hivemind.P2P instances
- [ ] call rpc_aggregate_part in allreduce patern, similarly to [this test](https://github.com/learning-at-home/hivemind/blob/master/tests/test_averaging.py#L151)
- [ ] test that it still works with large tensors (>10k values)

Since gRPC is not optimized for large messages, we slice these tensors into parts using split_for_streaming/combine_from_streaming. That said, P2P transport may not need this partitioning.

borzunov:
Solved in #323.

Issue #169, opened 2021-03-03
optimize load_state_from_peers


justheuristic:
__problem:__ if many peers join at once, they will all pick one averager (latest at the time) as a target for loading initial state. This is will cause choke points as one averager struggles to service many newcomers.

__possible solution:__
* modify the RPC load state so that each averager can only service T clients at a time. All subsequent clients get enqueued and wait.
* modify DecentralizedAverager.load_state_from_peers to dynamically switch away to alternative donor averagers after the first one sends you away
* maybe also negotiate for bandwidth: donors reveal their free bandwidth and clients can pick the best host

mryab:
Also, remove any remaining mentions of tensor chunks from the codebase

PR #272, opened 2021-06-14
Add per-tensor compression, faster and more flexible all-reduce

justheuristic:
* [x] TensorPartContainer / TensorPartition
   * [x] verify that the code before compression runs quickly
   * [x] account for compression size in partitioning
   * [x] reformulate chunk size in terms of bytes
* [x] ensure that compression happens asynchronously and with background threads (per chunk)
* [x] update AllreduceProtocol for new partitioning 
* [x] asynchronous AllReduceRunner core
* [x] per-tensor compression in TensorPartContainer
* [x] optional: find a way to minimize memory allocation (e.g. use iterator in update_tensors)
* [x] optional: should we lock averaged tensors while they are chunked? (throughout allreduce runtime!)
* [x] we will no longer need (1) hagenbach-bishoff (2) split/combine tensors
* [x] update DecentralizedAverager to use new AllReduceProtocol

* tests:
  * [x] test that partitioning recovers the original tensors
  * [x] test partitioning edge cases (e.g. empty tensors)
  * [x] test that partitioning is indeed asynchronous
  * [x] test new all-reduce protocol in separate file
  * [x] test asyncio utility functions
  * [x] benchmark performance under limited bandwidth


tests/test_averaging.py
```diff
@@ -201,68 +201,6 @@ def test_allgather():
     dht.shutdown()
 
 
-@pytest.mark.forked
-@pytest.mark.asyncio
-async def test_allreduce_protocol():
```
justheuristic, line 206: moved to test_allreduce.py

tests/test_averaging.py
```diff
@@ -201,68 +201,6 @@ def test_allgather():
...
-    for peer, allreduce in zip(peers, allreduce_protocols):
-        assert allreduce.future.done()
-        averaged_tensors = await allreduce
-        assert len(averaged_tensors) == len(reference_tensors)
-        assert all(torch.allclose(our, ref, atol=1e-6, rtol=0)
-                   for our, ref in zip(averaged_tensors, reference_tensors))
-
-
-@pytest.mark.forked
-def test_partitioning():
```
justheuristic, line 243: this functions were removed in favor of TensorPartContainer

hivemind/client/averaging/__init__.py
```diff
@@ -324,35 +312,43 @@ async def _step(self, *, future: MPFuture, gather_binary: bytes, weight: float,
                 future.set_exception(RuntimeError("Internal sanity check failed: averager.step left future pending."
                                                   " Please report this to hivemind issues."))
 
-    async def _make_allreduce_runner(self, group_info: GroupInfo, min_vector_size: int, **kwargs) -> AllReduceRunner:
-        """ Use a group description found by Matchmaking to form AllreduceRunner """
+    async def _run_allreduce(self, group_info: GroupInfo, min_vector_size: int, **kwargs) -> GatheredData:
+        """ run all-reduce in a given group and update tensors in-place, return gathered metadata """
         try:
```
justheuristic, line 317: Old parameterization: you define one all-reduce, we run it.

New parameterization: we give you group ids, you do whatever TF you want with them

For instance, PowerSGD can run 2 allreduce runners in a sequence

hivemind/client/averaging/__init__.py
```diff
@@ -61,7 +60,7 @@ class DecentralizedAverager(mp.Process, averaging_pb2_grpc.DecentralizedAveragin
       towards the (estimated) average by this coefficient. By default, local parameters are set equal to average.
     :param request_timeout: when looking for group, wait for a response from leader for at most this many seconds.
     :note: request_timeout must be smaller than averaging_expiration to avoid potential deadlocks.
-    :param chunk_size_bytes: tensors for AllReduce will be divided into chunks of this size (to improve gRPC throughput)
+    :param part_size_bytes: tensors for AllReduce will be divided into chunks of this size (to improve gRPC throughput)
```
justheuristic, line 63: Unified glossary:
- peer_fractions = fractions of total averaged_tensors for each peer
- *part* = small tensor sliced from a larger one for All-Reduce
- ~~chunk~~ deprecated, only used in load_state_from peers until we switch to libp2p

hivemind/client/averaging/__init__.py
```diff
@@ -324,35 +312,43 @@ async def _step(self, *, future: MPFuture, gather_binary: bytes, weight: float,
...
+                        tensor.add_(update, alpha=self._averaging_alpha)
+                    self.last_updated = get_dht_time()
+
+                return allreduce_protocol.gathered
+        except BaseException as e:
+            logger.exception(e)
+            raise MatchmakingException(f"Unable to run all-reduce ({e})")
+        finally:
+            _ = self._running_groups.pop(group_info.group_id, None)
+            self._pending_group_assembled.set()
```
justheuristic, line 351: Q: @mryab should we extract a contextmanager for this?

```python
async with register_allreduce_protocol(allreduce_protocol):
      # ... adds group to running_groups, accepts incoming connections
      <do something>
# ... removes group from running_groups
```

hivemind/client/averaging/__init__.py
```diff
@@ -20,23 +20,22 @@
 import numpy as np
 
 from hivemind.dht import DHT, DHTID
-from hivemind.client.averaging.allreduce import AllReduceRunner, AllreduceException, GroupID, AveragingMode
+from hivemind.client.averaging.allreduce import AllReduceProtocol, AllreduceException, GroupID, AveragingMode
```
justheuristic, line 23: Before:
- AllreduceProtocol contains all logic (splitting, averaging)
- AllreduceRunner contains protocol implementation

After:
- TensorPartContainer contains splitting logic
- TensorPartAverager contains averaging logic
- AllReduceProtocol contains protocol implementation

hivemind/client/averaging/load_balancing.py
```diff
@@ -28,6 +28,7 @@ def load_balance_peers(vector_size, throughputs: Sequence[Optional[float]], min_
         assert not all(throughput == 0 for throughput in throughputs), "Must have at least one nonzero throughput"
         scores = np.asarray([1.0 if throughput is None else 0.0 for throughput in throughputs])
 
+    #TODO(jheuristic) we no longer need hagenbach-bishoff with new AllReduceProtocol
     return tuple(hagenbach_bishoff(vector_size, scores))
```
justheuristic, line 33: // intentionally leaving this TODO for new LP PR

hivemind/utils/grpc.py
```diff
@@ -158,7 +158,8 @@ def top(self) -> ValueError:
         raise ValueError(f"Please use {self.__class__.__name__}.get_stub to get or create stubs")
 
 
-def split_for_streaming(serialized_tensor: runtime_pb2.Tensor, chunk_size_bytes: int) -> Iterator[runtime_pb2.Tensor]:
+def split_for_streaming(serialized_tensor: runtime_pb2.Tensor, chunk_size_bytes: int = 2 ** 16
+                        ) -> Iterator[runtime_pb2.Tensor]:
     """ Split serialized_tensor into multiple chunks for gRPC streaming """
     buffer = memoryview(serialized_tensor.buffer)
     num_chunks = len(range(0, len(buffer), chunk_size_bytes))
```
justheuristic, line 165: TODO remove this function after DecentralizedAverager moves to libp2p backend

hivemind/client/averaging/__init__.py
```diff
@@ -291,16 +289,10 @@ async def _step(self, *, future: MPFuture, gather_binary: bytes, weight: float,
...
-                    self._running_groups[group_id] = allreduce_runner
-                    self._pending_group_assembled.set()
-                    await asyncio.wait_for(allreduce_runner.run(), self._allreduce_timeout)
-                    if self.mode != AveragingMode.AUX:
-                        await loop.run_in_executor(None, self.update_tensors, allreduce_runner)
 
-                    # averaging is finished, exit the loop
-                    future.set_result(allreduce_runner.gathered)
+                    future.set_result(await asyncio.wait_for(
+                        self._run_allreduce(group_info, **self.allreduce_kwargs), self._allreduce_timeout))
```
justheuristic, line 294: Please note: this is a lie :)

Codecov does not collect lines that were used in a separate process. It should feel bad for not doing so.

hivemind/client/averaging/__init__.py
```diff
@@ -291,16 +289,10 @@ async def _step(self, *, future: MPFuture, gather_binary: bytes, weight: float,
...
-                    allreduce_runner = await self._make_allreduce_runner(group_info, **self.allreduce_kwargs)
-                    self._running_groups[group_id] = allreduce_runner
-                    self._pending_group_assembled.set()
-                    await asyncio.wait_for(allreduce_runner.run(), self._allreduce_timeout)
-                    if self.mode != AveragingMode.AUX:
-                        await loop.run_in_executor(None, self.update_tensors, allreduce_runner)
 
-                    # averaging is finished, exit the loop
-                    future.set_result(allreduce_runner.gathered)
+                    future.set_result(await asyncio.wait_for(
```
justheuristic, line 293: Please note: the codecov message is a lie :)

Codecov does not collect lines that were used in a separate process. It should feel bad for not doing so.

hivemind/client/averaging/__init__.py
```diff
@@ -265,10 +265,10 @@ def step(self, gather: Optional[DataForGather] = None, weight: float = 1.0, time
         :param wait: if True (default), return when finished. Otherwise return MPFuture and run in background.
         :returns: on success, update averaged_tensors and return group info; on failure, return None
         """
-        if self.mode == AveragingMode.AUX and weight != 1:
+        if self.mode == AveragingMode.AUX and weight is not None:
             logger.warning("Averager is running in auxiliary mode, weight is unused.")
-        else:
-            assert isinstance(weight, (int, float)) and weight > 0, f"Expected a positive int/float, got {type(weight)}"
+        weight = weight if weight is not None else float(self.mode != AveragingMode.AUX)
+        assert isinstance(weight, (int, float)) and weight >= 0, f"Expected a positive int/float, got {type(weight)}"
```
mryab, line 271: Maybe expect strictly a float, as suggested by the type hint?

mryab: BTW the assertion for type is somewhat redundant if we already have type hints

hivemind/client/averaging/__init__.py
```diff
@@ -324,35 +312,43 @@ async def _step(self, *, future: MPFuture, gather_binary: bytes, weight: float,
                 future.set_exception(RuntimeError("Internal sanity check failed: averager.step left future pending."
                                                   " Please report this to hivemind issues."))
 
-    async def _make_allreduce_runner(self, group_info: GroupInfo, min_vector_size: int, **kwargs) -> AllReduceRunner:
-        """ Use a group description found by Matchmaking to form AllreduceRunner """
+    async def _run_allreduce(self, group_info: GroupInfo, min_vector_size: int, **kwargs) -> GatheredData:
+        """ run all-reduce in a given group and update tensors in-place, return gathered metadata """
```
mryab, line 316:
```suggestion
        """ Run All-Reduce in a given group and update tensors in place, return gathered metadata """
```

hivemind/client/averaging/__init__.py
```diff
@@ -324,35 +312,43 @@ async def _step(self, *, future: MPFuture, gather_binary: bytes, weight: float,
...
+                    for tensor, update in zip(local_tensors, averaging_outputs):
+                        tensor.add_(update, alpha=self._averaging_alpha)
+                    self.last_updated = get_dht_time()
+
+                return allreduce_protocol.gathered
+        except BaseException as e:
+            logger.exception(e)
+            raise MatchmakingException(f"Unable to run all-reduce ({e})")
+        finally:
+            _ = self._running_groups.pop(group_info.group_id, None)
```
mryab, line 350: No need to assign a value

hivemind/client/averaging/__init__.py
```diff
@@ -324,35 +312,43 @@ async def _step(self, *, future: MPFuture, gather_binary: bytes, weight: float,
...
+                if modes[group_info.endpoints.index(self.endpoint)] != AveragingMode.AUX:
+                    assert len(local_tensors) == len(self._averaged_tensors)
+                    for tensor, update in zip(local_tensors, averaging_outputs):
+                        tensor.add_(update, alpha=self._averaging_alpha)
+                    self.last_updated = get_dht_time()
+
+                return allreduce_protocol.gathered
+        except BaseException as e:
+            logger.exception(e)
+            raise MatchmakingException(f"Unable to run all-reduce ({e})")
```
mryab, line 348:
```suggestion
            raise MatchmakingException(f"Unable to run All-Reduce: {e}")
```

hivemind/client/averaging/__init__.py
mryab: Generally agree with the idea, but what would be the usage scenarios outside of this specific method?

hivemind/utils/grpc.py
```diff
@@ -158,7 +158,8 @@ def top(self) -> ValueError:
         raise ValueError(f"Please use {self.__class__.__name__}.get_stub to get or create stubs")
 
 
-def split_for_streaming(serialized_tensor: runtime_pb2.Tensor, chunk_size_bytes: int) -> Iterator[runtime_pb2.Tensor]:
+def split_for_streaming(serialized_tensor: runtime_pb2.Tensor, chunk_size_bytes: int = 2 ** 16
```
mryab, line 161: Better to declare a global constant

hivemind/client/averaging/allreduce.py
```diff
@@ -21,256 +22,203 @@ class AveragingMode(Enum):
     AUX = 2
 
 
-class AllReduceProtocol:
+class AllReduceProtocol(averaging_pb2_grpc.DecentralizedAveragingServicer):
     """
     An internal class that runs butterfly AllReduce in a predefined group of averagers
 
-    :param tensors: local tensors that should be averaged with groupmates
+    :note: this class returns **differences** between averaged and local tensors in order to improve numeric stability
```
mryab, line 29:
```suggestion
    :note: this class returns **differences** between averaged and local tensors in order to improve numerical stability
```

hivemind/utils/asyncio.py
```diff
@@ -27,13 +34,31 @@ async def aiter(*args: T) -> AsyncIterator[T]:
...
 
 
+async def azip(*iterables: AsyncIterable[T]) -> AsyncIterator[Tuple[T, ...]]:
+    """ equivalent of zip for asynchronous iterables """
+    iterators = [iterable.__aiter__() for iterable in iterables]
+    while True:
+        try:
+            yield tuple(await asyncio.gather(*(itr.__anext__() for itr in iterators)))
+        except StopAsyncIteration:
+            break
```
mryab, line 44: TODO maybe replace these helper functions with equivalent versions from some lightweight library and use it as a dependency

hivemind/utils/asyncio.py
```diff
@@ -42,3 +67,26 @@ async def await_cancelled(awaitable: Awaitable) -> bool:
         return True
     except BaseException:
         return False
+
+
+async def async_map(func: callable, *iterables: AsyncIterable, max_prefetch: Optional[int] = None,
+                    executor: Optional[ThreadPoolExecutor] = None):
```
mryab, line 73: Return type?

hivemind/utils/compression.py
```diff
@@ -188,3 +188,15 @@ def deserialize_torch_tensor(serialized_tensor: runtime_pb2.Tensor) -> torch.Ten
 
     tensor.requires_grad_(serialized_tensor.requires_grad)
     return tensor
+
+
+def get_nbytes_per_value(dtype: torch.dtype, compression: CompressionType):
```
mryab, line 193:
```suggestion
def get_nbytes_per_value(dtype: torch.dtype, compression: CompressionType) -> int:
```

hivemind/utils/compression.py
```diff
@@ -188,3 +188,15 @@ def deserialize_torch_tensor(serialized_tensor: runtime_pb2.Tensor) -> torch.Ten
...
+    """ returns the number of bits per value for a given tensor (excluding metadata) """
+    if compression in (CompressionType.QUANTILE_8BIT, CompressionType.UNIFORM_8BIT):
+        return 1
+    elif compression in (CompressionType.FLOAT16, CompressionType.MEANSTD_16BIT):
+        return 2
+    elif compression == CompressionType.NONE:
+        return torch.finfo(dtype).bits / 8
```
mryab, line 200:
```suggestion
        return torch.finfo(dtype).bits // 8
```

mryab (changes requested):

hivemind/client/averaging/__init__.py
yhn112: pytest-cov.readthedocs.io/en/latest/subprocess-support.html

hivemind/client/averaging/__init__.py
justheuristic: done that already; the only remaining occurences relate to deprecated split_for_streaming, which will be removed in future

hivemind/client/averaging/__init__.py
justheuristic: This is an intended behavior.

**(1)** typing supports this:
>Rather than requiring that users write import numbers and then use numbers.Float etc., this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type float, an argument of type int is acceptable; similar, for an argument annotated as having type complex, arguments of type float or int are acceptable.

https://www.python.org/dev/peps/pep-0484/#the-numeric-tower

**(2)** isinstance requires explicit checks:
image

hivemind/client/averaging/__init__.py
justheuristic: it was required for pycharm, and it is still required by pycharm

hivemind/utils/grpc.py
justheuristic: thx; will do

hivemind/utils/asyncio.py
justheuristic: Added to backlog for later

hivemind/client/averaging/__init__.py
justheuristic: The trick is, we're deliberately not calling join in some cases.

Though, thois should not apply to DecentralizedAverager

hivemind/client/averaging/__init__.py
```diff
@@ -265,10 +265,10 @@ def step(self, gather: Optional[DataForGather] = None, weight: float = 1.0, time
         :param wait: if True (default), return when finished. Otherwise return MPFuture and run in background.
         :returns: on success, update averaged_tensors and return group info; on failure, return None
         """
-        if self.mode == AveragingMode.AUX and weight != 1:
+        if self.mode == AveragingMode.AUX and weight is not None:
             logger.warning("Averager is running in auxiliary mode, weight is unused.")
-        else:
-            assert isinstance(weight, (int, float)) and weight > 0, f"Expected a positive int/float, got {type(weight)}"
+        weight = weight if weight is not None else float(self.mode != AveragingMode.AUX)
```
yhn112, line 270:
```suggestion
        if weight is None:
            weight = float(self.mode != AveragingMode.AUX)
```

hivemind/client/averaging/__init__.py
justheuristic: There is only one, but it is still more idiomatic.

If we extend DecentralizedOptimizer to implement more complex averaging, the extended class would be able to reuse this.

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+class TensorPartContainer:
+    """
+    Auxiliary data structure for averaging, responsible for splitting tensors into parts and re-assembling them.
+    The class is designed to avoid excessive memory allocation and run all heavy computation in background
+    """
+
+    def __init__(self, tensors: Sequence[torch.Tensor], peer_fractions: Sequence[float],
+                 compression_type: Union[type(CompressionType), Sequence[type(CompressionType)]] = CompressionType.NONE,
```
yhn112, line 26: AFAIK,
```suggestion
                 compression_type: Union[Type[CompressionType], Sequence[Type[CompressionType]]] = CompressionType.NONE,
```

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
+"""
+Auxiliary data structures for AllReduceProtocol
+"""
+import asyncio
+from typing import Sequence, AsyncIterable, Tuple, Optional, TypeVar, Union
```
yhn112, line 5:
```suggestion
from typing import Sequence, AsyncIterable, Tuple, Optional, TypeVar, Union, Type
```

hivemind/client/averaging/partition.py
justheuristic: alas, this does not work with protobuf

hivemind/client/averaging/__init__.py
mryab: Understood, but do we need this assertion for types anyway if we explicitly indicate that we need a numeric type here? Seems like excessive checking to me

hivemind/client/averaging/__init__.py
mryab: OK then, but add it to a corresponding issue so that we don't forget about it

justheuristic:
## Performance benchmark:

__Method:__
- started 2 nodes, one on nora and another on my laptop
- created 4 tensor of shapes
- benchmarked full averaging step from client's perspective (which is slightly longer than from server's perspective)
- restarted 3 times, interchanging hivemind v0.9.9 and this branch



__Branch 0.9.9:__
- CompressionType: NONE: 37.2s +- 1.2s
- CompressionType: FLOAT16: 19.9 +- 0.7s


Network pattern (laptop side)
- two distinct phases: the first phase is mostly upload, the second mostly download
image

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+                    self._input_parts_by_peer[assigned_peer_index].append((part, tensor_compression))
+                else:
+                    self._input_parts_by_peer[current_peer_index].append((part, tensor_compression))
+                current_length += len(part)
+        assert current_length == self.total_size
+        for current_peer_index in range(self.group_size):
+            self.num_parts_by_peer.append(len(self._input_parts_by_peer[current_peer_index]))
+
+    @torch.no_grad()
+    def get_raw_input_parts(self, peer_index: int) -> Tuple[torch.Tensor, ...]:
```
mryab, line 74:
```suggestion
    def get_raw_parts(self, peer_index: int) -> Tuple[torch.Tensor, ...]:
```

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+    async def iterate_input_parts_for(self, peer_index: int) -> Tensor:
+        """ iterate serialized tensor parts for a peer at a given index. Run serialization in background. """
+        assert not self._inputs_consumed_by_peer[peer_index], "input parts of a given peer are already deallocated."
+        self._inputs_consumed_by_peer[peer_index] = True
+
+        async def _aiterate_parts():
+            for _ in range(self.num_parts_by_peer[peer_index]):
+                yield self._input_parts_by_peer[peer_index].popleft()
+
+        async for serialized_part in async_map(lambda args: serialize_torch_tensor(*args),
```
mryab, line 92: Can't we replace this with just `async_map(serialize_torch_tensor,` since we're basically passing tensors to it?

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+    def finalize(self):
+        """ terminate all iterators, delete intermediate data """
+        if not self.finished.is_set():
+            for peer_index in range(self.group_size):
+                self._inputs_consumed_by_peer[peer_index] = True
+                self._input_parts_by_peer[peer_index].clear()
+                self._output_parts_by_peer[peer_index].clear()
+                self._output_part_available[peer_index].set()
+            self._outputs_consumed = True
+            self.finished.set()
```
mryab, line 144: Is it possible to move this to `del` and then just call `del` for each instance of TensorPartContainer that we intend to finalize?

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+    :note: even if local peer is not sending data, local parts will be used for shape information
+    """
+    current_part_index: int = -1  # index in local_parts of the part that should be loaded next
+    current_part_accumulated_from: int = 0  # number of peers from which the current part was accumulated
+    accumulator: torch.Tensor  # sum of current tensor part from group peers
+    denominator: float  # total weight accumulated from all peers for current part
+    current_part_future: asyncio.Future  # this future will be set with the current averaged part, once it is ready
+
+    def __init__(self, part_shapes: Sequence[torch.Size], num_senders: int,
+                 weights: Optional[Sequence[float]] = None):
```
mryab, line 161: Also add `weights` to the docstring

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+    Auxiliary data structure responsible for running asynchronous all-reduce
+    :param part_shapes: a sequence of shapes of torch tensors that will be averaged by this reducer
+    :param num_senders: total number of peers in a given all-reduce group that will send gradients
+    :note: even if local peer is not sending data, local parts will be used for shape information
+    """
+    current_part_index: int = -1  # index in local_parts of the part that should be loaded next
+    current_part_accumulated_from: int = 0  # number of peers from which the current part was accumulated
+    accumulator: torch.Tensor  # sum of current tensor part from group peers
+    denominator: float  # total weight accumulated from all peers for current part
+    current_part_future: asyncio.Future  # this future will be set with the current averaged part, once it is ready
```
mryab, line 158: This probably isn't a singleton, so better to move all these class attributes inside the initialization

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+                 weights: Optional[Sequence[float]] = None):
+        self.part_shapes, self.num_senders, self.num_parts = part_shapes, num_senders, len(part_shapes)
+        self.weights = tuple(weights or (1 for _ in range(num_senders)))
+        assert len(self.weights) == self.num_senders, "The number of weights is inconsistent with num_senders"
+        assert all(isinstance(weight, (int, float)) for weight in self.weights)
+        self.finished = asyncio.Event()
+        self.reset_accumulators()
+
+    def reset_accumulators(self):
+        """ (re)create averaging buffers for the part part in line, pre-populates with local tensor part """
```
mryab, line 170:
```suggestion
        """ (re)create averaging buffers for the next part in line, prepopulate with local tensor part """
```

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+        assert self.current_part_accumulated_from == self.num_senders or self.current_part_index == -1
+        if self.current_part_index >= self.num_parts - 1:
+            self.finalize()
```
mryab, line 173: Should we call finalize here, or the object will be deleted anyway?

hivemind/utils/asyncio.py
```diff
@@ -42,3 +67,26 @@ async def await_cancelled(awaitable: Awaitable) -> bool:
         return True
     except BaseException:
         return False
+
+
+async def async_map(func: Callable[..., T], *iterables: AsyncIterable, max_prefetch: Optional[int] = None,
+                    executor: Optional[ThreadPoolExecutor] = None) -> AsyncIterator[T]:
+    """ iterate from an async iterable in a background thread, yield results to async iterable """
```
mryab, line 74: I think that it's not map but starmap, since you pass multiple iterables to it :) 
Two options — either rename, or create a map-only version and use it as starmap only where needed (passing azip as an argument)

tests/test_allreduce.py
```diff
@@ -0,0 +1,221 @@
...
+from hivemind.proto import averaging_pb2_grpc
+
+
+@pytest.mark.forked
+@pytest.mark.asyncio
+async def test_partitioning():
+    all_tensors = [
+        torch.randn(30_000, 128), torch.rand(128), torch.ones(1, 1, 1, 1, 1, 1, 8),
+        torch.ones(1, 0), torch.zeros(0), torch.zeros([]), torch.randn(65536),
+        torch.rand(512, 2048), torch.randn(1024, 1024).add(-9), torch.zeros(1020), torch.randn(4096)
```
mryab, line 24: Is 30_000 and 65536 really needed? I'm not sure that we need tensors this large to verify that everything works

tests/test_allreduce.py
```diff
@@ -0,0 +1,221 @@
...
+async def test_partitioning_edge_cases(tensors: Sequence[torch.Tensor], peer_fractions: Sequence[float]):
+    partition = TensorPartContainer(tensors, peer_fractions, part_size_bytes=16)
+    for peer_index in range(len(peer_fractions)):
+        async for part_index, part in aenumerate(partition.iterate_input_parts_for(peer_index)):
+            partition.register_processed_part(peer_index, part_index, deserialize_torch_tensor(part))
+
+    tensor_index = 0
+    async for output_tensor in partition.iterate_output_tensors():
+        assert torch.allclose(output_tensor, tensors[tensor_index])
+        tensor_index += 1
```
mryab, line 63: Replace with zip?

tests/test_allreduce.py
```diff
@@ -0,0 +1,221 @@
...
+                        async for part_index, part in aenumerate(partition.iterate_input_parts_for(peer_index)):
+                            output_tensor = torch.sin(deserialize_torch_tensor(part))
+                            partition.register_processed_part(peer_index, part_index, output_tensor)
+
+                task = asyncio.create_task(write_tensors())
+                tensor_index = 0
+                async for output_tensor in partition.iterate_output_tensors():
+                    assert torch.allclose(output_tensor, torch.sin(tensors[tensor_index]))
+                    tensor_index += 1
+                assert tensor_index == len(tensors)
```
mryab, line 45: Replace with zip?

tests/test_allreduce.py
```diff
@@ -0,0 +1,221 @@
...
+
+    async def wait_synchronously():
+        time_in_waiting = 0.0
+        while not read_finished.is_set():
+            await asyncio.sleep(0.01)
+            time_in_waiting += 0.01
+        return time_in_waiting
+
+    start_time = time.perf_counter()
+    _, _, time_in_waiting = await asyncio.gather(write_tensors(), read_tensors(), wait_synchronously())
```
mryab, line 96:
```suggestion
    *_, time_in_waiting = await asyncio.gather(write_tensors(), read_tensors(), wait_synchronously())
```

tests/test_allreduce.py
```diff
@@ -0,0 +1,221 @@
...
+        assert torch.allclose(output_tensor, tensors[tensor_index])
+        tensor_index += 1
+
+
+@pytest.mark.forked
+@pytest.mark.asyncio
+async def test_partitioning_asynchronous():
+    """ ensure that tensor partitioning does not interfere with asynchronous code """
+    tensors = [torch.randn(2048, 2048), torch.randn(1024, 4096),
+               torch.randn(4096, 1024), torch.randn(30_000, 1024)]
```
mryab, line 71: Here, the tensors are also probably too large for testing purposes

tests/test_allreduce.py
```diff
@@ -0,0 +1,221 @@
...
+        time_in_waiting = 0.0
+        while not read_finished.is_set():
+            await asyncio.sleep(0.01)
+            time_in_waiting += 0.01
+        return time_in_waiting
+
+    start_time = time.perf_counter()
+    _, _, time_in_waiting = await asyncio.gather(write_tensors(), read_tensors(), wait_synchronously())
+    wall_time = time.perf_counter() - start_time
+    assert time_in_waiting > wall_time / 3, f"Event loop could only run {time_in_waiting / wall_time :.5f} of the time"
```
mryab, line 98: The condition is not clear, why we're verifying that we waited at least one third of the time and then saying something about the event loop?

tests/test_allreduce.py
```diff
@@ -0,0 +1,221 @@
...
+
+    averaged_tensors_by_peer = await asyncio.gather(*map(send_tensors, range(num_senders)))
+
+    reference = [sum(local_tensors_by_sender[sender_index][part_index]
+                     for sender_index in range(num_senders)) / num_senders
+                 for part_index in range(num_parts)]
+
+    for averaged_tensors in averaged_tensors_by_peer:
+        assert len(averaged_tensors) == len(reference)
+        for avg, ref in zip(averaged_tensors, reference):
```
mryab, line 135:
```suggestion
        for averaging_result, reference_tensor in zip(averaged_tensors, reference):
```

tests/test_allreduce.py
```diff
@@ -0,0 +1,221 @@
...
+    async def _run_allreduce_inplace(allreduce: AllReduceProtocol):
+        async for tensor_index, tensor_delta in aenumerate(allreduce):
+            allreduce.tensor_part_container.local_tensors[tensor_index].add_(tensor_delta)
+
+    await asyncio.gather(*map(_run_allreduce_inplace, allreduce_protocols))
+
+    reference_tensors = [
+        sum(tensors_by_peer[peer][i] * averaging_weights[peer_index]
+            for peer_index, peer in enumerate(peers)
+            ) / sum(averaging_weights)
```
mryab, line 202: Maybe shrink that to one line?

tests/test_averaging.py
```diff
@@ -130,6 +129,48 @@ def test_allreduce_weighted(n_client_mode_peers: int = 2):
...
+def test_allreduce_compression():
+    dht = hivemind.DHT(start=True, endpoint=f'{hivemind.LOCALHOST}:*')
+
+    tensors1 = [torch.linspace(0, 500, 1000) ** 0.5, torch.randn(1000)]
+    tensors2 = [torch.linspace(300, 800, 1000) ** 0.5, torch.randn(1000)]
+    results = {}
+
+    FLOAT16, UINT8 = CompressionType.FLOAT16, CompressionType.UNIFORM_8BIT
+
+    for compression_type in [(FLOAT16, FLOAT16), (FLOAT16, UINT8), (UINT8, FLOAT16), (UINT8, UINT8)]:
```
mryab, line 142:
```suggestion
    for compression_type_pair in [(FLOAT16, FLOAT16), (FLOAT16, UINT8), (UINT8, FLOAT16), (UINT8, UINT8)]:
```

tests/test_averaging.py
```diff
@@ -130,6 +129,48 @@ def test_allreduce_weighted(n_client_mode_peers: int = 2):
...
+
+    assert torch.allclose(results[UINT8, FLOAT16][0], results[UINT8, UINT8][0])
+    assert torch.allclose(results[UINT8, FLOAT16][1], results[FLOAT16, FLOAT16][1])
+    assert torch.allclose(results[UINT8, UINT8][1], results[FLOAT16, UINT8][1])
+    assert torch.allclose(results[FLOAT16, UINT8][0], results[FLOAT16, FLOAT16][0])
+
+    assert not torch.allclose(results[UINT8, FLOAT16][1], results[UINT8, UINT8][1])
+    assert not torch.allclose(results[UINT8, FLOAT16][0], results[FLOAT16, FLOAT16][0])
+    assert not torch.allclose(results[UINT8, UINT8][0], results[FLOAT16, UINT8][0])
+    assert not torch.allclose(results[FLOAT16, UINT8][1], results[FLOAT16, FLOAT16][1])
```
mryab, line 165: I'm not sure that these checks are necessary; if the results happen to be close, is that really undesired?

tests/test_averaging.py
```diff
@@ -130,6 +129,48 @@ def test_allreduce_weighted(n_client_mode_peers: int = 2):
...
+
+    reference = [(tensors1[i] + tensors2[i]) / 2 for i in range(len(tensors1))]
+    assert 0 < torch.mean(torch.square(results[FLOAT16, FLOAT16][0] - reference[0])).item() <= 1e-5
+    assert 0 < torch.mean(torch.square(results[FLOAT16, FLOAT16][1] - reference[1])).item() <= 1e-5
+    assert 1e-5 < torch.mean(torch.square(results[UINT8, UINT8][0] - reference[0])).item() <= 1e-2
+    assert 1e-5 < torch.mean(torch.square(results[UINT8, UINT8][1] - reference[1])).item() <= 1e-2
```
mryab, line 171: Also, not sure that the left bound is necessary here

tests/test_averaging.py
```diff
@@ -130,6 +129,48 @@ def test_allreduce_weighted(n_client_mode_peers: int = 2):
...
+            future.result()
+
+        with averager1.get_tensors() as averaged_tensors:
+            results[compression_type] = averaged_tensors
+
+
+    assert torch.allclose(results[UINT8, FLOAT16][0], results[UINT8, UINT8][0])
+    assert torch.allclose(results[UINT8, FLOAT16][1], results[FLOAT16, FLOAT16][1])
+    assert torch.allclose(results[UINT8, UINT8][1], results[FLOAT16, UINT8][1])
+    assert torch.allclose(results[FLOAT16, UINT8][0], results[FLOAT16, FLOAT16][0])
```
mryab, line 160: Generall, the point of this test is not clear to me. Why do we check that for some pairs of tensors and some compression values the checks succeed and for others they fail?

tests/test_util_modules.py
```diff
@@ -271,3 +273,37 @@ def test_generic_data_classes():
     sorted_expirations = sorted([DHTExpiration(value) for value in range(1, 1000)])
     sorted_heap_entries = sorted([HeapEntry(DHTExpiration(value), key="any") for value in range(1, 1000)[::-1]])
     assert all([entry.expiration_time == value for entry, value in zip(sorted_heap_entries, sorted_expirations)])
+
+
+@pytest.mark.asyncio
+async def test_asyncio():
```
mryab, line 279:
```suggestion
async def test_asyncio_utils():
```

tests/test_util_modules.py
```diff
@@ -5,6 +5,7 @@
 import pytest
 import torch
 
+from hivemind import async_map, aiter, aenumerate, achain, anext, azip
```
mryab, line 8: Better to make an explicit import from hivemind.utils

mryab (changes requested):

hivemind/client/averaging/allreduce.py
```diff
@@ -21,256 +22,203 @@ class AveragingMode(Enum):
     AUX = 2
 
 
-class AllReduceProtocol:
+class AllReduceProtocol(averaging_pb2_grpc.DecentralizedAveragingServicer):
     """
     An internal class that runs butterfly AllReduce in a predefined group of averagers
 
-    :param tensors: local tensors that should be averaged with groupmates
+    :note: this class returns **differences** between averaged and local tensors in order to improve numerical stability
```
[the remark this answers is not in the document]
justheuristic: Alas, yes; it will complicate code in both partition.py and allreduce.py AND an additional lock that ensures that local tensors did not change.

Since #225, we already return deltas all the time and only leave an option to revert to non-deltas by adding them back. This behavior persists in the current PR.

Fortunately, 
- there are no drawbacks to sending deltas, the cost of subtraction is negligible compared to any(averaging, compression, transmission)
- if absolutely necessary for some reason, one can restore the average tensor by adding deltas to local tensors

tests/test_averaging.py
justheuristic: - we check that the FP16 compression stays the same even if other tensors are compressed as UINT8
- we check that UINT8 compression stays the same even if other tensors are compressed as FP16

This will root out any situations where we mistakenly compress some **parts** of FP16 tensors as UINT8 or vice versa.

hivemind/client/averaging/allreduce.py
```diff
@@ -21,256 +22,203 @@ class AveragingMode(Enum):
...
+
+        self.sender_endpoints, self.sender_weights = [], []
+        for endpoint, weight, mode in zip(self.ordered_group_endpoints, weights, modes):
+            if mode != AveragingMode.AUX:
+                self.sender_endpoints.append(endpoint)
+                self.sender_weights.append(weight)
+
+        self.tensor_part_container = TensorPartContainer(tensors, peer_fractions, **kwargs)
+        self.parts_for_local_averaging = self.tensor_part_container.get_raw_input_parts(
+            self.ordered_group_endpoints.index(self.endpoint))
```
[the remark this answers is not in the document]
justheuristic: i would rather not, for the sake of explicitness

tests/test_averaging.py
justheuristic: Make sure that UINT8 was not mistakenly encoded as FP16

hivemind/client/averaging/allreduce.py
```diff
@@ -21,256 +22,203 @@ class AveragingMode(Enum):
...
+    :note: this class returns **differences** between averaged and local tensors in order to improve numerical stability
+    :param tensors: local tensors that should be averaged with group-mates
     :param endpoint: your endpoint, must be included in ordered_group_endpoints
     :param ordered_group_endpoints: group endpoints ordered s.t. i-th endpoint is responsible for averaging i-th part
-    :param part_sizes: for each peer, a number of vector elements that this peer is responsible for averaging
-    :param return_deltas: if True, returns the element-wise differences (averaged_tensors - original_tensors)
-           default (False) - return averaged_tensors by themselves
+    :param peer_fractions: for each peer, a target fraction of vector elements that this peer should average
+      (the actual number of values by peer will be nearly proportional, but there are no exact guarantees)
+    :param weights: scaling coefficients for weighted averaging (default = equal weights for all non-aux peers)
```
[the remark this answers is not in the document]
justheuristic: done

hivemind/client/averaging/allreduce.py
```diff
@@ -21,256 +22,203 @@ class AveragingMode(Enum):
...
+    :param part_size_bytes: all tensors will be split into chunks of up to this size before being transferred
```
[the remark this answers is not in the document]
justheuristic: done, thx

hivemind/client/averaging/__init__.py
justheuristic: created a separate issue for this

https://github.com/learning-at-home/hivemind/issues/293

hivemind/client/averaging/__init__.py
justheuristic: Yes, we do.

hivemind/client/averaging/__init__.py
justheuristic: renamed everything to AllReduceRunner

hivemind/client/averaging/__init__.py
justheuristic: added to https://github.com/learning-at-home/hivemind/issues/169

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+
+class TensorPartContainer:
+    """
+    Auxiliary data structure for averaging, responsible for splitting tensors into parts and re-assembling them.
+    The class is designed to avoid excessive memory allocation and run all heavy computation in background
+    """
+
+    def __init__(self, tensors: Sequence[torch.Tensor], peer_fractions: Sequence[float],
+                 compression_type: Union[type(CompressionType), Sequence[type(CompressionType)]] = CompressionType.NONE,
+                 part_size_bytes: int = 2 ** 20, prefetch: int = 1):
```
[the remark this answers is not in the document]
justheuristic: done/done

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+        if not isinstance(compression_type, Sequence):
+            compression_type = [compression_type] * len(tensors)
+        assert len(compression_type) == len(tensors), "compression types do not match the number of tensors"
+        self.local_tensors, self.peer_fractions, self.group_size = tensors, peer_fractions, len(peer_fractions)
+        self.compression_type, self.part_size_bytes, self.prefetch = compression_type, part_size_bytes, prefetch
+        self.tensor_sizes = [tensor.numel() for tensor in tensors]
```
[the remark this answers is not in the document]
justheuristic: removed altogether

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+                    while current_length + len(part) > pivots[current_peer_index]:
+                        current_peer_index += 1
+                        current_peer_part_end = min(current_length + len(part), pivots[current_peer_index])
+                        peer_intersections.append(current_peer_part_end - pivots[current_peer_index - 1])
+                    assigned_peer_index = prev_peer_index + np.argmax(peer_intersections)
+                    self._input_parts_by_peer[assigned_peer_index].append((part, tensor_compression))
+                else:
+                    self._input_parts_by_peer[current_peer_index].append((part, tensor_compression))
+                current_length += len(part)
+        assert current_length == self.total_size
```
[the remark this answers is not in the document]
justheuristic: ya got it :)

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+        for current_peer_index in range(self.group_size):
+            self.num_parts_by_peer.append(len(self._input_parts_by_peer[current_peer_index]))
```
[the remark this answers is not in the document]
justheuristic: done :)

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+    def get_raw_input_parts(self, peer_index: int) -> Tuple[torch.Tensor, ...]:
+        """ get non-serialized tensor parts for a peer at a given index """
+        assert not self._inputs_consumed_by_peer[peer_index], "input parts of a given peer are already deallocated."
+        self._inputs_consumed_by_peer[peer_index] = True
+        input_parts = tuple(part for part, compression in self._input_parts_by_peer[peer_index])
+        self._input_parts_by_peer[peer_index].clear()
+        return input_parts
+
+    @torch.no_grad()
+    async def iterate_input_parts_for(self, peer_index: int) -> Tensor:
```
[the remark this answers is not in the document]
justheuristic: If you insist, please provide some rationale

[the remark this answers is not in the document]
justheuristic: good catch :)

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def finalize(self):
+        if not self.finished.is_set():
+            if hasattr(self, 'current_part_future'):
+                self.current_part_future.cancel()
+                del self.accumulator
+            self.finished.set()
+
+    def __await__(self):
+        return self.finished.wait().__await__()
```
borzunov, line 215: nit: It's unclear for me that awaiting the reducer should wait for its destructor to finish (not e.g. wait for the result). Is it a common asyncio pattern?

hivemind/client/averaging/partition.py
justheuristic: these are specifically input parts

hivemind/client/averaging/partition.py
justheuristic: no, we cannot. We pass compression types.

hivemind/client/averaging/partition.py
justheuristic: no, it is not.

hivemind/client/averaging/partition.py
justheuristic: done

tests/test_allreduce.py
justheuristic: Yes, we do. The default chunk size is 2 ** 20.

hivemind/client/averaging/partition.py
```diff
@@ -0,0 +1,221 @@
...
+        self._output_part_available = [asyncio.Event() for _ in range(self.group_size)]
+        self._outputs_registered_by_peer = [0 for _ in range(self.group_size)]
+        self._outputs_consumed = False
+        self.finished = asyncio.Event()
+
+        # split tensor parts in proportion to target_size_by_peer
+        current_length = 0
+        current_peer_index = 0
+        pivots = np.cumsum(peer_fractions) / np.sum(peer_fractions) * self.total_size
+        pivots = np.concatenate([pivots.astype(np.int64)[:-1], [self.total_size]])
```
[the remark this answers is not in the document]
justheuristic: done that, thx

hivemind/client/averaging/partition.py
justheuristic: That would result in code duplication in `__init__` vs `reset_accumulators`. If you insist, please provide a solution.

hivemind/client/averaging/allreduce.py
```diff
@@ -21,256 +22,203 @@ class AveragingMode(Enum):
...
-
-        self.num_senders = len([mode for mode in modes if mode != AveragingMode.AUX])
-
-        if self.num_senders == 0:
-            self.future.set_result(None)
-        for endpoint, mode in self.peer_modes.items():
-            if mode == AveragingMode.CLIENT:
-                self.averaged_tensor_parts[endpoint] = torch.tensor([])
+        modes = modes or [AveragingMode.CLIENT if frac == 0 else AveragingMode.NODE for frac in peer_fractions]
+        weights = weights or [1 if mode != AveragingMode.AUX else 0 for mode in modes]
```
borzunov, line 50: nit: I suggest to use `int(mode != AveragingMode.AUX)`.

hivemind/client/averaging/partition.py
justheuristic: I would rather keep that explicit. If you insist, please comment

tests/test_averaging.py
justheuristic: It would mean that the tensor was mistakenly compressed with the wrong compression type

tests/test_util_modules.py
justheuristic: good point, fixed

tests/test_averaging.py
justheuristic: renamed, thx

hivemind/client/averaging/allreduce.py
```diff
@@ -21,256 +22,203 @@ class AveragingMode(Enum):
...
+        stream = self._get_peer_stub(peer_endpoint).rpc_aggregate_part()
+        await stream.write(averaging_pb2.AveragingData(group_id=self.group_id, endpoint=self.endpoint, code=code))
+        await stream.done_writing()
 
-class AllreduceException(Exception):
-    """ A special exception that is raised when allreduce can't continue normally (e.g. disbanded/bad request/etc) """
+    def finalize(self, *, cancel: bool = False, exception: Optional[BaseException] = None):
+        assert not cancel or not exception, "finalize accepts either exception or cancel, but not both"
+        if not self._future.done():
+            logger.debug(f"{self} - {'cancelled' if cancel else exception or 'finished'}")
```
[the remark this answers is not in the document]
justheuristic: done

hivemind/client/averaging/partition.py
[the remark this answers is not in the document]
justheuristic: I would rather not, to highlight that this behaves differently from output parts

hivemind/utils/asyncio.py
justheuristic: fixd

hivemind/client/averaging/__init__.py
borzunov: Yes, this will definitely be useful for implementing other reducing methods.

borzunov (approved):
I have no major change requests. Thanks for the good code style :)

hivemind/utils/grpc.py
justheuristic: [added this #188 ]

tests/test_allreduce.py
justheuristic: why?

tests/test_allreduce.py
justheuristic: This is required to ensure that compression takes long enough that eventloop can predictably switch

tests/test_allreduce.py
justheuristic: upd: nevermind, i'll update the code

tests/test_allreduce.py
justheuristic: see above

tests/test_allreduce.py
justheuristic: good point, i've added clarifications

hivemind/client/averaging/__init__.py
justheuristic: done

tests/test_allreduce.py
justheuristic: updated to the best of my ability

yhn112 (approved):

mryab (approved):

PR #288, opened 2021-06-24
Update readthedocs with hivemind.optim

yhn112:


yhn112:
Any ideas for improving the documentation are welcome

justheuristic (approved):
I would recommend merging this as is and thinking about next steps in a separate thread :)

docs/modules/optim.rst
```diff
@@ -0,0 +1,14 @@
+**hivemind.optim**
+==================
+
+.. automodule:: hivemind.optim
+.. currentmodule:: hivemind.optim
+
```
justheuristic, line 6:
```suggestion

.. raw:: html

  This module contains decentralized optimizers that wrap regular pytorch optimizers to collaboratively train a shared model. Depending on the specific type, optimizers may average model parameters with peers, exchange gradients, or follow a more complicated distributed training strategy.
  <br><br>
```

docs/modules/index.rst
```diff
@@ -5,7 +5,8 @@
...
    :maxdepth: 2
 
+   optim
    averaging
+   dht
    client
-   server
-   dht
\ No newline at end of file
+   server
```
mryab: Maybe sort alphabetically?

mryab (approved):

docs/modules/index.rst
yhn112: Nope.
At least, I would like to see client and server next to each other.
Also, for now, optim and averaging seem most useful for an average user, thus we begin with them.

PR #291, opened 2021-06-24
Fix Codecov integration with Github Actions

yhn112:


.github/workflows/run-tests.yml
```diff
@@ -33,10 +33,7 @@ jobs:
       - name: Test
         run: |
           cd tests
-          pytest --cov=hivemind 
-      - name: Upload to codecov
-        run: |
-          codecov
+          pytest --durations=0
```
mryab, line 36:
```suggestion
          pytest --durations=0 --durations-min=1.0
```

mryab (approved):

PR #295, opened 2021-06-29
Refactor P2P before DHT integration

borzunov:
This PR contains a few trivial changes made in `hivemind.p2p` along with P2P integration to the DHT:

- [x] Change misleading names in hivemind.p2p.P2P
- [x] Improve P2P interfaces and docs
- [x] Add missing type annotations in `p2p_daemon.py`
- [x] Implement missing `P2P.call_unary_handler()` method
- [x] Implement `P2P.identify_maddrs()` method

The changes are presented separately to shorten the future DHT PR. See comments in the diff for the motivation behind some changes.

### DIFF
--- a/hivemind/p2p/p2p_daemon.py
+++ b/hivemind/p2p/p2p_daemon.py
@@ -3,7 +3,7 @@ from copy import deepcopy
 from dataclasses import dataclass
 from importlib.resources import path
 from subprocess import Popen
-from typing import List, Optional
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import google.protobuf
 from multiaddr import Multiaddr
@@ -24,17 +24,12 @@ NUM_RETRIES = 3
 RETRY_DELAY = 0.4
 
 
-class P2PInterruptedError(Exception):
-    pass
-
-
-@dataclass(frozen=False)
+@dataclass(frozen=True)
 class P2PContext(object):
-    id: str
-    port: int
     handle_name: str
-    peer_id: PeerID = None
-    peer_addr: Multiaddr = None
+    local_id: PeerID
+    remote_id: PeerID = None
+    remote_maddr: Multiaddr = None
 
 
 class P2P:
@@ -68,10 +63,12 @@ class P2P:
     @classmethod
     async def create(cls, *args, quic: bool = True, tls: bool = True, conn_manager: bool = True,
                      dht_mode: str = 'dht_server', force_reachability: Optional[str] = None,
-                     nat_port_map: bool = True, auto_nat: bool = True, bootstrap: bool = False,
-                     bootstrap_peers: Optional[List[str]] = None, use_global_ipfs: bool = False, host_port: int = None,
+                     nat_port_map: bool = True, auto_nat: bool = True,
+                     bootstrap_peers: Optional[List[Multiaddr]] = None,
+                     use_ipfs: bool = False, p2p_port: int = None,
                      daemon_listen_port: int = None, use_relay: bool = True, use_relay_hop: bool = False,
-                     use_relay_discovery: bool = False, use_auto_relay: bool = False, relay_hop_limit: int = 0, **kwargs):
+                     use_relay_discovery: bool = False, use_auto_relay: bool = False, relay_hop_limit: int = 0,
+                     **kwargs) -> 'P2P':
         """
         Start a new p2pd process and connect to it.
         :param args:
@@ -83,30 +80,27 @@ class P2P:
         :param nat_port_map: Enables NAT port mapping
         :param auto_nat: Enables the AutoNAT service
         :param bootstrap: Connects to bootstrap peers and bootstraps the dht if enabled
-        :param bootstrap_peers: List of bootstrap peers; defaults to the IPFS DHT peers
-        :param use_global_ipfs: Bootstrap to global ipfs (works only if bootstrap=True and bootstrap_peers=None)
-        :param host_port: port for p2p network
+        :param bootstrap_peers: List of bootstrap peers
+        :param use_ipfs: Bootstrap to IPFS (works only if bootstrap=True and bootstrap_peers=None)
+        :param p2p_port: port for p2p network
         :param daemon_listen_port: port for connection daemon and client binding
         :param use_relay: enables circuit relay
         :param use_relay_hop: enables hop for relay
         :param use_relay_discovery: enables passive discovery for relay
         :param use_auto_relay: enables autorelay
         :param relay_hop_limit: sets the hop limit for hop relays
-        :param kwargs:
-        :return: new wrapper for p2p daemon
+        :param kwargs: other CLI arguments for the p2p daemon
+        :return: a wrapper for the p2p daemon
         """
 
-        assert not (bootstrap and bootstrap_peers is None and not use_global_ipfs), \
-            'Trying to create with bootstrap node without bootstrap nodes list. ' \
-            'It is very dangerous, because p2pd connects to global ipfs and it is very unstable. ' \
-            'If you really want this, pass use_global_ipfs=True'
-        assert not (bootstrap_peers is not None and use_global_ipfs), \
-            'Non empty bootstrap_nodes and use_global_ipfs=True are incompatible.' \
-            'Choose one option: your nodes list (preferable) or global ipfs (very unstable)'
+        assert not (bootstrap_peers is not None and use_ipfs), \
+            'User-defined bootstrap_peers and use_ipfs=True are incompatible, please choose one option'
 
         self = cls()
         with path(cli, P2PD_FILENAME) as p:
             p2pd_path = p
+
+        need_bootstrap = bootstrap_peers is not None or use_ipfs
         bootstrap_peers = cls._make_bootstrap_peers(bootstrap_peers)
         dht = cls.DHT_MODE_MAPPING.get(dht_mode, {'dht': 0})
         force_reachability = cls.FORCE_REACHABILITY_MAPPING.get(force_reachability, {})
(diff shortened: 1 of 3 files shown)


hivemind/p2p/p2p_daemon.py
```diff
@@ -28,13 +28,11 @@ class P2PInterruptedError(Exception):
...
-@dataclass(frozen=False)
+@dataclass
 class P2PContext(object):
-    id: str
-    port: int
     handle_name: str
-    peer_id: PeerID = None
-    peer_addr: Multiaddr = None
+    client_id: PeerID = None
+    client_maddr: Multiaddr = None
```
borzunov, line 35: Naming in `P2PContext` was misleading: `id` and `port` denoted the server (local) credentials, `peer_id` and `peer_addr` denoted the client (remote) credentials.

This change makes naming more clear and removes the local credentials since they are already accessible as `P2P` class fields.

hivemind/p2p/p2p_daemon.py
```diff
@@ -68,10 +66,12 @@ def __init__(self):
     @classmethod
     async def create(cls, *args, quic: bool = True, tls: bool = True, conn_manager: bool = True,
                      dht_mode: str = 'dht_server', force_reachability: Optional[str] = None,
-                     nat_port_map: bool = True, auto_nat: bool = True, bootstrap: bool = False,
-                     bootstrap_peers: Optional[List[str]] = None, use_global_ipfs: bool = False, host_port: int = None,
+                     nat_port_map: bool = True, auto_nat: bool = True,
+                     bootstrap_peers: Optional[List[Multiaddr]] = None,
+                     use_ipfs: bool = False, p2p_port: int = None,
```
borzunov, line 71: IPFS is always global, so this renames `use_global_ipfs` to `use_ipfs`.

hivemind/p2p/p2p_daemon.py
```diff
@@ -68,10 +66,12 @@ def __init__(self):
     @classmethod
     async def create(cls, *args, quic: bool = True, tls: bool = True, conn_manager: bool = True,
                      dht_mode: str = 'dht_server', force_reachability: Optional[str] = None,
-                     nat_port_map: bool = True, auto_nat: bool = True, bootstrap: bool = False,
```
borzunov, line 71: `bootstrap: bool` parameter does not make any sense: it is clear that the user wants to enable bootstrapping if either `bootstrap_peers` or `use_ipfs=True` is specified (and does not want to do it otherwise).

hivemind/p2p/p2p_daemon.py
```diff
@@ -68,10 +66,12 @@ def __init__(self):
...
-                     bootstrap_peers: Optional[List[str]] = None, use_global_ipfs: bool = False, host_port: int = None,
```
borzunov, line 72: `host_port` parameter is misleading:

1. A user may expect that it contains both a host and a port, while it actually contains only a port.
2. It is unclear which port it defines (external p2p port, local daemon port, or local daemon controller port).

I hope the name `p2p_port` solves these problems.

hivemind/p2p/p2p_daemon.py
```diff
@@ -336,17 +349,31 @@ async def call_peer_handler(self, peer_id, handler_name, input_data):
...
+        try:
+            await P2P.send_protobuf(request_protobuf, type(request_protobuf), writer)
+            result, err = await P2P.receive_protobuf(response_proto_type, reader)
+            if err is not None:
+                raise Exception(f'Failed to call unary handler {handler_name} at {peer_id}: {err}')
+
+            return result
+        finally:
+            writer.close()
+
```
borzunov, line 365: A method to call unary handlers was missing.

hivemind/p2p/p2p_daemon.py
```diff
@@ -28,13 +28,12 @@ class P2PInterruptedError(Exception):
     pass
 
 
-@dataclass(frozen=False)
+@dataclass(frozen=True)
 class P2PContext(object):
```
borzunov, line 32: Naming in `P2PContext` was misleading: before this change, `id` and `port` denoted the server (local) credentials, `peer_id` and `peer_addr` denoted the client (remote) credentials.

This change fixes the naming and removes the local port from the context (the p2p handlers should only operate with PeerIDs and Multiaddrs, the port is too internal concept for them).

hivemind/p2p/p2p_daemon.py
justheuristic: for some reason I thought we've renamed it already. Good catch :)

hivemind/p2p/p2p_daemon.py
```diff
@@ -83,30 +80,27 @@ async def create(cls, *args, quic: bool = True, tls: bool = True, conn_manager:
         :param nat_port_map: Enables NAT port mapping
         :param auto_nat: Enables the AutoNAT service
         :param bootstrap: Connects to bootstrap peers and bootstraps the dht if enabled
-        :param bootstrap_peers: List of bootstrap peers; defaults to the IPFS DHT peers
-        :param use_global_ipfs: Bootstrap to global ipfs (works only if bootstrap=True and bootstrap_peers=None)
-        :param host_port: port for p2p network
+        :param bootstrap_peers: List of bootstrap peers
+        :param use_ipfs: Bootstrap to IPFS (works only if bootstrap=True and bootstrap_peers=None)
+        :param p2p_port: port for p2p network
```
justheuristic, line 85: nit: i would argue that "port for p2p network" is still a bit misleading, aka "how does an entire network have a single port?"

```suggestion
        :param p2p_port: port for external connections from other p2p instances
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -134,11 +128,11 @@ async def create(cls, *args, quic: bool = True, tls: bool = True, conn_manager:
         return self
 
     @classmethod
-    async def replicate(cls, daemon_listen_port: int, host_port: int):
+    async def replicate(cls, daemon_listen_port: int, p2p_port: int) -> 'P2P':
         """
         Connect to existing p2p daemon
         :param daemon_listen_port: port for connection daemon and client binding
-        :param host_port: port for p2p network
+        :param p2p_port: port for p2p network
```
justheuristic, line 135: [nit] see reasoning above
```suggestion
        :param p2p_port: port for external connections from other p2p instances
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -147,15 +141,23 @@ async def replicate(cls, daemon_listen_port: int, host_port: int):
         # Use external already running p2pd
         self._child = None
         self._alive = True
-        self._assign_daemon_ports(host_port, daemon_listen_port)
+        self._assign_daemon_ports(p2p_port, daemon_listen_port)
         self._client_listen_port = find_open_port()
         self._client = p2pclient.Client(
             Multiaddr(f'/ip4/127.0.0.1/tcp/{self._daemon_listen_port}'),
             Multiaddr(f'/ip4/127.0.0.1/tcp/{self._client_listen_port}'))
         await self._wait_for_client()
```
justheuristic, line 149: [not expected for this PR] this may be vulnerable to random failures where multiple peers request the same port which is open when they check for it, but then closen when they actually need it.

justheuristic (approved):
image

__Another optional request:__ the current docstring is a lie (it spawns process instead of forking)

__Suggested change:__
```
This class is responsible for establishing peer-to-peer connections through NAT and/or firewalls.
It creates and manages a libp2p daemon a background process, then kills the process when P2P is shut down.
In order to communicate, a P2P instance must be given one or more initial peers that will connect it to the rest 
of the swarm. By default, hivemind.P2P instances connect using the public IPFS network ( https://ipfs.io ).

For incoming connections, P2P instances can add RPC handlers that are triggered by other peers
- `P2P.add_unary_handler` accepts a protobuf message and returns another protobuf
- `P2P.add_stream_handler` can use bi-directional streaming to send raw data

To access these handlers, a P2P instance can .call_unary/stream_handler, using the recipient's unique `P2P.id` 
and the name of the corresponding handler.
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -68,10 +63,12 @@ def __init__(self):
...
+                     nat_port_map: bool = True, auto_nat: bool = True,
+                     bootstrap_peers: Optional[List[Multiaddr]] = None,
+                     use_ipfs: bool = False, p2p_port: int = None,
                      daemon_listen_port: int = None, use_relay: bool = True, use_relay_hop: bool = False,
-                     use_relay_discovery: bool = False, use_auto_relay: bool = False, relay_hop_limit: int = 0, **kwargs):
+                     use_relay_discovery: bool = False, use_auto_relay: bool = False, relay_hop_limit: int = 0,
+                     **kwargs) -> 'P2P':
         """
         Start a new p2pd process and connect to it.
         :param args:
```
justheuristic, line 74:
```suggestion
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -68,10 +63,12 @@ def __init__(self):
     @classmethod
     async def create(cls, *args, quic: bool = True, tls: bool = True, conn_manager: bool = True,
```
justheuristic, line 64: param *args appears to be unused throughout P2P.create calls
```suggestion
    async def create(cls, *, quic: bool = True, tls: bool = True, conn_manager: bool = True,
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -315,38 +323,50 @@ async def stop_listening(self):
...
 
-    async def shutdown(self):
+    async def shutdown(self) -> None:
         await asyncio.get_event_loop().run_in_executor(None, self._terminate)
 
-    def _terminate(self):
+    def _terminate(self) -> None:
         self._alive = False
         if self._child is not None and self._child.poll() is None:
             self._child.kill()
```
justheuristic, line 372: nit: this method does not seem to have a good justification for SIGKILL. I would recommend using SIGTERM
```suggestion
            self._child.terminate()
```
The difference is that a SIGTERM gives the program a chance to close gracefully (closing files, network connections, freeing memory, etc), whereas SIGKILL doesn't.

justheuristic:
All my comments are optional, feel free to merge when ready.
Good job :)

hivemind/p2p/p2p_daemon.py
borzunov: I've used your comment and renamed it to `external_port` :)

hivemind/p2p/p2p_daemon.py
borzunov: They are used in the `self._make_process_args()` call.

Issue #275, opened 2021-06-15
[REFACTOR] general codebase updates for v0.10


justheuristic:
Extracted from #98 

- [x] Switch away from wildcard imports (from ... import *) in favor of explicit symbols
- [x] move functionality away from `__init__.py` files
- [x] move hivemind.client.averaging -> hivemind.averaging
   - [x] make sure this does not break application code
- [x] replace prints from benchmark_dht/averaging with logging.info
- [x] see if there are fixable warnings
   - tests still raise warning about non-copyable ndarray in (de)serialize_torch_tensor
- [ ] create an issue template for `[REFACTOR]`
- [x] (after libp2p integration) rename hivemind.utils.grpc -> hivemind.utils.protobuf ?
   -  Renamed to `hivemind.utils.streaming` in #470

borzunov:
Mostly done, v0.10 was released a long time ago :)

PR #298, opened 2021-06-30
Refactor MPFuture to use a single pipe/thread per process

justheuristic:
TODO list:
* [x] implement the new MPFuture backend
   * MPFuture no longer spawns pipe + thread for each unique future
   * MPFuture is now a single object instead of a linked pair
   * MPFuture now returns the same exception types as regular future (and as asyncio.Future in `__await__`)
   * MPFuture.result/exception can now only be awaited from the process that created it
   * MPFuture can no longer be used in disconnected processes (i.e. only forked/forkserver-ed)
* [x] Actually remove finished MPFuture instances from ACTIVE_FUTURES once it was assigned a terminal state
* [x] update MPFuture tests
* [x] update MPFuture usage in averager, dht, server
* [x] remove hivemind.run_in_background and HIVEMIND_THREADS from core
* [x] remove hivemind.run_in_background and HIVEMIND_THREADS from examples
* [x] remove HIVEMIND_THREADS from readme
* [x] create a separate executor for compression
* Extra tests:
    * [x] send MPFuture over pipe
    * [x] ensure MPFutures are deleted from ACTIVE_FUTURES
    * [x] test add_done_callback

__MPFuture no longer uses SyncManager because SyncManager is not fault tolerant; If one process fails (e.g. is terminated) during interaction with SyncManager, it will randomly hang other processes that access the same object.__

__Note 1:__
forking a process results in fork not having background threads:

image

Moreover, if one explicitly prints the background thread inside a forked process, it will be displayed as stopped (while in master it is started)

image

__

__Note 2:__
sending pipes over pipes is not free
image

This is **not** due to lengthy serialization or large size, but due to the need to open new files on de-serialization.

In contrast, creating a shared value is much faster
image

image


justheuristic:
__Evidence 1:__
sending pipes over pipes is wasteful
image

justheuristic:
__Note:__ we have found that forking process *somehow* results in fork not having background threads:

image

Moreover, if one explicitly prints the background thread inside a forked process, it will be displayed as stopped (while in master it is started)

image

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,198 @@
...
 
-from hivemind.utils.threading import run_in_background
 
-ResultType = TypeVar('ResultType')
+logger = get_logger(__name__)
 
+# flavour types
+ResultType = TypeVar('ResultType')
+PID, UID, State, PipeEnd = int, int, Any, mp.connection.Connection
+ALL_STATES = base.PENDING, base.RUNNING, base.FINISHED, base.CANCELLED, base.CANCELLED
```
borzunov, line 19: Typo: `base.CANCELLED` twice instead of `base.CANCELLED, base.CANCELLED_AND_NOTIFIED`

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,198 @@
...
+ResultType = TypeVar('ResultType')
+PID, UID, State, PipeEnd = int, int, Any, mp.connection.Connection
+ALL_STATES = base.PENDING, base.RUNNING, base.FINISHED, base.CANCELLED, base.CANCELLED_AND_NOTIFIED
+TERMINAL_STATES = {base.FINISHED, base.CANCELLED, base.CANCELLED_AND_NOTIFIED}
+
+INITIALIZER_LOCK = mp.Lock()
+PIPE_WAITER: Optional[threading.Thread] = None
+MPFUTURE_PIPES: Dict[PID, Tuple[PipeEnd, PipeEnd]] = mp.Manager().dict()
+ACTIVE_FUTURES: Optional[Dict[PID, MPFuture]] = None
+ACTIVE_PID: Optional[PID] = None
```
borzunov, line 26: The entities in this block (from `INITIALIZER_LOCK` to `ACTIVE_PID`) are global variables, not constants. Their names should not be capitalized (see [this](https://www.python.org/dev/peps/pep-0008/#id44) PEP-8 section).

I would also suggest to make them private static members of the `MPFuture` class along with the `_initialize_mpfuture_backend`, `_send_update`, and `_process_updates_in_background` methods below. This is because the single purpose of them is to make the `MPFuture` instances work.

hivemind/utils/mpfuture.py
```diff
@@ -4,169 +4,209 @@
...
 import multiprocessing.connection
-import time
-from functools import lru_cache
-from typing import Optional, Tuple, Generic, TypeVar
+import os
+import threading
+from typing import Tuple, Generic, TypeVar, Dict, Optional, Any, Callable
 
-from hivemind.utils.threading import run_in_background
+import torch
```
justheuristic, line 11: Note on torch: it is indeed weird, but so far we're still not sure how else to implement shared value for py3.7

Options:
- current version (with torch.empty)
- mp.Value or mp.Event - cannot send to other processes (cannot serialize)
- using [multiprocessing.shared_memory](https://docs.python.org/3/library/multiprocessing.shared_memory.html) - incompatible with py3.7 (and thus colab & kaggle kernels)
- using [_posixshmem](https://github.com/dillonalaird/shared_numpy/blob/master/shared_numpy/posixshmem.c) (extra dependency to requirements.txt)
- using mp.Pipe - back to where we started, will need an extra pipe per each future

hivemind/utils/mpfuture.py
justheuristic: [implementing as suggested]

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,248 @@
...
+PID, UID, State, PipeEnd = int, int, Any, mp.connection.Connection
+ALL_STATES = base.PENDING, base.RUNNING, base.FINISHED, base.CANCELLED, base.CANCELLED_AND_NOTIFIED
+TERMINAL_STATES = {base.FINISHED, base.CANCELLED, base.CANCELLED_AND_NOTIFIED}
 
-class MPFuture(base.Future, Generic[ResultType]):
-    """ Multiprocessing version of concurrent.futures.Future. Can also be awaited like asyncio.Future """
+try:
+    from concurrent.futures import InvalidStateError
+except ImportError:
+    class InvalidStateError(Exception):
```
borzunov, line 30:
```suggestion
    # Python 3.7 doesn't raise concurrent.futures.InvalidStateError for repeating set_result/set_exception calls and
    # doesn't even define this error. In this module, we simulate the Python 3.8+ behavior,
    # defining and raising this error if necessary.
    class InvalidStateError(Exception):
```

hivemind/client/averaging/training.py
```diff
@@ -47,55 +47,63 @@ def __init__(self, opt: torch.optim.Optimizer, *, average_parameters: bool, aver
             averaged_tensors = [tensor.detach().cpu().float().clone() for tensor in self.local_tensors()]
         super().__init__(averaged_tensors=averaged_tensors, **kwargs)
 
-    @torch.no_grad()
```
justheuristic, line 50: Why removed: because decorators are incompatible with our current RTFD parser

(this unlocks #288 )

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,251 @@
...
+class MPFuture(base.Future, Generic[ResultType]):
+    """
+    A version of concurrent.futures.Future / asyncio.Future that can be fulfilled from a separate process.
+    Any process can access future status and set the result / exception and check for state.
+    However, only the original process (i.e. the process that created the future) can await the result or exception.
+
+    :note: This is an internal primitive that is not guaranteed to work outside of hivemind applications.
+     More specifically, there are two known limitations:
+       - MPFuture works between processes created through inheritance (e.g. fork), *not* for independent processes
+       - Different executors (non-origin processes) cannot call set_result / set_exception / cancel simultaneously
```
borzunov, line 52: This limitation disappears since `mp.Lock` is used below.

hivemind/hivemind_cli/run_server.py
```diff
@@ -6,7 +6,7 @@
 
 from hivemind.proto.runtime_pb2 import CompressionType
 from hivemind.server import Server
-from hivemind.utils.threading import increase_file_limit
```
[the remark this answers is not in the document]
justheuristic: done

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,251 @@
...
-        self.connection = connection
+        self._use_lock = use_lock
 
-    @classmethod
-    def make_pair(cls) -> Tuple[MPFuture, MPFuture]:
-        """ Create a pair of linked futures to be used in two processes """
-        connection1, connection2 = mp.Pipe()
-        return cls(connection1), cls(connection2)
+        if self.active_pid != self._origin_pid:
+            self._initialize_mpfuture_backend()
```
borzunov, line 71: Consider a process that didn't have MPFutures before. If two MPFutures are created simultaneously in this process, it will lead to a race condition in `self._initialize_mpfuture_backend()`.

I suggest to use a lock and the [double-checked locking pattern](https://en.wikipedia.org/wiki/Double-checked_locking) here.

justheuristic: maybe it's easier to create a separate mp.Lock for initialization only?

hivemind/utils/__init__.py
```diff
@@ -7,5 +7,5 @@
 from hivemind.utils.networking import *
 from hivemind.utils.serializer import *
 from hivemind.utils.tensor_descr import *
-from hivemind.utils.threading import *
 from hivemind.utils.timed_storage import *
+from hivemind.utils.limits import *
```
borzunov, line 11: nit: Please sort them (so it's easy to check whether a particular module is included)

tests/test_averaging.py
```diff
@@ -417,9 +417,9 @@ def test_training_averager(n_steps: int = 10, n_dims: int = 16):
...
         f2.result()
 
-        assert torch.allclose(x1, x_avg)
-        assert torch.allclose(x2, x_avg)
-        assert torch.allclose(x1.grad, grad_avg)
-        assert torch.allclose(x2.grad, grad_avg)
+        assert torch.allclose(x1, x_avg, atol=1e-6, rtol=0)
+        assert torch.allclose(x2, x_avg, atol=1e-6, rtol=0)
+        assert torch.allclose(x1.grad, grad_avg, atol=1e-6, rtol=0)
+        assert torch.allclose(x2.grad, grad_avg, atol=1e-6, rtol=0)
```
borzunov, line 423: Don't forget to remove this before merge :)

tests/test_util_modules.py
```diff
@@ -1,129 +1,249 @@
...
-        f1.result(timeout=1e-3)
+    p = mp.Process(target=_proc, args=(future,))
+    p.start()
+    p.join()
 
-    f2.set_result(['abacaba', 123])
-    assert f1.result() == ['abacaba', 123]
+    assert future.result() == 321
+    assert future.cancel() is False
+    assert future.done() and not future.running() and not future.cancelled()
```
borzunov, line 36: Please test that `future.exception() is None` here.

tests/test_util_modules.py
```diff
@@ -1,129 +1,249 @@
...
-    f1, f2 = hivemind.MPFuture.make_pair()
-    with pytest.raises(TimeoutError):
-        f1.result(timeout=1e-3)
+    p = mp.Process(target=_proc, args=(future,))
+    p.start()
+    p.join()
 
-    f2.set_result(['abacaba', 123])
-    assert f1.result() == ['abacaba', 123]
+    assert future.result() == 321
```
borzunov, line 34:
```suggestion
    assert future.result() == 321
    assert future.exception() is None
```

hivemind/utils/mpfuture.py
justheuristic: updated the code with double-checked locking
thanks!

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,253 @@
...
-            pass
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            loop = None
+        if loop == self.get_loop():
+            asyncio.create_task(self._event_setter())
+        else:
+            asyncio.run_coroutine_threadsafe(self._event_setter(), self._loop)
+
+    async def _event_setter(self):
```
borzunov, line 111: nit: I suggest to move this into `_set_event_threadsafe` (define `_event_setter` just before it is used) to make clear that this short coroutine is only used inside this method.

In this case, `self` argument becomes unnecessary.

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,253 @@
...
+        return self.global_mpfuture_receivers[self._origin_pid]
+
+    @classmethod
+    def _initialize_mpfuture_backend(cls):
+        pid = os.getpid()
+        logger.debug(f"Initializing MPFuture backend for pid {pid}")
+        assert pid != cls.active_pid and pid not in cls.global_mpfuture_receivers, "already initialized"
+
+        recv_pipe, send_pipe = mp.Pipe(duplex=False)
+        cls.active_pid, cls.active_futures, cls.global_mpfuture_receivers[pid] = pid, {}, send_pipe
```
borzunov, line 126: nit: Should not it be `global_mpfuture_senders` if these are `send_pipe`s?

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,253 @@
...
+
+    :note: This is an internal primitive that is not guaranteed to work outside of hivemind applications.
+     More specifically, there are two known limitations:
+       - MPFuture works between processes created through inheritance (e.g. fork), *not* for independent processes
+       - Different executors (non-origin processes) cannot call set_result / set_exception / cancel simultaneously
+    """
+    lock = mp.Lock()  # global lock that prevents simultaneous initialization and writing
+    pipe_waiter_thread: Optional[threading.Thread] = None  # process-specific thread that receives results/exceptions
+    global_mpfuture_receivers: Dict[PID, PipeEnd] = mp.Manager().dict()
+    active_futures: Optional[Dict[PID, MPFuture]] = None  # pending or running futures originated from current process
```
borzunov, line 63:
```suggestion
    active_futures: Optional[Dict[UID, MPFuture]] = None  # pending or running futures originated from current process
```

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,253 @@
...
+        cls.pipe_waiter_thread.start()
+
+    @classmethod
+    def _process_updates_in_background(cls, receiver_pipe: mp.connection.Connection):
+        pid = os.getpid()
+        while True:
+            try:
+                uid, update_type, payload = receiver_pipe.recv()
+                if uid not in cls.active_futures:
+                    logger.debug(f"Ignoring update to future with uid={uid}: the future is no longer active.")
```
borzunov, line 138: This message uses the term "active future" that is not standard and not clear for a user. AFAIU, not active = already done or destroyed, so I suggest to rewrite it as:

```suggestion
                    logger.debug(f"Ignoring update to future with uid={uid}: the future is already done or destroyed.")
```

examples/albert/README.md
```diff
@@ -88,7 +88,7 @@ Here's an example of a full trainer script for Google Colab:
 !pip install transformers datasets sentencepiece torch_optimizer==0.1.0
 !git clone https://github.com/learning-at-home/hivemind && cd hivemind && pip install -e .
 !curl -L YOUR_HOSTED_DATA | tar xzf -     # example: https://hivemind-data.s3.us-east-2.amazonaws.com/wikitext103.tar.gz
-!ulimit -n 4096 && HIVEMIND_THREADS=256 python ./hivemind/examples/albert/run_trainer.py \
+!ulimit -n 4096 && python ./hivemind/examples/albert/run_trainer.py \
```
borzunov, line 91: Not related to this PR: maybe set `ulimit` inside `run_trainer.py` automatically to simplify the start command?

hivemind/utils/mpfuture.py
borzunov: If there is an easy way to distinguish the "already done" and "destroyed" states here, I'd suggest to do it and log more specific message (should be useful for debug).

Also, what do you think of making this a `logger.warning`? (asyncio's non-awaited tasks cause a warning in this case)

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
+                    raise RuntimeError(f"MPFuture received unexpected update type {update_type}")
+            except (BrokenPipeError, EOFError):
+                logger.debug(f"MPFuture backend was shut down (pid={pid}).")
+            except Exception as e:
+                logger.warning(f"Internal error (type={e}, pid={pid}): could not retrieve update for MPFuture.")
+                logger.exception(e)
+
+    def _send_update(self, update_type: UpdateType, payload: Any = None):
+        """ this method sends result, exception or cancel to the MPFuture origin. """
+        with self.lock if self._use_lock else contextlib.nullcontext():
```
borzunov, line 156: Can we use separate global locks for `_initialize_mpfuture_backend` and `_send_update`?

This may be beneficial since initialization is a much longer procedure - stopping all updates until it finishes may slow down the program.

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
+                elif update_type == UpdateType.RESULT:
+                    cls.active_futures.pop(uid).set_result(payload)
+                elif update_type == UpdateType.EXCEPTION:
+                    cls.active_futures.pop(uid).set_exception(payload)
+                elif update_type == UpdateType.CANCEL:
+                    cls.active_futures.pop(uid).cancel()
+                else:
+                    raise RuntimeError(f"MPFuture received unexpected update type {update_type}")
+            except (BrokenPipeError, EOFError):
+                logger.debug(f"MPFuture backend was shut down (pid={pid}).")
```
borzunov, line 149:
```suggestion
                logger.error(f"MPFuture backend was shut down (pid={pid}).")
```

This looks like something that should never happen, so we'd better to know about these situations ASAP.

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
+            except Exception as e:
+                logger.warning(f"Internal error (type={e}, pid={pid}): could not retrieve update for MPFuture.")
```
borzunov, line 151:
```suggestion
                logger.error(f"Internal error (type={e}, pid={pid}): could not retrieve update for MPFuture.")
```

nit: Changing this to match the log level of `logger.exception(e)` below.

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
+
+    def cancel(self):
+        if os.getpid() == self._origin_pid:
+            self.active_futures.pop(self._uid, None)
+            return super().cancel()
+        elif self._state in [base.RUNNING, base.FINISHED]:
+            return False
+        else:
+            self._state_cache[self._state] = base.CANCELLED
+            self._send_update(UpdateType.CANCEL)
```
borzunov, line 187:
```suggestion
            self._send_update(UpdateType.CANCEL)
            return True
```

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
+        if os.getpid() == self._origin_pid:
+            self.active_futures.pop(self._uid, None)
+            super().set_exception(exception)
+        elif self._state in TERMINAL_STATES:
+            raise InvalidStateError(f"Can't set_exception to a future that is {self._state} ({self._uid})")
+        else:
+            self._state_cache[self._state], self._exception = base.FINISHED, exception
+            self._send_update(UpdateType.EXCEPTION, exception)
+
+    def cancel(self):
```
borzunov, line 179:
```suggestion
    def cancel(self) -> bool:
```

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
-            return False
-        self._state, self._exception = base.CANCELLED, base.CancelledError()
-        return self._send_updates()
+            raise InvalidStateError(f"Can't set_running_or_notify_cancel when future is in {self._state} ({self._uid})")
 
     def result(self, timeout: Optional[float] = None) -> ResultType:
-        self._await_terminal_state(timeout)
-        if self._exception is not None:
+        if self._state not in TERMINAL_STATES:
+            assert os.getpid() == self._origin_pid, "only the process that created MPFuture can await result."
```
borzunov, line 200: nit: I'd suggest to remove dots in the end of the message and capitalize the first letter, so the message style is consistent across the library :)

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
     def cancelled(self):
-        self._sync_updates()
         return self._state == base.CANCELLED
 
-    def add_done_callback(self, callback):
-        raise NotImplementedError(f"MPFuture doesn't support callbacks.")
-
-    def remove_done_callback(self, callback):
-        raise NotImplementedError(f"MPFuture doesn't support callbacks.")
+    def add_done_callback(self, callback: Callable):
```
borzunov, line 226: I'd suggest to make the expected signature clear:

```suggestion
    def add_done_callback(self, callback: Callable[['MPFuture'], None]):
```

Quotes around `MPFuture` are for Python 3.7.

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
+            raise asyncio.CancelledError()
 
     def __del__(self):
-        self._shutdown_trigger.set_result(True)
-        if hasattr(self, 'connection'):
-            self.connection.close()
+        if getattr(self, '_origin_pid', None) == os.getpid():
+            self.active_futures.pop(self._uid, None)
+        if getattr(self, '_aio_event', None):
+            self._aio_event.set()
```
borzunov, line 246: Seems like `del` can be called from another thread.

```suggestion
            self._set_event_threadsafe()
```

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
+
+    :note: This is an internal primitive that is not guaranteed to work outside of hivemind applications.
+     More specifically, there are two known limitations:
+       - MPFuture works between processes created through inheritance (e.g. fork), *not* for independent processes
+       - MPFuture is deterministic if only one process can call set_result/set_exception/set_running_or_notify_cancel
+         and only the origin process can call result/exception/cancel.
+    """
+    lock = mp.Lock()  # global lock that prevents simultaneous initialization and writing
+    pipe_waiter_thread: Optional[threading.Thread] = None  # process-specific thread that receives results/exceptions
+    global_mpfuture_senders: Dict[PID, PipeEnd] = mp.Manager().dict()
```
borzunov, line 62:
```suggestion
    global_mpfuture_senders: MutableMapping[PID, PipeEnd] = mp.Manager().dict()
```

Since it is not a dict but only has a similar interface.

hivemind/utils/mpfuture.py
borzunov: I suggest to protect all interactions with `active_futures` with a `threading.Lock`. Otherwise, it seems to be not thread-safe (e.g. if several processes use even different `MPFuture` instances).

borzunov (changes requested):
I'd like to go through `MPFuture` one more after the comments are fixed :)

hivemind/utils/mpfuture.py
justheuristic: Please move to #275

examples/albert/README.md
justheuristic: Not always acceptable command

hivemind/utils/mpfuture.py
justheuristic: fixed

tests/test_util_modules.py
```diff
@@ -1,129 +1,299 @@
...
-    for future in [f1, f2]:
-        assert future.running() and not future.done() and not future.cancelled()
-        with pytest.raises(RuntimeError):
-            future.set_running_or_notify_cancel()
-    f2.cancel()
-    for future in [f1, f2]:
+    future = hivemind.MPFuture()
+
+    def _proc1(future):
+        assert future.set_running_or_notify_cancel() is True
```
borzunov, line 94: If this assert fails, we won't know about it.

justheuristic: we do, in the logs

tests/conftest.py
```diff
@@ -4,6 +4,7 @@
 
 @pytest.fixture(autouse=True, scope='session')
 def cleanup_children():
+    """ reset shared memory manager for isolation, terminate any leftover processes after the test is finished """
```
borzunov, line 7:
```suggestion
```

I suggest to revert the comment (shared memory is not reset anymore, terminating children is obvious from the code).

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
+            asyncio.run_coroutine_threadsafe(_event_setter(), self._loop)
+
+    @classmethod
+    def _initialize_mpfuture_backend(cls):
+        pid = os.getpid()
+        logger.debug(f"Initializing MPFuture backend for pid {pid}")
+        assert pid != cls._active_pid, "already initialized"
+
+        receiver_pipe, sender_pipe = mp.Pipe(duplex=False)
+        cls._active_pid, cls._active_futures, cls._global_sender_pipe = pid, {}, sender_pipe
```
borzunov, line 124:
```suggestion
        receiver_pipe, cls._global_sender_pipe = mp.Pipe(duplex=False)
        cls._active_pid, cls._active_futures = pid, {}
```

borzunov (approved):
The solution without `mp.Manager` seems much more elegant :)

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
-        return self._result
+        else:
+            return self._result
 
-    def exception(self, timeout=None) -> BaseException:
-        self._await_terminal_state(timeout)
-        if self._state == base.CANCELLED:
+    def exception(self, timeout: Optional[float] = None) -> BaseException:
+        if self._state not in TERMINAL_STATES:
+            assert os.getpid() == self._origin_pid, "only the process that created MPFuture can await exception."
```
mryab, line 209: IMHO, in such cases raising a specific error is [better](https://stefan.sofa-rockers.org/2018/04/16/assertions-and-exceptions/) than just an assertion. The motivation is that this is an external method and errors can happen [on the user side](https://stackoverflow.com/a/54252850) (basically we are trying to validate the environment), second reason is that it will be more informative and specific if one decides to handle this error.

tests/test_util_modules.py
```diff
@@ -1,129 +1,304 @@
...
+    assert isinstance(out[1], hivemind.MPFuture)
+    out[1].set_result(['we', 'need', 'to', 'go', 'deeper'])
+
+    p.join()
+    assert evt.is_set()
+
+
+def test_mpfuture_done_callback():
+    receiver, sender = mp.Pipe(duplex=False)
+    evt1, evt2, evt3, evt4, evt5 = (mp.Event() for _ in range(5))
```
mryab, line 218: Maybe just create a list of events and just refer to events[0]?

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
 
+from hivemind.utils.logging import get_logger
+
+
+logger = get_logger(__name__)
+
+# flavour types
 ResultType = TypeVar('ResultType')
+PID, UID, State, PipeEnd = int, int, Any, mp.connection.Connection
+ALL_STATES = base.PENDING, base.RUNNING, base.FINISHED, base.CANCELLED, base.CANCELLED_AND_NOTIFIED
```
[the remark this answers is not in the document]
justheuristic: Alas, concurrent.futures thinks otherwise and we have to be compatible

hivemind/utils/mpfuture.py
justheuristic: thx, fixed

tests/test_util_modules.py
justheuristic: updated the code

hivemind/utils/mpfuture.py
justheuristic: agreed, fixed

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,260 @@
...
-            if self._state in self.TERMINAL_STATES:
-                self._shutdown_trigger.set_result(True)
-                self.connection.close()
-            return True
-        except BrokenPipeError:
-            return False
+            self._loop = loop or asyncio.get_event_loop()
+            self._aio_event = asyncio.Event()
+        except RuntimeError:
+            self._loop = self._aio_event = None
```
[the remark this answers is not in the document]
justheuristic: assigned to None separately while keeping the number of lines, does it solve the issue?

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,262 @@
...
+import torch    # used for py3.7-compatible shared memory
 
+from hivemind.utils.logging import get_logger
+
+
+logger = get_logger(__name__)
+
+# flavour types
 ResultType = TypeVar('ResultType')
+PID, UID, State, PipeEnd = int, int, Any, mp.connection.Connection
```
yhn112, line 23: Are you sure that the narrowest type for `State` is `Any`?

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,262 @@
...
+    _global_sender_pipe: Optional[PipeEnd] = None  # a pipe that is used to send results/exceptions to this process
+    _pipe_waiter_thread: Optional[threading.Thread] = None  # process-specific thread that receives results/exceptions
+    _active_futures: Optional[Dict[UID, MPFuture]] = None  # pending or running futures originated from current process
+    _active_pid: Optional[PID] = None  # pid of currently active process; used to handle forks natively
 
-    TERMINAL_STATES = {base.FINISHED, base.CANCELLED, base.CANCELLED_AND_NOTIFIED}
+    def __init__(self, use_lock: bool = True, loop: Optional[asyncio.BaseEventLoop] = None):
+        self._origin_pid, self._uid = os.getpid(), uuid.uuid4().int
+        self._shared_state_code = torch.empty([], dtype=torch.uint8).share_memory_()
+        self._state_cache = {}  # mapping from global to cached local future used that makes updates immediately
```
yhn112, line 70:
```suggestion
        self._state_cache:  Dict[State, State] = {}  # mapping from global to cached local future used that makes updates immediately
```
If I'm not mistaken

yhn112 (changes requested):

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,262 @@
...
+       - MPFuture works between processes created through inheritance (e.g. fork), *not* for independent processes
+       - MPFuture is deterministic if only one process can call set_result/set_exception/set_running_or_notify_cancel
+         and only the origin process can call result/exception/cancel.
+    """
+    _initialization_lock = mp.Lock()  # global lock that prevents simultaneous initialization of two processes
+    _update_lock = mp.Lock()  # global lock that prevents simultaneous writing to the same pipe
+    _global_sender_pipe: Optional[PipeEnd] = None  # a pipe that is used to send results/exceptions to this process
+    _pipe_waiter_thread: Optional[threading.Thread] = None  # process-specific thread that receives results/exceptions
+    _active_futures: Optional[Dict[UID, MPFuture]] = None  # pending or running futures originated from current process
+    _active_pid: Optional[PID] = None  # pid of currently active process; used to handle forks natively
```
yhn112, line 65: Is it a really good practice, to use `None` as a default value, if it's never used, or even checked for?
At least for `_active_futures` it seems really simple to have `{}` as a default.

hivemind/utils/compression.py
```diff
@@ -17,6 +19,8 @@
 FP16_MAX = 65_504
 UINT8_RANGE = 256
 
+COMPRESSION_EXECUTOR = ThreadPoolExecutor(max_workers=int(os.environ.get("QUANTILE_COMPRESSION_THREADS", 128)))
```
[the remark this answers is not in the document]
mryab: Decided to leave that as is and refactor the entire compression module after discussion

hivemind/utils/mpfuture.py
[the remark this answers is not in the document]
mryab: Understood, let's keep it a tuple

hivemind/utils/mpfuture.py
mryab: Though mutable defaults are generally frowned upon in function arguments, perhaps in that case it's ok, since we use the global class field anyway

And for non-mutable ones, maybe just a type annotation would suffice (unless we need to explicitly check that is was not initialized)

hivemind/utils/mpfuture.py
mryab: I'd rather not complicate that, so your call @justheuristic

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,262 @@
...
-from functools import lru_cache
-from typing import Optional, Tuple, Generic, TypeVar
+import os
+import threading
+import uuid
+from enum import Enum, auto
+from typing import Generic, TypeVar, Dict, Optional, Any, Callable
 
-from hivemind.utils.threading import run_in_background
+import torch    # used for py3.7-compatible shared memory
```
mryab, line 14: Still would prefer explicit imports from torch

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,262 @@
...
+                elif update_type == UpdateType.EXCEPTION:
+                    cls._active_futures.pop(uid).set_exception(payload)
+                elif update_type == UpdateType.CANCEL:
+                    cls._active_futures.pop(uid).cancel()
+                else:
+                    raise RuntimeError(f"MPFuture received unexpected update type {update_type}")
+            except (BrokenPipeError, EOFError):
+                logger.debug(f"MPFuture backend was shut down (pid={pid})")
+            except Exception as e:
+                logger.exception(f"MPFuture: could not retrieve update: caught {repr(e)} (pid={pid})")
```
mryab, line 148: You don't need to specify the class that the log entry refers to, since our logging includes class and method names — in essence, you're saying the same thing twice

mryab (approved):

hivemind/utils/mpfuture.py
justheuristic: changed to str

hivemind/utils/mpfuture.py
justheuristic: Sorry, Max, i'm with @yhn112 on this one

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,262 @@
...
+        elif self._exception:
             raise self._exception
-        return self._result
+        else:
+            return self._result
 
-    def exception(self, timeout=None) -> BaseException:
-        self._await_terminal_state(timeout)
-        if self._state == base.CANCELLED:
+    def exception(self, timeout: Optional[float] = None) -> BaseException:
```
yhn112, line 207:
```suggestion
    def exception(self, timeout: Optional[float] = None) -> Optional[BaseException]:
```

hivemind/utils/mpfuture.py
```diff
@@ -2,171 +2,262 @@
...
+        if os.getpid() == self._origin_pid:
+            super().set_result(result)
+            MPFuture._active_futures.pop(self._uid, None)
+        elif self._state in TERMINAL_STATES:
+            raise InvalidStateError(f"Can't set_result to a future that is {self._state} ({self._uid})")
+        else:
+            self._state_cache[self._state], self._result = base.FINISHED, result
+            self._send_update(UpdateType.RESULT, result)
 
     def set_exception(self, exception: BaseException):
```
yhn112, line 165:
```suggestion
    def set_exception(self, exception: Optional[BaseException]):
```

hivemind/utils/mpfuture.py
justheuristic: We have a similar pattern/issue in other parts of the repo, let's stick to None for now and discuss global issue on the nearest meeting

hivemind/utils/mpfuture.py
justheuristic: the explicit version imports
`from torch import empty, uint8`

which somehow looks even more confusing :)

hivemind/utils/mpfuture.py
justheuristic: fixed

yhn112 (approved):

PR #299, opened 2021-06-30
Fix race condition while reserving ports in P2P

borzunov:
`hivemind.p2p.P2P` now uses 3 ports for its operations (external port for other P2P instances + 2 ports for daemon-client interaction). These ports are chosen before running `p2pd` using `find_open_ports()`.

This leads to a race condition because another P2P instance (or another program) may acquire the port just between it is chosen and p2pd is started. The race condition become extremely probable for tests and benchmarks launching many DHT instances concurrently (since each DHT launches its own P2P instance).

This PR fixes it in several steps:

- [x] Switch daemon-client interaction to Unix sockets with unique names (e.g. `/tmp/hivemind-p2pd-XXXXX.sock`).
- [x] If a user doesn't provide a preferred external port, do not use `find_open_port()` for the external port. Instead, run p2pd and let it bind to any free port, then fetch the resulting port number from the daemon.

It makes the following changes along the way:

- [x] Allow to specify the full host multiaddrs that define interfaces, protocols, and ports made available for the external peers (instead of just one port)
- [x] Disable QUIC by default (since it doesn't work without providing external port in advance), but add a test checking that it does work if the port is provided
- [x] Improves graceful shutdown of P2P instances
- [x] Stops ping retries on p2pd starting if the daemon dies

### DIFF
--- a/hivemind/p2p/p2p_daemon.py
+++ b/hivemind/p2p/p2p_daemon.py
@@ -70,11 +70,11 @@ class P2P:
         self._server_stopped = asyncio.Event()
 
     @classmethod
-    async def create(cls, *args, quic: bool = True, tls: bool = True, conn_manager: bool = True,
+    async def create(cls, *args, quic: bool = False, tls: bool = True, conn_manager: bool = True,
                      dht_mode: str = 'dht_server', force_reachability: Optional[str] = None,
                      nat_port_map: bool = True, auto_nat: bool = True,
                      bootstrap_peers: Optional[List[Multiaddr]] = None, use_ipfs: bool = False,
-                     external_port: Optional[int] = None,
+                     host_maddrs: Optional[List[Multiaddr]] = None,
                      use_relay: bool = True, use_relay_hop: bool = False,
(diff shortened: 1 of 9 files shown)


hivemind/p2p/p2p_daemon.py
```diff
@@ -108,55 +109,65 @@ async def create(cls, *args, quic: bool = True, tls: bool = True, conn_manager:
...
-        self._assign_daemon_ports(external_port, daemon_listen_port)
+            b=need_bootstrap, q=quiet, **{**bootstrap_peers, **dht, **force_reachability, **host_maddrs, **kwargs})
 
+        self._initialize(proc_args)
         for try_count in range(NUM_RETRIES):
             try:
-                self._initialize(proc_args)
-                await self._wait_for_client(RETRY_DELAY * (2 ** try_count))
+                await asyncio.sleep(RETRY_DELAY * (2 ** try_count))
+                await self._ping_client()
```
justheuristic, line 133: try:
    await asyncio.wait_for(self._ping_client, timeout=...)
except asyncio.TimeoutError:
    continue

justheuristic (approved):
Good job!


/* Probably a good idea to ditch async waiting in later versions */

PR #317, opened 2021-07-13
Remove shared memory from MPFuture, fix minor bugs

justheuristic:
- re-written MPFuture to use pipe-only communication instead of SharedMemory
  - rationale: each shared memory object is a file, using thousands of them floods the open files limit
  - one alternative was to group multple shared memory objects into one file with different offsets, but that would require heavier engineering
- fixed `coroutine set_event_threadsafe is never awaited`


justheuristic:
[not black-compatible yet]

hivemind/utils/mpfuture.py
```diff
@@ -46,6 +48,7 @@ class MPFuture(base.Future, Generic[ResultType]):
     Any process can access future status and set the result / exception and check for state.
     However, only the original process (i.e. the process that created the future) can await the result or exception.
 
+    :param synchronize: TODO
```
borzunov, line 51: Please don't forget about the todo :)

tests/test_util_modules.py
```diff
@@ -452,6 +458,8 @@ def test_split_parts():
 
 
 def test_generic_data_classes():
+    from hivemind.utils import ValueWithExpiration, HeapEntry, DHTExpiration
```
borzunov, line 461: Can we return these imports back to the beginning of the file?

hivemind/utils/mpfuture.py
justheuristic: thx, updated

tests/test_util_modules.py
justheuristic: thx, fixed

hivemind/utils/mpfuture.py
```diff
@@ -111,91 +101,169 @@ def _set_event_threadsafe(self):
...
+                future_ref = cls._active_futures.get(uid)
+                if future_ref is not None:
+                    future = future_ref()
+
+                if msg_type == MessageType.STATE_REQUEST:
+                    future_state = None if future is None else future.__getstate__()
+                    payload.send((uid, MessageType.STATE_RESPONSE, future_state))
+
+                elif msg_type == MessageType.STATE_RESPONSE:
+                    future, state_updated_event = cls._status_requests.get(uid) or (None, None)
```
mryab, line 145:
```suggestion
                    future, state_updated_event = cls._status_requests.get(uid, (None, None))
```

hivemind/utils/mpfuture.py
```diff
@@ -111,91 +101,169 @@ def _set_event_threadsafe(self):
...
+                    )
+                elif msg_type == MessageType.RESULT:
+                    future.set_result(payload)
+                elif msg_type == MessageType.EXCEPTION:
+                    future.set_exception(payload)
+                elif msg_type == MessageType.RUNNING:
+                    try:
+                        future.set_running_or_notify_cancel()
+                    except (InvalidStateError, RuntimeError) as e:
+                        logger.debug(f"could set MPFuture (uid={uid}) to running due to {e}")
```
mryab, line 167:
```suggestion
                        logger.debug(f"Could not set MPFuture (uid={uid}) to running due to {e}", exc_info=True)
```

hivemind/utils/mpfuture.py
```diff
@@ -111,91 +101,169 @@ def _set_event_threadsafe(self):
...
-    def _send_update(self, update_type: UpdateType, payload: Any = None):
+    def _send_update(self, update_type: MessageType, payload: Any = None):
         """This method sends result, exception or cancel to the MPFuture origin."""
-        with MPFuture._update_lock if self._use_lock else nullcontext():
-            self._sender_pipe.send((self._uid, update_type, payload))
+        try:
+            with MPFuture._update_lock if self._use_lock else nullcontext():
+                self._sender_pipe.send((self._uid, update_type, payload))
+        except (ConnectionError, BrokenPipeError, EOFError) as e:
+            logger.debug(f"No updates were sent: pipe to origin process is no longer operational ({e}).")
```
mryab, line 187:
```suggestion
            logger.debug(f"No updates were sent: pipe to origin process is no longer operational ({e}).", exc_info=True)
```

hivemind/utils/mpfuture.py
```diff
@@ -111,91 +101,169 @@ def _set_event_threadsafe(self):
...
+                self._sender_pipe.send((self._uid, MessageType.STATE_REQUEST, self._process_wide_pipe))
+            status_updated.wait(MPFuture.SOFT_UPDATE_TIMEOUT)
+            if not status_updated.is_set():
+                logger.warning(f"Status update took over {MPFuture.SOFT_UPDATE_TIMEOUT}, expect performance issues")
+                status_updated.wait(MPFuture.HARD_UPDATE_TIMEOUT - MPFuture.SOFT_UPDATE_TIMEOUT)
+                if not status_updated.is_set():
+                    self.set_exception(
+                        TimeoutError(
+                            f"Status update took over {MPFuture.HARD_UPDATE_TIMEOUT} seconds, "
+                            f"mpfuture is cancelled"
```
mryab, line 216:
```suggestion
                            f"MPFuture is cancelled"
```

hivemind/utils/mpfuture.py
```diff
@@ -46,6 +48,7 @@ class MPFuture(base.Future, Generic[ResultType]):
     Any process can access future status and set the result / exception and check for state.
     However, only the original process (i.e. the process that created the future) can await the result or exception.
 
+    :param synchronize: if True (default), future will request state from origin, otherwise it will only use local state
```
borzunov, line 51: nit: Please clarify here that it can be used if the future is guaranteed to never be cancelled from the origin side and leads to a minor speedup.

hivemind/utils/mpfuture.py
```diff
@@ -60,49 +63,36 @@ class MPFuture(base.Future, Generic[ResultType]):
...
-        self._shared_state_code = torch.empty([], dtype=torch.uint8).share_memory_()
-        self._state_cache: Dict[State, State] = {}
-        # mapping from global to cached local future used that makes updates immediately
-        # available on setter side; dictionary-based cache works because future can visit any state at most once
+    SOFT_UPDATE_TIMEOUT = 0.1  # seconds spent awaiting status update before warning is printed
+    HARD_UPDATE_TIMEOUT = 10.0  # seconds spent awaiting status update before future is automatically cancelled
 
-        base.Future.__init__(self)  # parent init is deferred because it uses self._shared_state_code
+    def __init__(self, synchronize: bool = True, use_lock: bool = True, loop: Optional[asyncio.BaseEventLoop] = None):
+        base.Future.__init__(self)
```
borzunov, line 76:
```suggestion
        super().__init__(self)
```

mryab (approved):

hivemind/utils/mpfuture.py
```diff
@@ -60,49 +63,36 @@ class MPFuture(base.Future, Generic[ResultType]):
...
-
-    @_state.setter
-    def _state(self, new_state: State):
-        self._shared_state_code[...] = ALL_STATES.index(new_state)
-        if self._state in TERMINAL_STATES and self._loop is not None and not self._aio_event.is_set():
-            self._set_event_threadsafe()
-
-    def _set_event_threadsafe(self):
+    def _set_event_if_necessary(self):
+        if self._aio_event is None or self._aio_event.is_set():
```
borzunov, line 94: `self._aio_event.is_set()` is not thread-safe.

hivemind/utils/mpfuture.py
```diff
@@ -60,49 +63,36 @@ class MPFuture(base.Future, Generic[ResultType]):
...
-        self._origin_pid, self._uid = os.getpid(), uuid.uuid4().int
-        self._shared_state_code = torch.empty([], dtype=torch.uint8).share_memory_()
-        self._state_cache: Dict[State, State] = {}
-        # mapping from global to cached local future used that makes updates immediately
-        # available on setter side; dictionary-based cache works because future can visit any state at most once
+    SOFT_UPDATE_TIMEOUT = 0.1  # seconds spent awaiting status update before warning is printed
+    HARD_UPDATE_TIMEOUT = 10.0  # seconds spent awaiting status update before future is automatically cancelled
 
-        base.Future.__init__(self)  # parent init is deferred because it uses self._shared_state_code
+    def __init__(self, synchronize: bool = True, use_lock: bool = True, loop: Optional[asyncio.BaseEventLoop] = None):
```
borzunov, line 75:
```suggestion
    def __init__(self, *, synchronize: bool = True, use_lock: bool = True, loop: Optional[asyncio.BaseEventLoop] = None):
```

Since everything may be interpreted as `bool` in Python, a user may pass anything (e.g. `loop`) as the first argument by mistake and see no exceptions.

hivemind/utils/mpfuture.py
```diff
@@ -111,91 +101,169 @@ def _set_event_threadsafe(self):
...
+
+                elif msg_type == MessageType.STATE_RESPONSE:
+                    future, state_updated_event = cls._status_requests.get(uid, (None, None))
+                    if future is None:
+                        logger.debug("Received a state update for a future that does not await status update.")
+                    else:
+                        if payload is not None:
+                            future.__setstate__(payload)
+                        else:
+                            base.Future.cancel(future)
```
borzunov, line 152:
```suggestion
                            super().cancel(future)
```

hivemind/utils/mpfuture.py
```diff
@@ -111,91 +101,169 @@ def _set_event_threadsafe(self):
...
+        if maybe_existing_request is not None:
+            _, status_updated = maybe_existing_request
+            status_updated.wait(MPFuture.HARD_UPDATE_TIMEOUT)
+            return
+
+        # otherwise create a new request for synchronization
+
+        try:
+            status_updated = threading.Event()
+            self._status_requests[self._uid] = (self, status_updated)
```
borzunov, line 205: Concurrent calls to `_synchronize_if_necessary` may lead to overwriting an existing `threading.Event` with a new one here. In this case, the older event will never get an update and will break the future after reaching `HARD_UPDATE_TIMEOUT`.

hivemind/utils/mpfuture.py
justheuristic: decided to keep is_set for performance reasons

borzunov: [It's thread-safe in the current implementation, and we've agreed that it's hard to imagine that it will stop to be thread-safe.]

hivemind/utils/mpfuture.py
```diff
@@ -111,91 +103,171 @@ def _set_event_threadsafe(self):
...
+                        args=[receiver_pipe],
+                        name=f"{__name__}.BACKEND",
+                        daemon=True,
+                    )
+                    cls._pipe_waiter_thread.start()
 
     @classmethod
     def _process_updates_in_background(cls, receiver_pipe: mp.connection.Connection):
         pid = os.getpid()
         while True:
```
borzunov, line 134: nit: This loop never stops gracefully. Ideally, it would be cool to stop it via a `threading.Event` in the `__del__` finalizer of the last active future in the current process.

However, graceful shutdown may be out of scope of this PR.

hivemind/utils/mpfuture.py
```diff
@@ -111,91 +103,171 @@ def _set_event_threadsafe(self):
...
-                    cls._active_futures.pop(uid).cancel()
+                uid, msg_type, payload = receiver_pipe.recv()
+                future = None
+                future_ref = cls._active_futures.get(uid)
+                if future_ref is not None:
+                    future = future_ref()
+
+                if msg_type == MessageType.STATE_REQUEST:
+                    future_state = None if future is None else future.__getstate__()
+                    payload.send((uid, MessageType.STATE_RESPONSE, future_state))
```
borzunov, line 144: We use the `payload` pipe without lock here.

hivemind/utils/mpfuture.py
```diff
@@ -111,91 +103,171 @@ def _set_event_threadsafe(self):
...
+        self._synchronize_if_necessary()
+        try:
+            is_running = super().set_running_or_notify_cancel()
+            if is_running and os.getpid() != self._origin_pid:
+                self._send_update(MessageType.RUNNING)
+            return is_running
+        except RuntimeError as e:
+            raise InvalidStateError(str(e))
 
     def result(self, timeout: Optional[float] = None) -> ResultType:
```
borzunov, line 272: Can we replace the part from `elif self._state == base.CANCELLED:` with just `return super().result()`?

hivemind/utils/mpfuture.py
```diff
@@ -219,12 +291,15 @@ def exception(self, timeout: Optional[float] = None) -> Optional[BaseException]:
         return self._exception
```
borzunov, line 291: Can we replace the part from `elif self._state == base.CANCELLED:` with just `return super().exception()`?

hivemind/utils/mpfuture.py
justheuristic: Fixed, thx!

hivemind/utils/mpfuture.py
justheuristic: good catch

hivemind/utils/mpfuture.py
```diff
@@ -60,49 +64,37 @@ class MPFuture(base.Future, Generic[ResultType]):
...
-        # mapping from global to cached local future used that makes updates immediately
-        # available on setter side; dictionary-based cache works because future can visit any state at most once
+    SOFT_UPDATE_TIMEOUT = 0.1  # seconds spent awaiting status update before warning is printed
+    HARD_UPDATE_TIMEOUT = 10.0  # seconds spent awaiting status update before future is automatically cancelled
 
-        base.Future.__init__(self)  # parent init is deferred because it uses self._shared_state_code
+    def __init__(self, *, synchronize: bool = True, use_lock: bool = True):
+        super().__init__()
+        self.synchronize = synchronize
+        self._synchronization_lock = threading.Lock()
```
borzunov, line 79:
```suggestion
```

The lock is now unused.

borzunov (approved):

hivemind/utils/mpfuture.py
```diff
@@ -111,120 +102,197 @@ def _set_event_threadsafe(self):
...
+        if self._state in TERMINAL_STATES:
             raise InvalidStateError(f"Can't set_result to a future that is {self._state} ({self._uid})")
+        elif os.getpid() == self._origin_pid:
+            MPFuture._active_futures.pop(self._uid, None)
+            self._set_event_if_necessary()
         else:
-            self._state_cache[self._state], self._result = base.FINISHED, result
-            self._send_update(UpdateType.RESULT, result)
+            self._send_update(MessageType.RESULT, result)
+        super().set_result(result)
```
borzunov, line 241: nit: In Python 3.7, this may raise `RuntimeError` (instead of `InvalidStateError`) if `set_result` is called concurrently inside one process.

This is a minor issue and we've agreed it won't be fixed.

PR #318, opened 2021-07-14
Implement protobuf-based stream handlers over libp2p backend

borzunov:
**Current status:** Finished.

This PR implements protobuf-based stream handlers over libp2p backend (including unary-stream, stream-unary, and stream-stream). Similarly to gRPC, they can be used through the `Servicer` interface:

```python
class ExampleServicer(ServicerBase):
    async def rpc_sum(self, request: AsyncIterator[TestRequest], _: P2PContext) -> TestResponse:
        result = 0
        async for item in request:
            result += item.number
        return TestResponse(number=result)

    async def rpc_count(self, request: TestRequest, _: P2PContext) -> AsyncIterator[TestResponse]:
        for i in range(request.number):
            yield TestResponse(number=i)
```

__Future work:__

- [ ] Consider adding timeout to a server waiting for the request data from client. May be useful to prevent resource leaks in case of hard client disconnects.

### DIFF
--- a/hivemind/dht/protocol.py
+++ b/hivemind/dht/protocol.py
@@ -7,7 +7,7 @@ from typing import Optional, List, Tuple, Dict, Sequence, Union, Collection
 from hivemind.dht.crypto import DHTRecord, RecordValidatorBase
 from hivemind.dht.routing import RoutingTable, DHTID, BinaryDHTValue, Subkey
 from hivemind.dht.storage import DHTLocalStorage, DictionaryDHTValue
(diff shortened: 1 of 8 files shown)


justheuristic (approved):

borzunov:
### benchmark_dht.py

**nora, this PR:**

Screen Shot 2021-07-15 at 6 47 42 AM

**nora, master** (197666c2b6baa479ec782556c47a9f33de7b463f):

...

Issue #276, opened 2021-06-15
[REFACTOR] updates to DHT internals


justheuristic:
extracted from #98 

* [ ] Can we get rid of return_futures=True option in node.get/store/run_coro ?
   - replace return_future ->  wait=True
   - or 2nd method get_async / store_async

* [ ]  rewrite test_dht_protocol, test_empty_table and test_dht_node to use await instead of loop.run_until_complete

* [ ] make DHT resistant to KeyboardInterrupt when they are running interactively?
  - problem symptom: you're running in jupyter, you press ctrl+C and it kills all dht daemons
  - applies to DecentralizedAverager (fix or create an issue)

* [ ] TimedStorage: add .pop() method

* [ ]  MPFuture: make sure await asyncio.create_task(mpfuture) works without extra wrappers

* [ ] make alias DHTID.from(x) = DHTID.generate(source=x)? (@justheuristic )

justheuristic:
**Note:** we decided not to implement TimedStorage.pop: neither of its 5 use cases would actually benefit from this method: all of them make deletion conditional, i.e. if item = s.pop(); if time_is_right: del s[key].
![image](https://user-images.githubusercontent.com/3491902/130221377-13639210-2adb-4dce-ba1a-927154067f61.png)

PR #323, opened 2021-07-15
Convert averager to libp2p backend

borzunov:
**Current status:** Finished.

**What I have tested:**

- [x] Training (1 monitor + 2 trainers) works, peers average successfully.

**TODO:**

- [x] Provide benchmark results (default and for large tensors)
- [x] Add `experiment_prefix` to handler name for averager RPCs to enable using several different averagers simultaneously
- [ ] ~~Follow-up PR: rename PeerID -> Endpoint, use `bytes` for PeerIDs in protobufs (instead of `string`)~~ (moved to https://github.com/learning-at-home/hivemind/issues/276 )


benchmarks/benchmark_averaging.py
```diff
@@ -57,11 +57,10 @@ def run_averager(index):
         dht = hivemind.DHT(initial_peers=initial_peers, start=True)
         initial_bits = bin(index % num_groups)[2:].rjust(nbits, "0")
         averager = hivemind.averaging.DecentralizedAverager(
-            peer_tensors[i],
+            peer_tensors[index],
```
borzunov, line 60: This fixes a bug in `master` (previously, all tensors were the same because all threads used the same `i` value from the closure).

hivemind/p2p/servicer.py
```diff
@@ -41,35 +42,65 @@ class ServicerBase:
       to calls to the remote peer.
     """
```
borzunov, line 44: The changes to `Servicer` are:

1. Support classes `ServicerB` inherited like `ServicerBase -> ServicerA -> ServicerB`
2. Make `get_stub()` a classmethod, so you don't need to a servicer yourself to call a method of another peer (used in tests)
3. Improve reflection:
  - Allow any parameter name for `TInputProtobuf` in `rpc_*` methods (previously, only `request` was allowed)
  - Use more robust type annotation parsing (`typing.get_type_hints()` instead of manually parsing string annotations in some cases)

tests/test_training.py
```diff
@@ -200,7 +200,7 @@ def test_decentralized_optimizer_averaging():
     (param1.sum() + param2.sum()).backward()
 
     for _ in range(100):
-        time.sleep(0.01)
+        time.sleep(0.1)
```
borzunov, line 203: See the comment in `test_allreduce_grid()`.

.github/workflows/run-tests.yml
```diff
@@ -60,12 +60,12 @@ jobs:
...
         run: |
           cd tests
-          pytest -k "p2p" 
+          pytest -k "p2p"
 
   codecov_in_develop_mode:
 
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
```
borzunov, line 68: This is because we need to create many DHT instances (=> P2P instances) for averagers in tests. Another solution to consider is making the number of instances smaller.

.github/workflows/run-tests.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         python-version: [ 3.7, 3.8, 3.9 ]
-    timeout-minutes: 10
+    timeout-minutes: 12
```
borzunov, line 13: This is because we need to create many DHT instances (=> P2P instances) for averagers in tests. Another solution to consider is making the number of instances smaller.

justheuristic: Do we still need 12min if we drop out test_overcrowded?

hivemind/averaging/averager.py
```diff
@@ -138,19 +121,14 @@ def __init__(
 
         super().__init__()
         self.dht = dht
-        self.client_mode, self.listen_on, self.kwargs = client_mode, listen_on, kwargs
+        self.client_mode = client_mode
```
justheuristic, line 124: nit: perhaps we should set default client_mode=None

and
```python
if client_mode is None:
    client_mode = dht.client_mode
```
?

hivemind/averaging/matchmaking.py
```diff
@@ -169,20 +171,21 @@ async def request_join_group(self, leader: Endpoint, expiration_time: DHTExpirat
...
-                        endpoint=self.endpoint,
+                        endpoint=self.endpoint.to_base58(),
                         schema_hash=self.schema_hash,
                         expiration=expiration_time,
                         client_mode=self.client_mode,
                         gather=self.data_for_gather,
                     )
-                )
-                message = await asyncio.wait_for(call.read(), timeout=self.request_timeout)
+                ).__aiter__()
```
justheuristic, line 187: please use [aiter from hivemind.utils.asyncio](https://github.com/learning-at-home/hivemind/blob/master/hivemind/utils/asyncio.py#L31-L34)

hivemind/averaging/matchmaking.py
```diff
@@ -198,51 +201,52 @@ async def request_join_group(self, leader: Endpoint, expiration_time: DHTExpirat
...
             self.was_accepted_to_group.clear()
             self.current_leader = None
-            if call is not None:
-                await call.code()
+            if stream is not None:
+                await stream.aclose()
 
     async def rpc_join_group(
-        self, request: averaging_pb2.JoinRequest, context: grpc.ServicerContext
+        self, request: averaging_pb2.JoinRequest, _: P2PContext
```
justheuristic, line 237: nit: let's ask @mryab & @yhn112 on the best practice for naming unused __function args__ as _

hivemind/dht/__init__.py
```diff
@@ -266,6 +282,19 @@ def get_visible_maddrs(self, latest: bool = False) -> List[Multiaddr]:
     async def _get_visible_maddrs(self, node: DHTNode, latest: bool = False) -> List[Multiaddr]:
         return await node.get_visible_maddrs(latest=latest)
 
+    async def replicate_p2p(self) -> P2P:
+        """
+        Get a replica of a P2P instance used in the DHT process internally.
```
justheuristic, line 287:
```suggestion
        Get a replica of a P2P instance used in the DHT process internally.
        The replica uses the same backend as the main DHT and only works while DHT is alive.
```

tests/test_averaging.py
```diff
@@ -251,6 +255,10 @@ def test_allreduce_grid():
     prev_means, prev_stds = means0, stds0
 
     for i in range(5):
+        time.sleep(0.2)
+        # FIXME (@justheuristic): An averager does not have time to update the key in GroupKeyManager
+        # if .step() is called without any delays
```
justheuristic, line 260: acknowledged)

justheuristic (approved):

examples/albert/arguments.py
```diff
@@ -45,10 +45,6 @@ class AveragerArguments:
     averaging_timeout: float = field(
         default=30.0, metadata={"help": "Give up on averaging step after this many seconds"}
     )
-    listen_on: str = field(
```
justheuristic, line 48: Request: before merging, please also check that this argument is removed from tutorials/examples

borzunov: I've checked this, all remaining occurrences of `listen_on` are related to MoE.

hivemind/averaging/matchmaking.py
borzunov: Currently, `hivemind.utils.asyncio.aiter()` is not related to `__aiter__()` and does something different :) We should probably rename it and create the real `aiter()` (calling `__aiter__()`) instead.

hivemind/averaging/averager.py
borzunov: Sounds like a good idea since it is easy to set this flag for the DHT and forget to also set it for the averager.

I've implemented this.

borzunov:
We suspect #317 to be the reason of periodic test failures. Now, we are testing a version with this commit reverted in a separate branch.

hivemind/averaging/allreduce.py
```diff
@@ -136,46 +154,35 @@ async def _communicate_with_peer(self, peer_endpoint: Endpoint):
...
         )
         async for part in parts_aiter:
-            await stream.write(averaging_pb2.AveragingData(tensor_part=part))
-
-        await stream.done_writing()
+            yield averaging_pb2.AveragingData(tensor_part=part)
 
     async def rpc_aggregate_part(
-        self, stream: AsyncIterator[averaging_pb2.AveragingData], context: grpc.ServicerContext
+        self, stream: AsyncIterator[averaging_pb2.AveragingData], _: P2PContext
```
mryab, line 185: IMO it's also better to have meaningful variable names in such cases, one possible option (if the linter nags you) is to precede a name with an underscore

hivemind/averaging/averager.py
```diff
@@ -539,7 +492,7 @@ async def _declare_for_download_periodically(self):
             await asyncio.sleep(self._matchmaking.averaging_expiration)
 
     async def rpc_download_state(
-        self, request: averaging_pb2.DownloadRequest, context: grpc.ServicerContext
+        self, request: averaging_pb2.DownloadRequest, _: P2PContext
```
mryab, line 495: Same here with `_context`

hivemind/averaging/averager.py
```diff
@@ -138,19 +122,19 @@ def __init__(
 
         super().__init__()
         self.dht = dht
-        self.client_mode, self.listen_on, self.kwargs = client_mode, listen_on, kwargs
+        self.prefix = prefix
+
+        if client_mode is None:
+            client_mode = dht.client_mode
+        self.client_mode = client_mode
```
mryab, line 129: Maybe rewrite to one line, like `self.client_mode if client_mode is not None else dht.client_mode`?

hivemind/p2p/servicer.py
```diff
@@ -98,25 +109,41 @@ async def caller(
...
+                self._get_handle_name(namespace, handler.method_name),
                 getattr(servicer, handler.method_name),
                 handler.request_type,
                 stream_input=handler.stream_input,
             )
 
-    def get_stub(self, p2p: P2P, peer: PeerID) -> StubBase:
-        return self._stub_type(p2p, peer)
+    @classmethod
+    def get_stub(cls, p2p: P2P, peer: PeerID, *, namespace: Optional[str] = None) -> StubBase:
```
mryab, line 137: Can you please explain the purpose of namespaces for RPCs? Are they used to avoid possible collisions for remote calls?

hivemind/averaging/matchmaking.py
mryab: Since this will break for multiple arguments, I propose to use something like [here](https://stackoverflow.com/a/47061782) and to precede arguments with an underscore while keeping their name for clarity

hivemind/averaging/key_manager.py
```diff
@@ -93,11 +94,11 @@ async def get_averagers(self, group_key: GroupKey, only_active: bool) -> List[Tu
             logger.debug(f"Allreduce group not found: {group_key}, creating new group.")
             return []
         averagers = [
-            (key, entry.expiration_time)
+            (Endpoint.from_base58(key), entry.expiration_time)
             for key, entry in result.value.items()
             if key != self.RESERVED_KEY_FOR_NBITS and (not only_active or entry.value is True)
         ]
-        num_active_averagers = len([key for key, entry in result.value.items() if entry.value is True])
+        num_active_averagers = sum(1 for entry in result.value.values() if entry.value is True)
```
mryab, line 101: Also, `entry` is slightly unclear here (that's why the RHS expression is hard to decipher on its own) — consider renaming

hivemind/dht/__init__.py
```diff
@@ -254,6 +263,24 @@ def add_validators(self, record_validators: Iterable[RecordValidatorBase]) -> No
...
         node.protocol.record_validator.extend(record_validators)
 
+    @property
+    def peer_id(self) -> PeerID:
+        if self._peer_id is None:
+            self._peer_id = self.run_coroutine(DHT._get_peer_id)
+        return self._peer_id
+
+    async def _get_peer_id(self, node: DHTNode) -> PeerID:
+        return node.peer_id
```
mryab, line 273: Can this be a static method?

mryab:
```suggestion
    @staticmethod
    async def _get_peer_id(dht: DHT, node: DHTNode) -> PeerID:
        return node.peer_id
```
Rewriting it as follows gets rid of the typing error

hivemind/dht/__init__.py
```diff
@@ -254,6 +263,24 @@ def add_validators(self, record_validators: Iterable[RecordValidatorBase]) -> No
...
+
+    @property
+    def client_mode(self) -> bool:
+        if self._client_mode is None:
+            self._client_mode = self.run_coroutine(DHT._get_client_mode)
+        return self._client_mode
+
+    async def _get_client_mode(self, node: DHTNode) -> bool:
+        return node.protocol.client_mode
```
mryab, line 282: Also can be made a static method

hivemind/dht/__init__.py
```diff
@@ -254,6 +263,24 @@ def add_validators(self, record_validators: Iterable[RecordValidatorBase]) -> No
...
+            self._peer_id = self.run_coroutine(DHT._get_peer_id)
+        return self._peer_id
+
+    async def _get_peer_id(self, node: DHTNode) -> PeerID:
+        return node.peer_id
+
+    @property
+    def client_mode(self) -> bool:
+        if self._client_mode is None:
+            self._client_mode = self.run_coroutine(DHT._get_client_mode)
```
mryab, line 278: By the way, can't we just call `await self._node.protocol.client_mode`? Is that because they're in different processes?

hivemind/dht/__init__.py
```diff
@@ -266,6 +293,20 @@ def get_visible_maddrs(self, latest: bool = False) -> List[Multiaddr]:
...
+        Get a replica of a P2P instance used in the DHT process internally.
+        The replica uses the same P2P daemon as the DHT and only works while DHT is alive.
+        """
+
+        if self._p2p_replica is None:
+            daemon_listen_maddr = self.run_coroutine(DHT._get_p2p_daemon_listen_maddr)
+            self._p2p_replica = await P2P.replicate(daemon_listen_maddr)
+        return self._p2p_replica
+
+    async def _get_p2p_daemon_listen_maddr(self, node: DHTNode) -> Multiaddr:
```
mryab, line 307: Also a static method

hivemind/optim/collaborative.py
```diff
@@ -354,7 +354,7 @@ def report_training_progress(self):
             with self.lock_local_progress:
                 current_time = get_dht_time()
                 local_state_info = TrainingState(
-                    endpoint=self.averager.endpoint,
+                    peer_id=self.averager.endpoint.to_base58(),
```
mryab, line 357: I'm slightly against casting to base58 in multiple places all over the code; would appreciate if it was possible to come up with a way to reduce this casting :)

Or maybe you can just call `__str__` everywhere, since `to_base58` is an implementation detail

hivemind/averaging/allreduce.py
```diff
@@ -224,9 +231,9 @@ async def _accumulate_parts_streaming(self, stream: AsyncIterator[averaging_pb2.
             yield averaging_pb2.AveragingData(code=averaging_pb2.AVERAGED_PART, tensor_part=serialized_delta)
 
     async def _send_error_to_peer(self, peer_endpoint: Endpoint, code: averaging_pb2.MessageCode):
-        stream = self._get_peer_stub(peer_endpoint).rpc_aggregate_part()
-        await stream.write(averaging_pb2.AveragingData(group_id=self.group_id, endpoint=self.endpoint, code=code))
-        await stream.done_writing()
+        error = averaging_pb2.AveragingData(group_id=self.group_id, endpoint=self.endpoint.to_base58(), code=code)
+        async for _ in self._get_peer_stub(peer_endpoint).rpc_aggregate_part(aiter(error)):
+            pass
```
mryab, line 236: Maybe a comment with an explanation of why we don't make an unary call here would be helpful

hivemind/p2p/p2p_daemon.py
```diff
@@ -472,9 +470,21 @@ async def add_binary_stream_handler(self, name: str, handler: p2pclient.StreamHa
         await self._client.stream_handler(name, handler)
 
     async def call_binary_stream_handler(
-        self, peer_id: PeerID, handler_name: str
+        self, peer_id: PeerID, handler_name: str, n_attempts: int = 5, delay: float = 0.5
```
mryab, line 473: A docstring that explains that we actually retry the call multiple times would be useful IMO

hivemind/p2p/servicer.py
```diff
@@ -41,39 +42,49 @@ class ServicerBase:
...
+            return
 
+        cls._rpc_handlers = []
+        for method_name, method in inspect.getmembers(cls, predicate=inspect.isfunction):
+            if method_name.startswith("rpc_"):
+                spec = inspect.getfullargspec(method)
+                if len(spec.args) < 3:
+                    raise ValueError(
+                        f"{method_name} is expected to at least three positional arguments "
+                        f"(self: TServicer, request: TInputProtobuf, context: hivemind.p2p.P2PContext)"
```
mryab, line 60: Thankfully, `TServicer` and `TInputProtobuf` are no more :)

mryab (changes requested):
Thanks for this monumental contribution! Before we merge, though, I'd also like to see two things:
1) Passing tests (maybe https://github.com/learning-at-home/hivemind/pull/337 and https://github.com/learning-at-home/hivemind/pull/334 will be useful in that regard)
2) Performance benchmarks comparing this one with the master branch

tests/test_p2p_servicer.py
```diff
@@ -25,24 +25,24 @@ async def rpc_square(self, request: test_pb2.TestRequest, _: P2PContext) -> test
...
+    stub = ExampleServicer.get_stub(client, server.id)
 
     assert await stub.rpc_square(test_pb2.TestRequest(number=10)) == test_pb2.TestResponse(number=100)
 
 
 @pytest.mark.asyncio
 async def test_stream_unary(server_client):
     class ExampleServicer(ServicerBase):
-        async def rpc_sum(self, request: AsyncIterator[test_pb2.TestRequest], _: P2PContext) -> test_pb2.TestResponse:
+        async def rpc_sum(self, numbers: AsyncIterator[test_pb2.TestRequest], _: P2PContext) -> test_pb2.TestResponse:
```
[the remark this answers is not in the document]
borzunov: I've renamed it the `numbers` parameter to `stream` (like in `test_stream_stream()`).

hivemind/averaging/averager.py
borzunov: I'd like to keep `client_mode` and `self.client_mode` equal in all cases (to avoid bugs when a developer accidentally used `client_mode` instead `self.client_mode` and got unexpected behavior). Therefore, one line (only the `self.client_mode` assignment) is not enough.

hivemind/p2p/servicer.py
borzunov: By default, RPC handlers are named as `ClassName.method_name`. This does not allow to use two averagers of the same class that work with different tensors and parameters, however @justheuristic requested this functionality for experiments.

Defining `namespace` changes the handler name to `namespace::ClassName.method_name`, therefore one can serve two sets of averager RPCs (for two averagers) at one endpoint.

hivemind/dht/__init__.py
borzunov: That's true, `self._node` exists in the child process but not in the parent one.

hivemind/optim/collaborative.py
borzunov: We can use `str(endpoint)`, however we would still need to call `PeerID.from_base58(value)` to deserialize. Therefore, I would keep these operations symmetric.

Also, it is actually more natural to use `bytes` for representing `PeerID`s in protobufs (and change `endpoint.to_base58()`/`PeerID.from_base58(value)` to `endpoint.to_bytes()`/`PeerID(value)`). However, DHT already uses `str` for `PeerIDs` in protobufs, so I'd like to make the code consistent in this PR (but I don't mind changing everything to `bytes` in a separate PR).

hivemind/p2p/servicer.py
borzunov: I've removed mentioning of `TServicer` from this comment. However, I'd still suggest to use the `T` prefix for `TypeVar`s to distinguish them from the usual types, so I am keeping `TInputProtobuf` for now :)

hivemind/averaging/allreduce.py
borzunov: We can't use a unary call here just because the function has a `stream -> stream` signature.

Nevertheless, I've made the code more readable with the new `asingle()` function and added a comment.

hivemind/p2p/p2p_daemon.py
borzunov: I've removed the retrying functionality here (initially, I thought it may help with passing tests, but it turned out to be not useful).

tests/test_averaging.py
```diff
@@ -278,17 +281,20 @@ def test_allgather():
...
     ]
 
     futures = []
     for i, averager in enumerate(averagers):
         futures.append(averager.step(wait=False, gather=dict(batch_size=123 + i, foo="bar")))
 
-    assert len(set(repr(sorted(future.result())) for future in futures)) == 2
+    gathered_data = [future.result() for future in futures]
+    gathered_data_reprs = [
+        repr(sorted({peer_id.to_base58(): data for peer_id, data in result.items()})) for result in gathered_data
```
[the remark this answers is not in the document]
borzunov: I've removed this check since the next check already ensures that this condition is true (but written in a more clear way).

tests/test_averaging.py
```diff
@@ -278,17 +281,20 @@ def test_allgather():
...
+    ]
+    assert len(set(gathered_data_reprs)) == 2
```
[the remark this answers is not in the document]
borzunov: I've removed this check since the next check already ensures that this condition is true (but written in a more clear way).

hivemind/utils/asyncio.py
```diff
@@ -59,6 +59,16 @@ async def aenumerate(aiterable: AsyncIterable[T]) -> AsyncIterable[Tuple[int, T]
         index += 1
 
 
+async def asingle(aiter: AsyncIterable[T]) -> T:
```
borzunov, line 62: The name is inspired by [Single()](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.single?view=net-5.0) from LINQ (.NET functional programming functions).

borzunov:
## Benchmark Results

### Setup

```python
num_peers = 16
target_group_size = 16
request_timeout = 1
hid_size = 8192
num_layers = 1
averaging_expiration = 300
```

### Branch `master` (39afa97e297beec406da0d2639a7c90077cb7933)

**Part size:** 2 ** 20
**Averaging time:** mean 23.3 sec (std 1.2 sec)

### Branch `averager-libp2p`



**Part size (optimal):** 2 ** 19
**Averaging time:** mean 25.8 sec (std 0.8 sec)

hivemind/optim/collaborative.py
mryab: Sure, let's change it in a follow-up

hivemind/averaging/allreduce.py
```diff
@@ -224,9 +231,9 @@ async def _accumulate_parts_streaming(self, stream: AsyncIterator[averaging_pb2.
             yield averaging_pb2.AveragingData(code=averaging_pb2.AVERAGED_PART, tensor_part=serialized_delta)
 
     async def _send_error_to_peer(self, peer_endpoint: Endpoint, code: averaging_pb2.MessageCode):
-        stream = self._get_peer_stub(peer_endpoint).rpc_aggregate_part()
-        await stream.write(averaging_pb2.AveragingData(group_id=self.group_id, endpoint=self.endpoint, code=code))
-        await stream.done_writing()
+        # In case of reporting the error, we expect the response stream to contain exactly one item
```
mryab, line 234: Maybe move to the next line, since the "expect" part is inside asingle

hivemind/averaging/averager.py
mryab: Still, seems slightly odd that in some cases this argument has a proper name and in some it doesn't

hivemind/averaging/matchmaking.py
mryab: up

hivemind/p2p/servicer.py
```diff
@@ -24,9 +24,10 @@ class StubBase:
     adding the necessary rpc_* methods. Calls to these methods are translated to calls to the remote peer.
     """
 
-    def __init__(self, p2p: P2P, peer: PeerID):
+    def __init__(self, p2p: P2P, peer: PeerID, namespace: Optional[str]):
```
mryab, line 27:
```suggestion
    def __init__(self, p2p: P2P, peer: PeerID, namespace: Optional[str]=None):
```

hivemind/utils/asyncio.py
```diff
@@ -59,6 +59,16 @@ async def aenumerate(aiterable: AsyncIterable[T]) -> AsyncIterable[Tuple[int, T]
...
+    """If ``aiter`` has exactly one item, returns this item. Otherwise, raises `ValueError`."""
+    count = 0
+    async for item in aiter:
+        count += 1
+    if count != 1:
+        raise ValueError(f"Iterable has {count} items instead of one (as expected)")
```
mryab, line 68: Maybe throw an error as soon as we encounter 2+ items?

mryab (approved):

hivemind/averaging/averager.py
borzunov: Agreed.

hivemind/averaging/matchmaking.py
borzunov: I decided to agree with @mryab.

hivemind/p2p/servicer.py
borzunov: I'd suggest to avoid setting default parameter values in functions that should not be directly called by user (this constructor is one of them because the stub should be instantiated via `Servicer.get_stub()`).

This is to (1) avoid forgetting to pass a non-default value in the library code and (2) avoid duplicating defaults.

.github/workflows/run-tests.yml
borzunov: We've reverted this change thanks to #334.

PR #325, opened 2021-07-16
Fix typing annotations

yhn112:


hivemind/averaging/partition.py
```diff
@@ -2,7 +2,7 @@
 Auxiliary data structures for AllReduceRunner
 """
 import asyncio
-from typing import Sequence, AsyncIterable, Tuple, Optional, TypeVar, Union, AsyncIterator
+from typing import Sequence, AsyncIterable, Tuple, Optional, TypeVar, Union, AsyncIterator, Type, Deque, List
```
borzunov, line 5: (optional) Consider sorting the imports, but this may be out of scope of this PR.

hivemind/utils/crypto.py
```diff
@@ -97,7 +97,7 @@ def to_bytes(self) -> bytes:
 
     @classmethod
     def from_bytes(cls, key: bytes) -> RSAPublicKey:
-        key = serialization.load_ssh_public_key(key)
-        if not isinstance(key, rsa.RSAPublicKey):
-            raise ValueError(f"Expected an RSA public key, got {key}")
-        return cls(key)
+        loaded_key = serialization.load_ssh_public_key(key)
+        if not isinstance(loaded_key, rsa.RSAPublicKey):
+            raise ValueError(f"Expected an RSA public key, got {str(key)}")
```
borzunov, line 102:
```suggestion
            raise ValueError(f"Expected an RSA public key, got {str(loaded_key)}")
```

borzunov (approved):

hivemind/dht/traverse.py
```diff
@@ -117,13 +118,12 @@ async def get_neighbors(peer, queries) -> {query1: ([nearest1, nearest2, ...], F
...
     """
     if len(queries) == 0:
-        return {}, dict(visited_nodes or {})
+        return {}, visited_nodes
 
     unfinished_queries = set(queries)  # all queries that haven't triggered finish_search yet
     candidate_nodes: Dict[DHTID, List[Tuple[int, DHTID]]] = {}  # heap: unvisited nodes, ordered nearest-to-farthest
     nearest_nodes: Dict[DHTID, List[Tuple[int, DHTID]]] = {}  # heap: top-k nearest nodes, farthest-to-nearest
     known_nodes: Dict[DHTID, Set[DHTID]] = {}  # all nodes ever added to the heap (for deduplication)
-    visited_nodes: Dict[DHTID, Set[DHTID]] = dict(visited_nodes or {})  # nodes that were chosen for get_neighbors call
```
mryab, line 126: I'm not sure that mutating the input object is the intended (or expected) behavior here, so it's best left as is

mryab (changes requested):

pyproject.toml
```diff
@@ -8,3 +8,25 @@ line_length = 119
 combine_as_imports = true
 combine_star = true
 known_local_folder = ["arguments", "test_utils", "tests", "utils"]
+
+
```
borzunov, line 12:
```suggestion
```

borzunov (approved):

borzunov:
In my opinion, the best strategy would be to achieve zero mypy errors on some module (e.g., hivemind.utils), then expand that to one more module, etc.

yhn112:
It is impossible to achieve 
> zero mypy errors

(in any module of the project) without

> making the code too verbose and actually making it less readable as a result.

borzunov:
Isn't there an alternative mypy config or an alternative type checker that doesn't require that?

PR #328, opened 2021-07-18
Optimize unary handlers with persistent connections to P2P daemon

deniskamazur:
This PR implements unary handlers over a persistent daemon connection.

The daemon part (Golang): https://github.com/learning-at-home/go-libp2p-daemon/pull/3


hivemind/p2p/p2p_daemon.py
```diff
@@ -402,16 +406,12 @@ async def add_protobuf_handler(
...
-                input = requests
-            else:
-                count = 0
-                async for input in requests:
-                    count += 1
-                if count != 1:
-                    raise ValueError(f"Got {count} requests for handler {name} instead of one")
+        if not stream_input:
+            await self._add_protobuf_unary_handler(name, handler, input_protobuf_type)
+            return
```
borzunov, line 411: This wouldn't work for unary -> stream handlers (we have them in the averager).

A way to fix this is to add `stream_output: bool` parameter to this function (we can set it in `Servicer`) and call the optimized unary handler implementation if `not stream_input and not stream_output`.

hivemind/p2p/p2p_daemon.py
```diff
@@ -169,6 +168,12 @@ async def create(
 
         return self
 
+    async def add_unary_handler(self, handle_name: str, handler: p2pclient.TUnaryHandler):
+        return await self._client.add_unary_handler(handle_name, handler)
+
+    async def unary_call(self, peer_id: PeerID, handle_name: str, data: bytes) -> bytes:
```
borzunov, line 174:
```suggestion
    async def call_unary_handler(self, peer_id: PeerID, handle_name: str, data: bytes) -> bytes:
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -422,15 +422,38 @@ async def _stream_handler(requests: P2P.TInputStream, context: P2PContext) -> P2
...
 
+    # only registers request-response handlers
+    async def _add_protobuf_unary_handler(
+        self,
+        handle_name: str,
+        handler: Callable[[TInputProtobuf, P2PContext], Awaitable[TOutputProtobuf]],
+        input_protobuf_type: type,
+    ) -> None:
+        async def _unary_handler(request: bytes) -> bytes:
+            input_serialized = input_protobuf_type().FromString(request)
```
borzunov, line 433:
```suggestion
            input_serialized = input_protobuf_type.FromString(request)
```

hivemind/proto/p2pd.proto
```diff
@@ -7,18 +7,23 @@ syntax = "proto2";
...
+    DHT                      = 4;
+    LIST_PEERS               = 5;
+    CONNMANAGER              = 6;
+    DISCONNECT               = 7;      
+    PUBSUB                   = 8;
+
+    PERSISTENT_CONN_UPGRADE  = 9;
+    CALL_UNARY              = 10;
+    ADD_UNARY_HANDLER       = 11;
+    SEND_RESPONSE_TO_REMOTE = 12;
```
borzunov, line 24:
```suggestion
    RETURN_UNARY_RESPONSE = 12;
```

This can be fixed after running the DHT benchmark :)

hivemind/proto/p2pd.proto
```diff
@@ -147,8 +159,9 @@ message PSRequest {
   optional bytes data = 3;
 }
 
+
```
borzunov, line 162:
```suggestion
```

Looks like the second empty line is added accidentally - let's remove it from diff :)

hivemind/proto/p2pd.proto
```diff
@@ -147,8 +159,9 @@ message PSRequest {
...
 message PSMessage {
-  optional bytes from_id = 1;
+  optional bytes from = 1;
```
borzunov, line 164: Looks like this name is not used - maybe keep it as is to minimize diff?

```suggestion
  optional bytes from_id = 1;
```

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -93,6 +103,99 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
         async with server:
             yield self
 
+    async def _read_from_persistent_conn(self, reader: asyncio.StreamReader):
+        while True:
+            resp: p2pd_pb.Response = p2pd_pb.Response()  # type: ignore
```
borzunov, line 108:
```suggestion
            resp = p2pd_pb.Response()  # type: ignore
```

nit: Seems like obvious type annotation.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -179,3 +282,9 @@ async def stream_handler(self, proto: str, handler_cb: StreamHandler) -> None:
 
         # if success, add the handler to the dict
         self.handlers[proto] = handler_cb
+
+
+class RemoteException(Exception):
```
borzunov, line 287: We already have `P2PHandlerError` for protobuf handlers, let's keep one of them.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -93,6 +103,99 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+                asyncio.create_task(self._handle_persistent_request(resp.requestHandling))
+                pass
+
+    async def _write_to_persistent_conn(self, writer: asyncio.StreamWriter):
+        while True:
+            msg = await self.pending_messages.get()
+            await write_pbmsg(writer, msg)
+
+    async def _handle_persistent_request(self, request):
+        assert request.proto in self.unary_handlers
```
borzunov, line 132: Together with @deniskamazur, we've agreed to change it to `logger.warning` (or similar)

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -65,17 +80,12 @@ def __init__(
...
-            # should never enter here... daemon should reject the stream for us.
-            writer.close()
-            raise DispatchFailure(e)
-        await handler(stream_info, reader, writer)
+        # persistent connection readers & writers
+        self._pers_conn_open: bool = False
+        self.unary_handlers: Dict[str, TUnaryHandler] = {}
+
+        self.pending_messages: asyncio.Queue[p2pd_pb.Request] = asyncio.Queue()
+        self.pending_calls: Dict[CallID, asyncio.Future] = {}
```
borzunov, line 88:
```suggestion
        self.pending_calls: Dict[CallID, asyncio.Future[bytes]] = {}
```

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -93,6 +103,99 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+            # should never enter here... daemon should reject the stream for us.
+            writer.close()
+            raise DispatchFailure(e)
+        await handler(stream_info, reader, writer)
+
+    async def _ensure_persistent_conn(self):
+        if not self._pers_conn_open:
+            reader, writer = await self.daemon_connector.open_persistent_connection()
+            asyncio.create_task(self._read_from_persistent_conn(reader))
+            asyncio.create_task(self._write_to_persistent_conn(writer))
```
borzunov, line 160: Let's save the tasks to the class fields and cancel them in the end of `listen()`.

Also, let's wrap the code in `_write_to_persistent_conn` with the `with closing(writer):` statement.

hivemind/proto/p2pd.proto
deniskamazur: Let's leave it consistent with the [current version](https://github.com/libp2p/go-libp2p-daemon/blob/bfa03bdcc0e24a9663575648cd7b52a01fe3f42f/pb/p2pd.proto#L147) of p2pd.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -93,6 +103,100 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+        while True:
+            msg = await self.pending_messages.get()
+            await write_pbmsg(writer, msg)
+
+    async def _handle_persistent_request(self, request):
+        assert request.proto in self.unary_handlers
+
+        try:
+            remote_id = PeerID(request.peer)
+            response_payload: bytes = self.unary_handlers[request.proto](request.data, remote_id)
```
borzunov, line 136:
```suggestion
            response_payload: bytes = await self.unary_handlers[request.proto](request.data, remote_id)
```

hivemind/proto/p2pd.proto
```diff
@@ -43,6 +52,9 @@ message Response {
   optional DHTResponse dht = 5;
   repeated PeerInfo peers = 6;
   optional PSResponse pubsub = 7;
+
+  optional CallUnaryResponse callUnaryResponse = 8;
+  optional CallUnaryRequest requestHandling = 9;
```
borzunov, line 57:
```suggestion
  optional CallUnaryResponse sendResponseToRemote = 8;
  optional CallUnaryRequest callUnary = 9;
```

Maybe name those bi-directional protobufs exactly as in `Request`?

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -93,6 +103,100 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+            proto=proto,
+            data=data,
+            callId=call_id.bytes,
+        )
+        req = p2pd_pb.Request(
+            type=p2pd_pb.Request.CALL_UNARY,
+            callUnary=call_unary_req,
+        )
+
+        await self._ensure_persistent_conn()
```
borzunov, line 191: There is a race condition if two `call_unary_handler`s are called simultaneously. I suggest to resolve this with `asyncio.Lock` and [the double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking) pattern.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -93,6 +103,100 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+        add_unary_handler_req = p2pd_pb.AddUnaryHandlerRequest(proto=proto)
+        req = p2pd_pb.Request(
+            type=p2pd_pb.Request.ADD_UNARY_HANDLER,
+            addUnaryHandler=add_unary_handler_req,
+        )
+        await self.pending_messages.put(req)
+
+        if self.unary_handlers.get(proto):
+            raise ValueError(f"Handler for protocol {proto} already assigned")
+        self.unary_handlers[proto] = handler
```
borzunov, line 176: Let's move these lines before `await self.pending_messages.put(req)` because:

  1. We want to raise ValueError before registering the duplicate handler in the daemon.
  2. In case if some badly written coroutine will take asyncio loop for a long time, other peers may start sending requests to the handler before `await self.pending_messages.put(req)` returns control to this function.

hivemind/p2p/p2p_daemon.py
```diff
@@ -422,13 +426,39 @@ async def _stream_handler(requests: P2P.TInputStream, context: P2PContext) -> P2
...
+        await self.add_unary_handler(handle_name, _unary_handler)
+
     async def call_protobuf_handler(
         self,
         peer_id: PeerID,
         name: str,
         input: Union[TInputProtobuf, TInputStream],
         output_protobuf_type: type,
+        *,
+        stream_output: bool = False,
```
borzunov, line 464:
```suggestion
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -422,13 +426,39 @@ async def _stream_handler(requests: P2P.TInputStream, context: P2PContext) -> P2
...
     ) -> Awaitable[TOutputProtobuf]:
+
+        if not (isinstance(input, AsyncIterableABC) or stream_output):
+            return await self._call_unary_protobuf_handler(peer_id, name, input, output_protobuf_type)
+
         requests = input if isinstance(input, AsyncIterableABC) else aiter(input)
         responses = self._iterate_protobuf_stream_handler(peer_id, name, requests, output_protobuf_type)
```
borzunov, line 471:
```suggestion
        responses = self._iterate_protobuf_stream_handler(peer_id, name, input, output_protobuf_type)
```

hivemind/proto/p2pd.proto
```diff
@@ -161,6 +191,28 @@ message PSResponse {
...
 }
 
+message CallUnaryRequest {
+  required bytes peer = 1;
+  required string proto = 2;
+  required bytes data = 3;
+}
+
+message CallUnaryResponse {
+  optional bytes result = 1;
```
borzunov, line 201: Can we use `oneof` here as well?

justheuristic:
Measured max # open files

TL;DR 32 peers, 8 concurrent jobs per peer, measure max n open files

before (5 runs): 1234, 1233, 1234, 1234, 1236
after (5 runs): 416, 448, 448, 448, 448

benchmark code:
```
import random
import asyncio
import hivemind

from tqdm.auto import tqdm

async def main(num_peers=32,
               num_tasks_per_peer=8,
               num_repeats=3):
    
    dht = [await hivemind.dht.DHTNode.create()]
    while len(dht) < num_peers:
        dht.append(
            await hivemind.dht.DHTNode.create(
                initial_peers=await random.choice(dht).get_visible_maddrs())
        )
        print(end='.', flush=True)
        
    async def do_work(node):
        for i in range(num_repeats):
            key = random.randint(0, 10 ** 9)
            value = random.randint(0, 10 ** 9)
            expiration_time = hivemind.get_dht_time() + random.random() * 10
            await node.store(key, value, expiration_time)
            await node.get(key, latest=True)
            print(end='.', flush=True)
        
    await asyncio.gather(*(do_work(dht[i % len(dht)]) for i in range(num_peers * num_tasks_per_peer)))
    
    print(hivemind.dht.protocol.MAX_CONCURRENT)
    
if __name__ == '__main__':
    asyncio.run(main())
```

Additional code changes: 
```
--- a/hivemind/dht/protocol.py
+++ b/hivemind/dht/protocol.py
@@ -21,6 +21,12 @@ from hivemind.utils.timed_storage import (
 logger = get_logger(__name__)
 
 
+import random
+
+CONCURRENT = 0
+MAX_CONCURRENT = 0
+
+
 class DHTProtocol(ServicerBase):
     # fmt:off
     p2p: P2P
@@ -99,12 +105,17 @@ class DHTProtocol(ServicerBase):
 
         :return: node's DHTID, if peer responded and decided to send his node_id
         """
+        global CONCURRENT, MAX_CONCURRENT
         try:
+            
             async with self.rpc_semaphore:
+                CONCURRENT += 1
+                MAX_CONCURRENT = max(CONCURRENT, MAX_CONCURRENT)
                 ping_request = dht_pb2.PingRequest(peer=self.node_info, validate=validate)
                 time_requested = get_dht_time()
                 response = await self.get_stub(peer).rpc_ping(ping_request, timeout=self.wait_timeout)
                 time_responded = get_dht_time()
+            CONCURRENT -= 1
         except Exception as e:
             logger.debug(f"DHTProtocol failed to ping {peer}", exc_info=True)
             response = None
@@ -138,6 +149,7 @@ class DHTProtocol(ServicerBase):
 
     async def rpc_ping(self, request: dht_pb2.PingRequest, context: P2PContext) -> dht_pb2.PingResponse:
         """Some node wants us to add it to our routing table."""
+        await asyncio.sleep(random.expovariate(1. / 0.1))
 
         response = dht_pb2.PingResponse(peer=self.node_info, dht_time=get_dht_time(), available=False)
 
@@ -183,6 +195,7 @@ class DHTProtocol(ServicerBase):
         :return: list of [True / False] True = stored, False = failed (found newer value or no response)
                  if peer did not respond (e.g. due to timeout or congestion), returns None
         """
+        global CONCURRENT, MAX_CONCURRENT
         if isinstance(expiration_time, DHTExpiration):
             expiration_time = [expiration_time] * len(keys)
         if subkeys is None:
@@ -215,7 +228,10 @@ class DHTProtocol(ServicerBase):
         )
         try:
             async with self.rpc_semaphore:
+                CONCURRENT += 1
+                MAX_CONCURRENT = max(CONCURRENT, MAX_CONCURRENT)
                 response = await self.get_stub(peer).rpc_store(store_request, timeout=self.wait_timeout)
+            CONCURRENT -= 1
             if response.peer and response.peer.node_id:
                 peer_id = DHTID.from_bytes(response.peer.node_id)
                 asyncio.create_task(self.update_routing_table(peer_id, peer, responded=True))
@@ -227,6 +243,7 @@ class DHTProtocol(ServicerBase):
 
     async def rpc_store(self, request: dht_pb2.StoreRequest, context: P2PContext) -> dht_pb2.StoreResponse:
         """Some node wants us to store this (key, value) pair"""
+        await asyncio.sleep(random.expovariate(1. / 0.1))
         if request.peer:  # if requested, add peer to the routing table
             asyncio.create_task(self.rpc_ping(dht_pb2.PingRequest(peer=request.peer), context))
         assert len(request.keys) == len(request.values) == len(request.expiration_time) == len(request.in_cache)
@@ -280,11 +297,15 @@ class DHTProtocol(ServicerBase):
          neighbors: a dictionary[node_id : peer_id] containing nearest neighbors from peer's routing table
          If peer didn't respond, returns None
         """
+        global CONCURRENT, MAX_CONCURRENT
         keys = list(keys)
         find_request = dht_pb2.FindRequest(keys=list(map(DHTID.to_bytes, keys)), peer=self.node_info)
         try:
             async with self.rpc_semaphore:
+                CONCURRENT += 1
+                MAX_CONCURRENT = max(CONCURRENT, MAX_CONCURRENT)
                 response = await self.get_stub(peer).rpc_find(find_request, timeout=self.wait_timeout)
+            CONCURRENT -= 1
             if response.peer and response.peer.node_id:
                 peer_id = DHTID.from_bytes(response.peer.node_id)
                 asyncio.create_task(self.update_routing_table(peer_id, peer, responded=True))
@@ -330,6 +351,7 @@ class DHTProtocol(ServicerBase):
         Someone wants to find keys in the DHT. For all keys that we have locally, return value and expiration
         Also return :bucket_size: nearest neighbors from our routing table for each key (whether or not we found value)
         """
+        await asyncio.sleep(random.expovariate(1. / 0.1))
         if request.peer:  # if requested, add peer to the routing table
             asyncio.create_task(self.rpc_ping(dht_pb2.PingRequest(peer=request.peer), context))
 

```

hivemind/proto/p2pd.proto
deniskamazur: Enums from 10 to 12 are actually deprecated -- I'll remove them later.

hivemind/proto/p2pd.proto
```diff
@@ -161,6 +188,30 @@ message PSResponse {
...
+message DaemonError {
   optional string message = 1;
 }
+
+message Cancel {
+}
+
+message RPCError {
+  optional string message = 1;
+}
```
borzunov, line 217: nit: \n

hivemind/proto/p2pd.proto
```diff
@@ -45,6 +47,31 @@ message Response {
   optional PSResponse pubsub = 7;
 }
 
+// Persistent connection request
+message PCRequest {
```
borzunov, line 51: The need for the comment here hints that we may need a more clear name.

```suggestion
message PersistentConnectionRequest {
```

hivemind/proto/p2pd.proto
```diff
@@ -45,6 +47,31 @@ message Response {
...
+  oneof message {
+    AddUnaryHandlerRequest addUnaryHandler = 2;
+    CallUnaryRequest  callUnary = 3;
+    CallUnaryResponse unaryResponse = 4;
+    Cancel cancel = 5;
+  }
+}
+
+// Persistent connection response
+message PCResponse {
```
borzunov, line 63: The need for the comment here hints that we may need a more clear name.

```suggestion
message PersistentConnectionResponse {
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -4,15 +4,17 @@
 from collections.abc import AsyncIterable as AsyncIterableABC
 from contextlib import closing, suppress
 from dataclasses import dataclass
+from google.protobuf.message import Message
```
borzunov, line 7: This import should be moved to the next group (third-party imports). Otherwise, `isort` won't accept it.

hivemind/p2p/p2p_daemon.py
```diff
@@ -169,6 +170,12 @@ async def create(
 
         return self
 
+    async def add_unary_handler(self, handle_name: str, handler: p2pclient.TUnaryHandler):
+        return await self._client.add_unary_handler(handle_name, handler)
+
+    async def call_unary_handler(self, peer_id: PeerID, handle_name: str, data: bytes) -> bytes:
+        return await self._client.call_unary_handler(peer_id, handle_name, data)
+
```
borzunov, line 178: I don't think we need these public bytes-level methods since it's unlikely that a user may want to call them directly (so they their presence will only confuse the user).

The private versions also seem useless since other private methods can call `self._client.add/call_unary_handler()` directly.

```suggestion
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -396,12 +402,18 @@ async def add_protobuf_handler(
...
         stream_input: bool = False,
+        stream_output: bool = False,
     ) -> None:
         """
         :param stream_input: If True, assume ``handler`` to take ``TInputStream``
                              (not just ``TInputProtobuf``) as input.
+        :param stream_output: If True, assume ``handler`` to return ``TOutputStream``
         """
 
+        if not (stream_input or stream_output):
```
borzunov, line 413: nit: This may be more clear:

```suggestion
        if not stream_input and not stream_output:
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -28,7 +30,6 @@ class P2PContext(object):
     handle_name: str
     local_id: PeerID
     remote_id: PeerID = None
-    remote_maddr: Multiaddr = None
```
borzunov, line 31: To future reviewers: I and @deniskamazur have agreed to remove this because

1) Handlers are supposed to work with peer IDs only (not with multiaddrs, which is a lower-level concept)
2) Absence of `remote_maddr` simplifies the unary handler implementation

hivemind/p2p/p2p_daemon.py
```diff
@@ -422,15 +434,38 @@ async def _stream_handler(requests: P2P.TInputStream, context: P2PContext) -> P2
 
         await self._add_protobuf_stream_handler(name, _stream_handler, input_protobuf_type)
 
+    # only registers request-response handlers
```
borzunov, line 437: nit: Move this to docstring and also explain that the request-response handlers are processed differently for the sake of optimization of performance/number of open files.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -65,17 +80,17 @@ def __init__(
...
-        await read_pbmsg_safe(reader, pb_stream_info)
-        stream_info = StreamInfo.from_protobuf(pb_stream_info)
-        try:
-            handler = self.handlers[stream_info.proto]
-        except KeyError as e:
-            # should never enter here... daemon should reject the stream for us.
-            writer.close()
-            raise DispatchFailure(e)
-        await handler(stream_info, reader, writer)
+        # persistent connection readers & writers
```
borzunov, line 83: nit: The comment seems unrelated.

```suggestion
```

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -65,17 +80,17 @@ def __init__(
...
+        self._pers_conn_open: bool = False
```
borzunov, line 84: nit: In my opinion, the `pers` part is unclear.

```suggestion
        self._persistent_conn_open: bool = False
```

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -65,17 +80,17 @@ def __init__(
...
+        self.unary_handlers: Dict[str, TUnaryHandler] = {}
+
+        self._ensure_conn_lock = asyncio.Lock()
+        self.pending_messages: asyncio.Queue[p2pd_pb.PCRequest] = asyncio.Queue()
+        self.pending_calls: Dict[CallID, asyncio.Future[bytes]] = {}
+        self.handler_tasks: Dict[CallID, asyncio.Task] = {}
```
borzunov, line 90: Please make `unary_handlers`, `pending_messages`, `pending_calls`, `handler_tasks` private if possible.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -90,8 +105,130 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+                resp = p2pd_pb.PCResponse()
+                await read_pbmsg_safe(reader, resp)
+
+                call_id = uuid.UUID(bytes=resp.callId)
+
+                if resp.HasField("callUnaryResponse"):
+                    if call_id in self.pending_calls and resp.callUnaryResponse.HasField("response"):
+                        self.pending_calls[call_id].set_result(resp.callUnaryResponse.response)
+                    elif call_id in self.pending_calls and resp.callUnaryResponse.HasField("error"):
+                        remote_exc = P2PHandlerError(resp.callUnaryResponse.error.decode())
```
borzunov, line 129:
```suggestion
                        remote_exc = P2PHandlerError(resp.callUnaryResponse.error.decode(errors='ignore'))
```

Otherwise, a remote party may crash the daemon by sending the error with invalid UTF-8 sequence.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -90,8 +105,130 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+                        self.pending_calls[call_id].set_exception(remote_exc)
+                    else:
+                        logger.debug(f"received unexpected unary call")
```
borzunov, line 132:
```suggestion
                        logger.debug("received unexpected unary call")
```

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -90,8 +105,130 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+
+    async def _write_to_persistent_conn(self, writer: asyncio.StreamWriter):
+        with closing(writer):
+            while True:
+                msg = await self.pending_messages.get()
+                await write_pbmsg(writer, msg)
+
+    async def _handle_persistent_request(self, call_id: uuid.UUID, request: p2pd_pb.CallUnaryRequest):
+        if request.proto not in self.unary_handlers:
+            logger.warning(f"Protocol {request.proto} not supported")
```
borzunov, line 149:
```suggestion
            logger.warning(f"Protocol {request.proto} not supported")
            return
```

borzunov (changes requested):
Thanks for the contribution :)

I've left several comments to be addressed. Also, please rebase the code and make sure style checks pass.

hivemind/p2p/p2p_daemon.py
```diff
@@ -375,15 +375,21 @@ async def add_protobuf_handler(
...
-        input_protobuf_type: type,
+        input_protobuf_type: Message,
         *,
         stream_input: bool = False,
+        stream_output: bool = False,
     ) -> None:
         """
         :param stream_input: If True, assume ``handler`` to take ``TInputStream``
                              (not just ``TInputProtobuf``) as input.
+        :param stream_output: If True, assume ``handler`` to return ``TOutputStream``
```
borzunov, line 386:
```suggestion
        :param stream_output: If True, assume ``handler`` to return ``TOutputStream``
                              (not ``Awaitable[TOutputProtobuf]``).
```

hivemind/p2p/p2p_daemon.py
borzunov: Let's remove the comment since we now have the docstring :)

```suggestion
```

hivemind/p2p/p2p_daemon.py
```diff
@@ -396,23 +402,67 @@ async def _stream_handler(requests: P2P.TInputStream, context: P2PContext) -> P2
...
+    async def _call_unary_protobuf_handler(
+        self,
+        peer_id: PeerID,
+        handle_name: str,
+        input: TInputProtobuf,
+        output_protobuf_type: Message,
+    ) -> Awaitable[TOutputProtobuf]:
+        serialized_input = input.SerializeToString()
+        response = await self._client.call_unary_handler(peer_id, handle_name, serialized_input)
+        return output_protobuf_type().FromString(response)
```
borzunov, line 458:
```suggestion
        return output_protobuf_type.FromString(response)
```

We usually call static methods using the type itself (not an instance, creation of which is redundant).

tests/test_p2p_daemon.py
```diff
@@ -36,7 +36,7 @@ async def test_daemon_killed_on_del():
 
 @pytest.mark.asyncio
 async def test_startup_error_message():
-    with pytest.raises(P2PDaemonError, match=r"Failed to connect to bootstrap peers"):
+    with pytest.raises(P2PDaemonError, match=r"failed to connect to bootstrap peers"):
```
borzunov, line 39:
```suggestion
    with pytest.raises(P2PDaemonError, match=r"(?i)Failed to connect to bootstrap peers"):
```

`(?i)` will make the regex case-insensitive (it seems like we have both upper- and lowercase in different versions of the daemon).

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -90,8 +104,135 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+
+        await self._ensure_persistent_conn()
+
+        try:
+            self._pending_calls[call_id] = asyncio.Future()
+            await self._pending_messages.put(req)
+            return await self._pending_calls[call_id]
+
+        except asyncio.CancelledError:
+            asyncio.create_task(self._send_call_cancel(call_id))
```
borzunov, line 231:
```suggestion
            await self._send_call_cancel(call_id)
```

I suggest to stick with `await` unless it creates more warnings.

The reason is that if `self._send_call_cancel` fails, the error won't be hidden and will be propagated to the caller.

I remember that I suggested to try `asyncio.create_task()` earlier, but that was an attempt to deal with the asyncio warnings (that were fixed in #346).

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -90,8 +104,135 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+
+        await self._pending_messages.put(
+            p2pd_pb.PersistentConnectionRequest(
+                callId=call_id.bytes,
+                unaryResponse=response,
+            )
+        )
+        self._handler_tasks.pop(call_id)
+
+    async def _handler(self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter):
```
borzunov, line 166: For future reviewers: this coroutine is not new, just moved from another place.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -90,8 +104,135 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+        stream_info = StreamInfo.from_protobuf(pb_stream_info)
+        try:
+            handler = self.handlers[stream_info.proto]
+        except KeyError as e:
+            # should never enter here... daemon should reject the stream for us.
+            writer.close()
+            raise DispatchFailure(e)
+        await handler(stream_info, reader, writer)
+
+    async def _send_call_cancel(self, call_id: uuid.UUID):
```
borzunov, line 178:
```suggestion
    async def _cancel_unary_call(self, call_id: uuid.UUID):
```

Let's highlight that this concerns _unary_ calls :)

hivemind/p2p/p2p_daemon_bindings/control.py
borzunov, line 166: Can we move this coroutine back to the original place? The reasons are:

1. One may think that it is related to persistent connections and unary calls because it is located in the middle of the unary call machinery.
2. Moving it back to the original place will shorten the diff.

borzunov (changes requested):
Thanks for the changes! I've left one more bunch of change requests.

borzunov (approved):

hivemind/p2p/p2p_daemon_bindings/p2pclient.py
```diff
@@ -10,16 +10,24 @@
...
+    def __init__(self) -> None:
+        self.control = None
+
+    @classmethod
+    async def create(cls, control_maddr: Multiaddr = None, listen_maddr: Multiaddr = None) -> "Client":
+        client = cls()
+
         daemon_connector = DaemonConnector(control_maddr=control_maddr)
-        self.control = ControlClient(daemon_connector=daemon_connector, listen_maddr=listen_maddr)
+        client.control = await ControlClient.create(daemon_connector=daemon_connector, listen_maddr=listen_maddr)
```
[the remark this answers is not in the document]
deniskamazur: The daemon termination feature. The ControlClient can now open a persistent connection to the daemon on startup.

hivemind/p2p/p2p_daemon.py
```diff
@@ -258,7 +263,7 @@ async def send_protobuf(protobuf: Union[TOutputProtobuf, RPCError], writer: asyn
 
     @staticmethod
     async def receive_protobuf(
-        input_protobuf_type: type, reader: asyncio.StreamReader
+        input_protobuf_type: Message, reader: asyncio.StreamReader
```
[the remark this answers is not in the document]
deniskamazur: image
Yes!

tests/test_p2p_daemon_bindings.py
```diff
@@ -193,7 +193,8 @@ def test_parse_conn_protocol_invalid(maddr_str):
 
 
 @pytest.mark.parametrize("control_maddr_str", ("/unix/123", "/ip4/127.0.0.1/tcp/6666"))
-def test_client_ctor_control_maddr(control_maddr_str):
+@pytest.mark.asyncio
+async def test_client_ctor_control_maddr(control_maddr_str):
```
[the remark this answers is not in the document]
deniskamazur: I'm not sure, but it might just mean "Constructor".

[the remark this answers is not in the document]
deniskamazur: Should I rename the test to `test_client_create_control_maddr`?

hivemind/p2p/p2p_daemon.py
```diff
@@ -396,23 +407,66 @@ async def _stream_handler(requests: P2P.TInputStream, context: P2PContext) -> P2
 
         await self._add_protobuf_stream_handler(name, _stream_handler, input_protobuf_type)
 
+    async def _add_protobuf_unary_handler(
+        self,
+        handle_name: str,
```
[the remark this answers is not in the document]
deniskamazur: Currently, we both use `handle_name` and `handler_name` in the code base. Maybe, we should rather clean it up in a followup pull request.

hivemind/p2p/p2p_daemon.py
```diff
@@ -396,23 +407,66 @@ async def _stream_handler(requests: P2P.TInputStream, context: P2PContext) -> P2
...
         name: str,
         input: Union[TInputProtobuf, TInputStream],
-        output_protobuf_type: type,
+        output_protobuf_type: Message,
     ) -> Awaitable[TOutputProtobuf]:
+
+        if not isinstance(input, AsyncIterableABC):
+            return await self._call_unary_protobuf_handler(peer_id, name, input, output_protobuf_type)
+
         requests = input if isinstance(input, AsyncIterableABC) else aiter(input)
```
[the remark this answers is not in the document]
deniskamazur: Fixed it.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -90,8 +135,116 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+            call_id = uuid.UUID(bytes=resp.callId)
+
+            if resp.HasField("callUnaryResponse"):
+                if call_id in self._pending_calls and resp.callUnaryResponse.HasField("response"):
+                    self._pending_calls[call_id].set_result(resp.callUnaryResponse.response)
+                elif call_id in self._pending_calls and resp.callUnaryResponse.HasField("error"):
+                    remote_exc = P2PHandlerError(resp.callUnaryResponse.error.decode(errors="ignore"))
+                    self._pending_calls[call_id].set_exception(remote_exc)
+                else:
+                    logger.debug("received unexpected unary call")
```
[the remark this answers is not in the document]
deniskamazur: Added the unary call protobuf to the log message.

tests/test_p2p_daemon_bindings.py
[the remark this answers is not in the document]
deniskamazur: Renamed these test functions.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -54,17 +55,61 @@ async def open_connection(self) -> (asyncio.StreamReader, asyncio.StreamWriter):
...
+        """
+        reader, writer = await self.open_connection()
+        req = p2pd_pb.Request(type=p2pd_pb.Request.PERSISTENT_CONN_UPGRADE)
+        await write_pbmsg(writer, req)
+
+        return reader, writer
+
+
+TUnaryHandler = Callable[[bytes, PeerID], Awaitable[bytes]]
+CallID = uuid.UUID
```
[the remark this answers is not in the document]
deniskamazur: Done.

hivemind/p2p/p2p_daemon.py
```diff
@@ -375,15 +379,22 @@ async def add_protobuf_handler(
...
     ) -> None:
         """
         :param stream_input: If True, assume ``handler`` to take ``TInputStream``
                              (not just ``TInputProtobuf``) as input.
+        :param stream_output: If True, assume ``handler`` to return ``TOutputStream``
+                              (not ``Awaitable[TOutputProtobuf]``).
         """
 
+        if not stream_input and not stream_output:
+            await self._add_protobuf_unary_handler(name, handler, input_protobuf_type)
```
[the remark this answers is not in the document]
deniskamazur: I'm afraid this might lead to copy-pasting some code. Should we should vote on this.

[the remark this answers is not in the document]
borzunov: @mryab Since this PR provides an optimization for unary-unary handlers, I think it's reasonable to enable it when possible instead of having two separate APIs for unary handlers (non-optimized handlers added with `add_protobuf_handler()` and optimized handlers added with `add_protobuf_unary_handler()`).

Also, please note that currently this function are not called directly by the user (though it's possible in future) but called inside the `Servicer` (that just sets `stream_input`/`stream_output` flags depending on whether the request/response types are `AsyncIterator` or not).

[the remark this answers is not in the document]
mryab: Sounds reasonable, let's keep it as is then. Thanks for an explanation!

hivemind/p2p/p2p_daemon.py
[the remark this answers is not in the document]
mryab: Yup, if possible, let's make a note of it somewhere (as an issue or as a TODO in the top message of this PR)

hivemind/p2p/p2p_daemon.py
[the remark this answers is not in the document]
mryab: Renamed to Type[Message] to avoid passing instances

hivemind/p2p/p2p_daemon_bindings/p2pclient.py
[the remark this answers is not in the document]
mryab: I thought of some ways of avoiding it (e.g. running just one coroutine that opens a persistent connection synchronously), but let's keep it as is for simplicity

mryab (approved):

borzunov:
@deniskamazur Please remove the file `connection_handler_p2p.py` from this PR before merging :)

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -90,8 +135,116 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+        finally:
+            if self._read_task is not None:
+                self._read_task.cancel()
+            if self._write_task is not None:
+                self._write_task.cancel()
+
+    async def _read_from_persistent_conn(self, reader: asyncio.StreamReader):
+        while True:
+            resp = p2pd_pb.PersistentConnectionResponse()
+            await read_pbmsg_safe(reader, resp)
```
borzunov, line 150:
```suggestion
            try:
                await read_pbmsg_safe(reader, resp)
            except asyncio.IncompleteReadError:
                break
```

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -90,8 +135,119 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+                handler_task = asyncio.create_task(self._handle_persistent_request(call_id, resp.requestHandling))
+                self._handler_tasks[call_id] = handler_task
+
+            elif call_id in self._handler_tasks and resp.HasField("cancel"):
+                self._handler_tasks[call_id].cancel()
+
+    async def _write_to_persistent_conn(self, writer: asyncio.StreamWriter):
+        with closing(writer):
+            while True:
+                msg = await self._pending_messages.get()
```
borzunov, line 176:
```suggestion
                try:
                    msg = await self._pending_messages.get()
                except ValueError:
                    # Ignore "ValueError: I/O operation on closed file" because it is normal behavior
                    # if the connection is closed by the other party
                    break
```

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -54,17 +55,61 @@ async def open_connection(self) -> (asyncio.StreamReader, asyncio.StreamWriter):
...
+        *,
+        _initialized_with_create=False,
     ) -> None:
+        assert _initialized_with_create, "Please use ControlClient.create coroutine to spawn new control instances"
+
         self.listen_maddr = listen_maddr
         self.daemon_connector = daemon_connector
         self.handlers: Dict[str, StreamHandler] = {}
 
+        self._is_persistent_conn_open: bool = False
```
borzunov, line 89:
```suggestion
```

This flag is not used.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -90,8 +135,124 @@ async def listen(self) -> AsyncIterator["ControlClient"]:
...
+            ),
+        )
+
+    async def _ensure_persistent_conn(self):
+        reader, writer = await self.daemon_connector.open_persistent_connection()
+
+        self._read_task = asyncio.create_task(self._read_from_persistent_conn(reader))
+        self._write_task = asyncio.create_task(self._write_to_persistent_conn(writer))
+
+        self._is_persistent_conn_open = True
```
borzunov, line 219:
```suggestion
```

This flag is not used.

hivemind/p2p/p2p_daemon_bindings/control.py
borzunov: This turned out to be a wrong fix.

hivemind/p2p/p2p_daemon.py
```diff
@@ -90,6 +92,7 @@ async def create(
         use_auto_relay: bool = False,
         relay_hop_limit: int = 0,
         startup_timeout: float = 15,
+        idle_timeout: float = 0,
```
borzunov, line 95:
```suggestion
        idle_timeout: float = 30,
```

Let's enable this by default.

borzunov (changes requested):
I'd like to ensure that tests are stable before merging :)

hivemind/dht/__init__.py
```diff
@@ -94,6 +94,9 @@ def __init__(
         self._client_mode = None
         self._p2p_replica = None
 
+        if "p2p" in self.kwargs:
+            self._daemon_listen_maddr = self.kwargs.pop("p2p").daemon_listen_maddr
```
borzunov, line 98: Let's add `p2p: P2P` to the `__init__` args and do it this way:

```suggestion
        self._daemon_listen_maddr = p2p.daemon_listen_maddr if p2p is not None else None
```

The difference is that (1) we won't keep the extra reference to the original P2P in `self.kwargs` (so the caller may destroy it if it wants to) and (2) we will have explicit argument instead of accessing kwargs (so we will still have type and name checking).

hivemind/dht/__init__.py
```diff
@@ -105,10 +108,16 @@ def run(self) -> None:
 
             async def _run():
                 try:
+                    if hasattr(self, "_daemon_listen_maddr"):
+                        replicated_p2p = await P2P.replicate(self._daemon_listen_maddr)
+                    else:
+                        replicated_p2p = None
```
borzunov, line 114:
```suggestion
                    if self._daemon_listen_maddr is not None:
                        replicated_p2p = await P2P.replicate(self._daemon_listen_maddr)
                    else:
                        replicated_p2p = None
```

It's better to avoid using `hasattr()` if possible (so no field names are a part of string constants, thus renames/type checking work well).

justheuristic (changes requested):
* [x] in DaemonConnector create_persistent_connection, await an ok message from daemon to ensure that it was created successfully before returning connection
* [ ] add a test that attempts to call: (1) non-existing unary handler (2) existing handler with wrong protobuf types (3) self-dial [non-]existing  handler. Ensure that it raises error instead of hanging
* [ ] p2pd: do **not** send ok (p2pd->pyclient) on receiving unary response
* [ ] update p2pd versions (merge libp2p master into ours) - https://github.com/libp2p/go-libp2p-daemon/commit/193eab6a96ca727058caba58913264de99e73681
* [ ] print warning on any unexpected messages over persistent connection

hivemind/p2p/servicer.py
```diff
@@ -131,6 +131,7 @@ async def add_p2p_handlers(self, p2p: P2P, wrapper: Any = None, *, namespace: Op
                 getattr(servicer, handler.method_name),
                 handler.request_type,
                 stream_input=handler.stream_input,
```
justheuristic, line 133: Nitpicking: should we create them asynchronously via gather

justheuristic: upd: this is definitely not a bottleneck, feel free to ignore/resolve

borzunov: I support this because it may improve startup time. Good catch!

deniskamazur:
Just in case, the `build_and_test_p2pd` job is expected to fail, as the release, used in `setup.py` points to the master branch of p2pd.

hivemind/p2p/servicer.py
deniskamazur: Done!

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -131,6 +131,9 @@ def from_protobuf(cls, peer_info_pb: p2pd_pb2.PeerInfo) -> "PeerInfo":
     def __str__(self):
         return f"{self.peer_id.pretty()} {','.join(str(a) for a in self.addrs)}"
 
+    def __repr__(self):
+        return str(self)
```
borzunov, line 135:
```suggestion
        return f"PeerInfo(peer_id={repr(self.peer_id)}, addrs={repr(self.addrs)})"
```

`repr()` usually shows the class name and contents in a Python-like notation.

borzunov (approved):
Approving this. We have also decided to revert the first change from https://github.com/learning-at-home/go-libp2p-daemon/pull/11 in a future PR (it is unnecessary but increases startup time).

justheuristic (approved):

PR #331, opened 2021-07-21
Fix deadlocks in DecentralizedAverager and MPFuture

borzunov:
This PR does the following:

1. Fix a bug in `DecentralizedAverager.rpc_join_group()`.

    Previously, if a peer didn't manage to update its group key in the DHT in time (or didn't manage to propagate the updates), another peer could have chosen it as a leader based on the group key from the previous stage. The groups have been assembling incorrectly, which led to occasional deadlocks in tests (where the time periods between sequential averager steps are low).

    The fix forces leaders to compare followers' group keys before allowing them to join.

2. Fix a bug with MPFuture state after killing child processes.

    Many tests don't perform graceful shutdowns, so we are using a fixture that kills leftover child processes (if they don't respond to `terminate()` in 1 sec). Such killings may leave the global MPFuture state broken (e.g. the global locks acquired and the background thread blocked).

    The fix resets MPFuture backend (including its global state) after killing the child processes between tests.

3. Add `-v` flag to pytest in CI.

    Without this flag, we had no opportunity to determine a particular test that caused a deadlock since pytest used one output line per file (and Github CI erases the last output line when timeout was exceeded).

    The fix makes pytest to use one output line per test.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Test
         run: |
           cd tests
-          pytest --durations=0 --durations-min=1.0
+          pytest --durations=0 --durations-min=1.0 -v
 
   build_and_test_p2pd:
     runs-on: ubuntu-latest
@@ -60,7 +60,7 @@ jobs:
       - name: Test
         run: |
           cd tests
-          pytest -k "p2p" 
+          pytest -k "p2p"
 
   codecov_in_develop_mode:
 
@@ -87,6 +87,6 @@ jobs:
           pip install -e .
       - name: Test
         run: |
-          pytest --cov=hivemind tests
+          pytest --cov=hivemind -v tests
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -60,7 +60,7 @@ jobs:
       - name: Test
         run: |
           cd tests
-          pytest -k "p2p"
+          pytest -k "p2p" -v
 
   codecov_in_develop_mode:
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -28,5 +28,5 @@ def cleanup_children():
             with suppress(psutil.NoSuchProcess):
                 child.kill()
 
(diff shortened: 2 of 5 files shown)


justheuristic (approved):

.github/workflows/run-tests.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       - name: Test
         run: |
           cd tests
-          pytest -k "p2p" 
+          pytest -k "p2p"
```
mryab, line 63: Maybe add -v here as well for consistency

tests/conftest.py
```diff
@@ -26,3 +27,6 @@ def cleanup_children():
         for child in children:
             with suppress(psutil.NoSuchProcess):
                 child.kill()
+
+    # Killing child processes may leave the MPFuture backend broken
+    MPFuture.reset_backend()
```
mryab, line 32: Perhaps it's only necessary to do this when we had to terminate some child processes by force?
Solution 2: it might be preferable to reser the global state between the tests regardless of whether we actually kill processes. In that case, an explicit cleanup might still be useful, but a comment can be more general

tests/conftest.py
```diff
@@ -5,6 +5,7 @@
 import pytest
 
 from hivemind.utils import get_logger
+from hivemind.utils.mpfuture import MPFuture
```
mryab, line 8: Import from utils as well?

mryab (approved):

tests/conftest.py
justheuristic: I weakly support option 2, rationale: if some fail in such a way that backend is corrupted, resetting backend ensures that no other tests will be affected, hence we will know the main culprit right away

yhn112 (approved):

tests/conftest.py
borzunov: In general, I'd argue for avoiding importing from `hivemind.utils` directly to avoid having one module with a lot of mixed-purpose stuff and the corresponding long import line.

To follow this, I've changed `from hivemind.utils import get_logger` above to `from hivemind.utils.logging import get_logger`.

tests/conftest.py
borzunov: I've implemented the 2nd option (updated the comment).

PR #336, opened 2021-07-23
Resolve deadlock for MPFuture [duplex version]

justheuristic:


hivemind/utils/mpfuture.py
```diff
@@ -62,8 +63,10 @@ class MPFuture(base.Future, Generic[ResultType]):
     """
 
     _initialization_lock = mp.Lock()  # global lock that prevents simultaneous initialization of two processes
-    _update_lock = mp.Lock()  # global lock that prevents simultaneous writing to the same pipe
-    _process_wide_pipe: Optional[PipeEnd] = None  # a pipe that is used to send results/exceptions to this process
+    _update_lock = mp.Lock()  # global lock that prevents simultaneous writing of results/exceptions through same pipe
+    _status_lock = mp.Lock()  # global lock that prevents simultaneous sening of status updates through same pipe
```
borzunov, line 67:
```suggestion
    _status_lock = mp.Lock()  # global lock that prevents simultaneous sending of status updates through same pipe
```

hivemind/utils/mpfuture.py
```diff
@@ -140,67 +142,72 @@ def reset_backend(cls):
...
+            selector.register(cls._process_inner_pipe, EVENT_READ, data=cls._process_inner_pipe)
+            selector.register(cls._process_outer_pipe, EVENT_READ, data=cls._process_outer_pipe)
+
+            while True:
+                try:
+                    if cls._pipe_waiter_thread is not threading.current_thread():
+                        break  # Backend was reset, a new background thread has started
+
+                    pipe = next((key.data for (key, events) in selector.select()))
+                    uid, msg_type, payload = pipe.recv()
```
borzunov, line 158:
```suggestion
                    (key, events), *_ = selector.select()
                    uid, msg_type, payload = key.fileobj.recv()
```

hivemind/utils/mpfuture.py
```diff
@@ -140,67 +142,72 @@ def reset_backend(cls):
...
-                elif msg_type == MessageType.STATE_RESPONSE:
-                    future, state_updated_event = cls._status_requests.get(uid, (None, None))
-                    if future is None:
-                        logger.debug("Received a state update for a future that does not await status update.")
-                    else:
-                        if payload is not None:
-                            future.__setstate__(payload)
+        with DefaultSelector() as selector:
+            selector.register(cls._process_inner_pipe, EVENT_READ, data=cls._process_inner_pipe)
+            selector.register(cls._process_outer_pipe, EVENT_READ, data=cls._process_outer_pipe)
```
borzunov, line 150:
```suggestion
            selector.register(cls._process_inner_pipe, EVENT_READ)
            selector.register(cls._process_outer_pipe, EVENT_READ)
```

This is already stored in `SelectorKey.fileobj`.

borzunov (approved):
Please consider this approve valid only if the 10x test run does not freeze.

borzunov (changes requested):
We have found a case where `status_lock` still causes a deadlock and moved on to an alternative approach in #337.

PR #337, opened 2021-07-24
Resolve deadlock in MPFuture

justheuristic:
This PR partially rolls back MPFuture to an earlier state that still uses posix shared memory, with some additional safeguards that protect it against hanging on fork / termination.

Note: we have also tried queue-based MPFuture, but we couldn't figure out how to make it stable. This implementation corresponds to 7bd26fa

### DIFF
--- a/hivemind/utils/mpfuture.py
+++ b/hivemind/utils/mpfuture.py
@@ -43,6 +43,7 @@ class SharedBytes:
     The chunks are deallocated by the garbage collector,
     when it detects that all processes no longer use any bytes from this chunk.
     """
+
     _lock = mp.Lock()
     _pid: Optional[PID] = None
     _buffer: Optional[torch.Tensor] = None
@@ -89,7 +90,7 @@ class MPFuture(base.Future, Generic[ResultType]):
     _update_lock = mp.Lock()  # global lock that prevents simultaneous writing to the same pipe
     _global_sender_pipe: Optional[PipeEnd] = None  # a pipe that is used to send results/exceptions to this process
     _pipe_waiter_thread: Optional[threading.Thread] = None  # process-specific thread that receives results/exceptions
-    _active_futures: Optional[Dict[UID, 'ref[MPFuture]']] = None  # non-done futures originated from this process
+    _active_futures: Optional[Dict[UID, "ref[MPFuture]"]] = None  # non-done futures originated from this process
     _active_pid: Optional[PID] = None  # pid of currently active process; used to handle forks natively
 
     def __init__(self, *, use_lock: bool = True):
--- a/hivemind/utils/mpfuture.py
+++ b/hivemind/utils/mpfuture.py
@@ -36,22 +36,26 @@ except ImportError:
 
 
 class SharedBytes:
-    lock = mp.Lock()
-    pid: Optional[PID] = None
-    buffer: Optional[torch.Tensor] = None
-    index: Optional[int] = 0
+    """
+    A process-wide object that allocates large chunks of shared memory and partitions it into individual bytes.
+    Note: this process is only responsible for bulk allocation, it does not manage/remove unused bytes.
+    """
+    _lock = mp.Lock()
+    _pid: Optional[PID] = None
+    _buffer: Optional[torch.Tensor] = None
+    _index: Optional[int] = 0
 
     @classmethod
     def next(cls):
-        with cls.lock:
-            if cls.pid != os.getpid() or cls.buffer is None or cls.index >= len(cls.buffer):
+        with cls._lock:
+            if cls._pid != os.getpid() or cls._buffer is None or cls._index >= len(cls._buffer):
                 buffer_size = os.environ.get("HIVEMIND_SHM_BUFFER_SIZE", 4096)
-                cls.pid = os.getpid()
-                cls.buffer = torch.empty([buffer_size], dtype=torch.uint8).share_memory_()
-                cls.index = 0
+                cls._pid = os.getpid()
+                cls._buffer = torch.empty([buffer_size], dtype=torch.uint8).share_memory_()
+                cls._index = 0
 
-            cls.index += 1
-            return cls.buffer[cls.index - 1]
+            cls._index += 1
+            return cls._buffer[cls._index - 1]
 
 
 class UpdateType(Enum):
--- a/hivemind/utils/mpfuture.py
+++ b/hivemind/utils/mpfuture.py
@@ -157,6 +157,14 @@ class MPFuture(base.Future, Generic[ResultType]):
         )
         cls._pipe_waiter_thread.start()
 
+    @staticmethod
+    def reset_backend():
+        """Last-resort function to reset internals of MPFuture. All current MPFuture instances will be broken"""
+        MPFuture._active_pid = None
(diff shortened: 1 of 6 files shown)


borzunov (approved):

mryab (approved):

hivemind/utils/mpfuture.py
```diff
@@ -32,13 +35,29 @@ class InvalidStateError(Exception):
...
 
 
-class MessageType(Enum):
+class SharedBytes:
+    lock = mp.Lock()
+    pid: Optional[PID] = None
+    buffer: Optional[torch.Tensor] = None
+    index: Optional[int] = 0
+
+    @classmethod
```
yhn112, line 44: Please add docstrings

hivemind/utils/mpfuture.py
```diff
@@ -32,13 +35,33 @@ class InvalidStateError(Exception):
         """Raised when attempting to change state of a future in a terminal state (e.g. finished)"""
 
 
-class MessageType(Enum):
+class SharedBytes:
+    """
+    A process-wide object that allocates large chunks of shared memory and partitions it into individual bytes.
+    Note: this process is only responsible for bulk allocation, it does not manage/remove unused bytes.
```
borzunov, line 41:
```suggestion

    Note: this process is only responsible for bulk allocation, it does not manage/remove unused bytes.
    The chunks are deallocated by the garbage collector,
    when it detects that all processes no longer use any bytes from this chunk.
```

hivemind/utils/mpfuture.py
```diff
@@ -32,13 +35,33 @@ class InvalidStateError(Exception):
...
+    """
+    _lock = mp.Lock()
+    _pid: Optional[PID] = None
+    _buffer: Optional[torch.Tensor] = None
+    _index: Optional[int] = 0
```
borzunov, line 46:
```suggestion
    _index: int = 0
```

hivemind/utils/mpfuture.py
```diff
@@ -63,36 +83,49 @@ class MPFuture(base.Future, Generic[ResultType]):
 
     _initialization_lock = mp.Lock()  # global lock that prevents simultaneous initialization of two processes
     _update_lock = mp.Lock()  # global lock that prevents simultaneous writing to the same pipe
-    _process_wide_pipe: Optional[PipeEnd] = None  # a pipe that is used to send results/exceptions to this process
+    _global_sender_pipe: Optional[PipeEnd] = None  # a pipe that is used to send results/exceptions to this process
     _pipe_waiter_thread: Optional[threading.Thread] = None  # process-specific thread that receives results/exceptions
     _active_futures: Optional[Dict[UID, Type[ref][MPFuture]]] = None  # non-done futures originated from this process
```
borzunov, line 88:
```suggestion
    _active_futures: Optional[Dict[UID, 'ref[MPFuture]']] = None  # non-done futures originated from this process
```

`ref[MPFuture]` is made syntactically correct in Python 3.9 (as well as `asyncio.Future[int]`) but we need quotes for earlier Python versions.

hivemind/utils/mpfuture.py
```diff
@@ -63,36 +83,49 @@ class MPFuture(base.Future, Generic[ResultType]):
...
-    _status_requests: Optional[Dict[UID, Tuple[MPFuture, threading.Event]]] = None  # futures to be updated by origin
     _active_pid: Optional[PID] = None  # pid of currently active process; used to handle forks natively
 
-    SOFT_UPDATE_TIMEOUT = 0.5  # seconds spent awaiting status update before warning is printed
-    HARD_UPDATE_TIMEOUT = 10.0  # seconds spent awaiting status update before future is automatically cancelled
-
-    def __init__(self, *, synchronize: bool = True, use_lock: bool = True):
-        super().__init__()
-        self.synchronize = synchronize
+    def __init__(self, use_lock: bool = True):
```
borzunov, line 91:
```suggestion
    def __init__(self, *, use_lock: bool = True):
```

Otherwise, a user write accidentally call smth like `MPFuture(another_future_to_clone)` and don't get an exception right away.

hivemind/utils/mpfuture.py
```diff
@@ -101,213 +134,133 @@ def _set_event_if_necessary(self):
         async def _event_setter():
             self._aio_event.set()
 
-        if self._loop.is_running() and loop == self.get_loop():
+        if self.get_loop().is_running() and loop == self.get_loop():
```
borzunov, line 137: nit: I'd suggest to:

1. Remove the `self.get_loop()` method and just use `self._loop`
2. Rename the local variable `loop` to `running_loop`

This would make clear that `self._loop` is this MPFuture's event loop and `running_loop` is the current running event loop.

hivemind/utils/mpfuture.py
```diff
@@ -101,213 +134,133 @@ def _set_event_if_necessary(self):
...
-                    )
-                    cls._pipe_waiter_thread.start()
-
-    @classmethod
-    def reset_backend(cls):
-        """
-        Reset the MPFuture backend. This is useful when the state may have been corrupted
-        (e.g. killing child processes may leave the locks acquired and the background thread blocked).
+        logger.debug(f"Initializing MPFuture backend for pid {pid}")
+        assert pid != cls._active_pid, "already initialized"
```
borzunov, line 148:
```suggestion
```

nit: I'm not sure this assert is necessary since this condition is already checked twice before calling this method :)

tests/conftest.py
```diff
@@ -28,5 +29,7 @@ def cleanup_children():
             with suppress(psutil.NoSuchProcess):
                 child.kill()
 
-    # Broken code or killing of child processes may leave the MPFuture backend corrupted
-    MPFuture.reset_backend()
+    MPFuture._active_pid = None
```
borzunov, line 32: Can we keep the `reset_backend()` method and the comment here, so everything related to `MPFuture` internals is located in `mpfuture.py`?

hivemind/utils/mpfuture.py
```diff
@@ -335,9 +288,8 @@ def __del__(self):
 
     def __getstate__(self):
         return dict(
-            synchronize=self.synchronize,
             _sender_pipe=self._sender_pipe,
-            _state=self._state,
+            _shared_state_code=self._shared_state_code,
```
borzunov, line 292: Pickling does not seem to serialize/deserialize `_shared_state_code` in a way that it is still shared with the original byte.

hivemind/utils/mpfuture.py
justheuristic: thx, added

hivemind/utils/mpfuture.py
justheuristic: Everything you said is correct, but we are not using pickle for serialization. Instead, MPFutures are passed through ForkingPickler: a serializer that is meant for passing data between processes spawned with fork.

tests/conftest.py
justheuristic: done

hivemind/utils/mpfuture.py
justheuristic: done that

borzunov (approved):

yhn112 (approved):

PR #343, opened 2021-08-02
Speed up P2P client creation

### LINKED ISSUE #300: [Feature Request] Remove active waiting from hivemind.P2P initialization
As discussed in #299 , current hivemind.P2P will connect to a running daemon over several retries and wait a fixed `0.4 * 2 ** num_retries` seconds between retries.

This is suboptimal, as we will likely loose loose time sleeping after p2pd has already been initialized.

A better strategy is to modify p2pd so that it notifies the master process of successful initialization and simply await for that notification (or timeout).

deniskamazur:
This PR resolves #300 and speeds up the creation of P2P clients by reading the daemon's stdout instead of pinging it repeatedly. 

This optimization speeds up running the tests by 35% locally. Here are the benchmarks:
image
image

### DIFF
--- a/hivemind/p2p/p2p_daemon.py
+++ b/hivemind/p2p/p2p_daemon.py
@@ -453,7 +453,9 @@ class P2P:
 
     async def shutdown(self) -> None:
         await self._stop_listening()
-        await asyncio.get_event_loop().run_in_executor(None, self._terminate)
+        self._terminate()
+        if self._child is not None:
+            await self._child.wait()
 
     def _terminate(self) -> None:
         if self._reader_task is not None:
--- a/hivemind/p2p/p2p_daemon.py
+++ b/hivemind/p2p/p2p_daemon.py
@@ -159,7 +159,8 @@ class P2P:
         )
 
         self._child = await asyncio.subprocess.create_subprocess_exec(
-            *proc_args, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.STDOUT)
+            *proc_args, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.STDOUT
+        )
         self._alive = True
 
         ready = asyncio.Future()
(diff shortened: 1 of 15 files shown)


hivemind/p2p/p2p_daemon.py
```diff
@@ -157,37 +155,18 @@ async def create(
             autoRelay=use_auto_relay,
             relayHopLimit=relay_hop_limit,
             b=need_bootstrap,
-            q=quiet,
+            q=False,
```
borzunov, line 158:
```suggestion
```

`q=False` is equivalent to just skipping it.

borzunov:
Fantastic results!

One more thing to check is how the daemon behave in case of incorrect arguments. Ideally, we'd like to capture the daemon's stderr, read the error message, and convert it to a Python error with this message.

hivemind/p2p/p2p_daemon.py
```diff
@@ -68,8 +67,8 @@ def __init__(self):
         self.peer_id = None
         self._child = None
         self._alive = False
+        self._reader_task = None
         self._listen_task = None
-        self._server_stopped = asyncio.Event()
```
borzunov, line 72: This event is redundant, we can just cancel `self._listen_task`.

hivemind/p2p/p2p_daemon.py
```diff
@@ -437,20 +419,10 @@ def iterate_protobuf_handler(
     def _start_listening(self) -> None:
         async def listen() -> None:
             async with self._client.listen():
-                await self._server_stopped.wait()
+                await asyncio.Future()  # Wait until this task will be cancelled in _terminate()
 
         self._listen_task = asyncio.create_task(listen())
 
-    async def _stop_listening(self) -> None:
```
borzunov, line 444: Cancelling `self._listen_task` is enough here. The corresponding code moved to `self._terminate()` (since it is not async anymore).

hivemind/p2p/p2p_daemon.py
```diff
@@ -504,8 +481,22 @@ def _convert_process_arg_type(val: Any) -> Any:
...
+            line = await self._child.stdout.readline()
+            if not line:  # Stream closed
+                break
+            last_line = line.rstrip().decode(errors="ignore")
+
+            if last_line.startswith("Peer ID:"):
+                ready.set_result(None)
+
+        if not ready.done():
+            ready.set_exception(P2PDaemonError(f"Daemon failed to start: {last_line}"))
```
justheuristic, line 496: As it appears, P2P.create will hang if daemon does not print the desired line but does not stop either. If so, it would be great to introduce some kind of timeout for ready.set_result and consider daemon failed if it doesn't succeed in time

justheuristic (approved):

borzunov (approved):
I approve @deniskamazur's changes, and Denis approves mine as well :)

PR #347, opened 2021-08-03
Propagate startup errors from DHT and averager processes

borzunov:
Currently, startup failure of DHT and/or averager (e.g. when `initial_peers` are unreachable) leads to a deadlock in the parent process. The exception message is printed to stdout but the parent process does not stop, and this is not nice for tests and end users.

This PR makes startup errors to be propagated to the parent process.

### DIFF
--- a/hivemind/averaging/averager.py
+++ b/hivemind/averaging/averager.py
@@ -3,13 +3,13 @@
 from __future__ import annotations
 
 import asyncio
+import concurrent.futures
 import contextlib
 import ctypes
 import multiprocessing as mp
 import os
 import threading
 import weakref
-from concurrent.futures.thread import ThreadPoolExecutor
 from dataclasses import asdict
 from typing import Any, AsyncIterator, Dict, Optional, Sequence, Tuple, Union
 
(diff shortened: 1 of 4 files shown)


justheuristic (approved):

hivemind/averaging/averager.py
```diff
@@ -214,25 +214,31 @@ def _run_internal(self):
...
+                    if not self.client_mode:
+                        asyncio.create_task(self._declare_for_download_periodically())
+
+                    self._pending_group_assembled = asyncio.Event()
+                    self._pending_group_assembled.set()
+                except Exception as e:
+                    # Loglevel is DEBUG since normally the exception is propagated to the caller
+                    logger.debug(e, exc_info=True)
+                    self._ready.set_exception(e)
+                    return
```
mryab, line 240: Do I understand correctly that the child averager process will still get shut down because of return?

mryab (approved):

hivemind/averaging/averager.py
borzunov: Yes, `return` will finish the `_run()` coroutine, which will finish `loop.run_until_complete(_run())`, which will finish the child process.

PR #348, opened 2021-08-09
Add daemon termination support

deniskamazur:
Currently, there are situations in which the daemon's clients fail to send a SIGTERM, which leads to dangling p2pd instances. This PR fixes this issue by killing the daemon as soon as the persistent connection counter hits zero (but only if at least one has been open). The PR also adds an idleTimeout flag. If the timeout's been reached and no persistent connections have been opened, the daemon terminates itself.

### DIFF
--- a/hivemind/p2p/p2p_daemon.py
+++ b/hivemind/p2p/p2p_daemon.py
@@ -91,6 +91,7 @@ class P2P:
         use_auto_relay: bool = False,
         relay_hop_limit: int = 0,
         startup_timeout: float = 15,
+        idle_timeout: int = 0,
     ) -> "P2P":
         """
         Start a new p2pd process and connect to it.
@@ -112,6 +113,8 @@ class P2P:
         :param use_auto_relay: enables autorelay
         :param relay_hop_limit: sets the hop limit for hop relays
         :param startup_timeout: raise a P2PDaemonError if the daemon does not start in ``startup_timeout`` seconds
+        :param idle_timeout: kill daemon if client has been idle for a given number of
+                             seconds before opening persistent streams
         :return: a wrapper for the p2p daemon
         """
 
@@ -151,6 +154,7 @@ class P2P:
             relayDiscovery=use_relay_discovery,
             autoRelay=use_auto_relay,
             relayHopLimit=relay_hop_limit,
+            idleTimeout=str(idle_timeout) + "s",
             b=need_bootstrap,
             **process_kwargs,
         )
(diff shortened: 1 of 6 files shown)


hivemind/p2p/p2p_daemon.py
```diff
@@ -151,6 +154,7 @@ async def create(
             relayDiscovery=use_relay_discovery,
             autoRelay=use_auto_relay,
             relayHopLimit=relay_hop_limit,
+            idleTimeout=str(idle_timeout) + "s",
```
borzunov, line 157:
```suggestion
            idleTimeout=f"{idle_timeout}s",
```

nit: We usually prefer f-strings over concatenation in our code base.

hivemind/p2p/p2p_daemon.py
```diff
@@ -91,6 +91,7 @@ async def create(
         use_auto_relay: bool = False,
         relay_hop_limit: int = 0,
         startup_timeout: float = 15,
+        idle_timeout: int = 0,
```
borzunov, line 94:
```suggestion
        idle_timeout: float = 0,
```

Usually, timeouts are `float` (e.g. this is useful to specify the timeout such that `0 < idle_timeout < 1`). Can we implement it here as well?

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -74,23 +74,42 @@ class ControlClient:
...
 
         self._read_task: Optional[asyncio.Task] = None
         self._write_task: Optional[asyncio.Task] = None
 
+    @classmethod
+    async def create(
+        cls,
+        daemon_connector: DaemonConnector,
+        listen_maddr: Multiaddr = Multiaddr(DEFAULT_LISTEN_MADDR),
+        do_connect: bool = True,
```
borzunov, line 104:
```suggestion
        use_persistent_conn: bool = True,
```

Let's use a clear name and add a docstring clarifying that disabling this flag breaks unary handlers.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -74,23 +74,42 @@ class ControlClient:
...
 
     def __init__(
-        self, daemon_connector: DaemonConnector, listen_maddr: Multiaddr = Multiaddr(DEFAULT_LISTEN_MADDR)
+        self,
+        daemon_connector: DaemonConnector,
+        listen_maddr: Multiaddr = Multiaddr(DEFAULT_LISTEN_MADDR),
+        *,
+        _initialized_with_create=False,
     ) -> None:
+        assert _initialized_with_create, " Please use ControlClient.create coroutine to spawn new control instances "
```
borzunov, line 83: 👍

borzunov (changes requested):
Thanks for the PR! I've left a bunch of minor comments to fix.

hivemind/p2p/p2p_daemon_bindings/control.py
borzunov:
```suggestion
        assert _initialized_with_create, "Please use ControlClient.create coroutine to spawn new control instances "
```

borzunov (approved):
Approving this with one minor suggestion :)

PR #352, opened 2021-08-18
Fix flaps in test_p2p_daemon_bindings.py

borzunov:
@deniskamazur has found that tests in master fail with an exception in [4 of 90 runs](https://github.com/learning-at-home/hivemind/actions/runs/1139793757).

3 of 4 fails are related to race conditions with `get_free_port()` leading to the `Address already in use` error in `test_p2p_daemon_bindings.py`.

This PR:

1. Fixes the race conditions by switching to Unix sockets for client-daemon communication (as in the `hivemind.p2p.P2P`) and `/tcp/0` instead of `tcp/{get_free_port()}` for `-hostAddr`.
2. Fixes the code to remove the sockets after closing the server.
3. Adds a note about race condition to the `get_free_port()` docstring to discourage future users.
4. Removes a duplicate of `test_client_identify()` (surprisingly, there were two identical tests).
5. Decreases timeout values for timeout tests to avoid flaps like [this](https://github.com/learning-at-home/hivemind/runs/3358128875).

### DIFF
--- a/tests/test_p2p_daemon_bindings.py
+++ b/tests/test_p2p_daemon_bindings.py
@@ -18,7 +18,7 @@ from hivemind.p2p.p2p_daemon_bindings.utils import (
 )
 from hivemind.proto import p2pd_pb2 as p2pd_pb
 
-from test_utils.p2p_daemon import connect_safe, make_p2pd_pair_ip4
+from test_utils.p2p_daemon import connect_safe, make_p2pd_pair_unix
 
 
 def test_raise_if_failed_raises():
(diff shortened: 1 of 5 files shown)


tests/test_p2p_daemon_bindings.py
```diff
@@ -376,11 +376,6 @@ async def p2pcs():
         yield tuple(p2pd_tuple.client for p2pd_tuple in p2pd_tuples)
 
 
-@pytest.mark.asyncio
-async def test_client_identify_unix_socket(p2pcs):
-    await p2pcs[0].identify()
```
borzunov, line 381: Identical to the test below.

justheuristic (approved):
Well done!

PR #354, opened 2021-08-20
Refactor: extract tests for DHTProtocol from test_dht_node, re-write with asyncio

justheuristic:
This test addresses 2 minor requests from @borzunov  and @mryab posed in https://github.com/learning-at-home/hivemind/issues/276

* Move test_dht_protocol() and test_empty_table() from test_dht_node.py to a new file test_dht_protocol.py (these tests use only DHTProtocol, not DHTNode)
*  rewrite test_dht_protocol, test_empty_table and test_dht_node to use await instead of loop.run_until_complete
* made "bootstrap failed" from warning into a (bypassable) error

### DIFF
--- a/hivemind/dht/node.py
+++ b/hivemind/dht/node.py
@@ -235,7 +235,11 @@ class DHTNode:
                 finished_pings |= finished_in_time
 
             if not finished_pings or all(ping.result() is None for ping in finished_pings):
-                logger.warning("DHTNode bootstrap failed: none of the initial_peers responded to a ping.")
+                message = "DHTNode bootstrap failed: none of the initial_peers responded to a ping."
+                if validate:
+                    raise ValueError(f"{message} (set validate=False to ignore)")
+                else:
+                    logger.warning(message)
 
             if strict:
                 for task in asyncio.as_completed(finished_pings):
--- a/hivemind/dht/node.py
+++ b/hivemind/dht/node.py
@@ -237,7 +237,7 @@ class DHTNode:
             if not finished_pings or all(ping.result() is None for ping in finished_pings):
                 message = "DHTNode bootstrap failed: none of the initial_peers responded to a ping."
                 if validate:
-                    raise ValueError(f"{message} (set validate=False to ignore)")
+                    raise RuntimeError(f"{message} (set validate=False to ignore)")
                 else:
                     logger.warning(message)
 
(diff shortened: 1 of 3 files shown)


borzunov (approved):

hivemind/dht/node.py
```diff
@@ -235,7 +235,11 @@ async def create(
                 finished_pings |= finished_in_time
 
             if not finished_pings or all(ping.result() is None for ping in finished_pings):
-                logger.warning("DHTNode bootstrap failed: none of the initial_peers responded to a ping.")
+                message = "DHTNode bootstrap failed: none of the initial_peers responded to a ping."
+                if validate:
+                    raise RuntimeError(f"{message} (set validate=False to ignore)")
+                else:
+                    logger.warning(message)
```
borzunov, line 242:
```suggestion
                raise RuntimeError("DHTNode bootstrap failed: none of the initial_peers responded to a ping")
```

tests/test_dht_protocol.py
```diff
@@ -0,0 +1,163 @@
...
+        dummy_key = DHTID.generate()
+        empty_item, nodes_found_2 = (await protocol.call_find(peer2_id, [dummy_key]))[dummy_key]
+        assert empty_item is None, "Non-existent keys shouldn't have values"
+        (recv_id, recv_peer_id) = next(iter(nodes_found_2.items()))
+        assert (
+            recv_id == peer1_node_id and recv_peer_id == peer1_id
+        ), f"expected id={peer1_node_id}, peer={peer1_id} but got {recv_id}, {recv_peer_id}"
+
+        # cause a non-response by querying a nonexistent peer
+        assert (await protocol.call_find(PeerID.from_base58("fakeid"), [key])) is None
```
mryab, line 107: After rewriting to coroutines, IMO it becomes even more clear that we need to separate the "await" and "operate on await result" steps here. Aside from readability, one possible downside is that it becomes harder to find out which step failed in case of an error

mryab (approved):

tests/test_dht_protocol.py
justheuristic: [re-written most of such lines, only kept (await ...) for a few short expressions]

hivemind/dht/node.py
justheuristic: Summary of a short debate: while there are indeed reasons to always throw this warning, we ultimately agreed to keep the old behavior for 2 reasons:
* validate is True by default and it is difficult to change it without being very explicit
* there is a use case there not raising error is preferable: when several peers use pre-generated peer ids, each peer should connect to others, but some peer will inevitably become "first". Thus, we will need to set this flag to avoid try-except clauses and repeated daemon creation.

However, we also agreed to rename this flag to something more specific.

PR #357, opened 2021-08-21
Improve Matchmaking finalizers

borzunov:
This PR resolves the frequent warning below, as well as a number of other finalizer issues.

```
Task was destroyed but it is pending!
task: <Task pending name='Task-11' coro=<PotentialLeaders._declare_averager_periodically() done, defined at /home/jheuristic/Documents/exp/hivemind/hivemind/averaging/matchmaking.py:513> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7efcec662fd0>()]>>
```

__Discussion:__

- The new code does not return from `begin_search()` until `declare_averager(..., looking_for_group=False)` is finished. While it is idiomatic for asyncio, this makes averaging a little longer. Is it a desired behavior? Or maybe we should just suppress the `Task was destroyed but it is pending!` warning instead? Or maybe add to an averager-wide pool of background tasks?

### DIFF
--- a/hivemind/averaging/matchmaking.py
+++ b/hivemind/averaging/matchmaking.py
@@ -15,7 +15,7 @@ from hivemind.dht import DHT, DHTID, DHTExpiration
 from hivemind.p2p import P2P, P2PContext, P2PHandlerError, PeerID, ServicerBase
 from hivemind.proto import averaging_pb2
 from hivemind.utils import TimedStorage, get_dht_time, get_logger, timed_storage
-from hivemind.utils.asyncio import anext
+from hivemind.utils.asyncio import anext, cancel_and_wait
 
 logger = get_logger(__name__)
 
@@ -127,8 +127,9 @@ class Matchmaking:
                 raise
 
             finally:
-                request_leaders_task.cancel()
+                await cancel_and_wait(request_leaders_task)
                 self.assembled_group.cancel()
+
                 while len(self.current_followers) > 0:
                     await self.follower_was_discarded.wait()
                     self.follower_was_discarded.clear()
@@ -158,6 +159,8 @@ class Matchmaking:
                         elif len(self.current_followers) > 0:
                             await self.leader_disband_group()
                         continue
+                except (concurrent.futures.CancelledError, asyncio.CancelledError):
+                    break  # note: this is a compatibility layer for python3.7
                 except Exception as e:
                     if not self.assembled_group.done():
                         self.assembled_group.set_exception(e)
@@ -292,6 +295,8 @@ class Matchmaking:
                 ordered_peer_ids=[item.to_bytes() for item in group_info.peer_ids],
                 gathered=group_info.gathered,
             )
+        except (concurrent.futures.CancelledError, asyncio.CancelledError):
+            return  # note: this is a compatibility layer for python3.7
         except Exception as e:
             logger.exception(e)
             yield averaging_pb2.MessageFromLeader(code=averaging_pb2.INTERNAL_ERROR)
@@ -407,9 +412,9 @@ class PotentialLeaders:
             try:
                 yield self
             finally:
-                update_queue_task.cancel()
+                await cancel_and_wait(update_queue_task)
                 if declare:
-                    declare_averager_task.cancel()
+                    await cancel_and_wait(declare_averager_task)
 
                 for field in (
                     self.past_attempts,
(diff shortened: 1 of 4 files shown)


hivemind/averaging/matchmaking.py
```diff
@@ -127,10 +127,8 @@ async def look_for_group(self, *, data_for_gather: bytes, timeout: Optional[floa
                 raise
 
             finally:
-                if not request_leaders_task.done():
```
borzunov, line 130: `.cancel()` just returns False when a task/future is already done.

hivemind/averaging/matchmaking.py
```diff
@@ -160,8 +158,6 @@ async def _request_join_potential_leaders(self, timeout: Optional[float]) -> Gro
                         elif len(self.current_followers) > 0:
                             await self.leader_disband_group()
                         continue
-                except (concurrent.futures.CancelledError, asyncio.CancelledError):
```
justheuristic, line 163: The reason we needed this was to avoid propagating CancelledError that would trigger from networking issues.

It would appear that this is indeed no longer necessary, but we may need to revert this one specific change if we encounter stochastic collaboration-wide freezes after peers disconnect.

hivemind/averaging/matchmaking.py
```diff
@@ -521,8 +512,6 @@ async def _declare_averager_periodically(self, key_manager: GroupKeyManager):
                     await asyncio.sleep(self.declared_expiration_time - get_dht_time())
                     if self.running.is_set() and len(self.leader_queue) == 0:
                         await key_manager.update_key_on_not_enough_peers()
-            except (concurrent.futures.CancelledError, asyncio.CancelledError):
-                pass  # note: this is a compatibility layer for python3.7
```
justheuristic, line 525: [non-blocking]
Q: are you certain that we no longer need to handle concurrent.futures.CancelledError  here or is it an educated guess?

justheuristic (approved):

hivemind/averaging/matchmaking.py
borzunov: I have edited this PR to only remove the `except` statements in `_update_queue_periodically()` and `_declare_averager_periodically()` methods.

These methods are only awaited in `cancel_and_wait()` that handles both cancel and normal errors by itself.

hivemind/averaging/matchmaking.py
```diff
@@ -477,37 +475,31 @@ def request_expiration_time(self) -> float:
         else:
             return min(get_dht_time() + self.averaging_expiration, self.search_end_time)
 
-    async def _update_queue_periodically(self, key_manager: GroupKeyManager):
```
borzunov, line 480: The code is identical besides removing the `try-except` block (rationale is explained in [this](https://github.com/learning-at-home/hivemind/pull/357/files#r695271601) comment).

hivemind/averaging/matchmaking.py
```diff
@@ -127,10 +127,9 @@ async def look_for_group(self, *, data_for_gather: bytes, timeout: Optional[floa
                 raise
 
             finally:
-                if not request_leaders_task.done():
-                    request_leaders_task.cancel()
-                if not self.assembled_group.done():
```
borzunov, line 132: `.cancel()` just returns False when the awaitable is done.

hivemind/averaging/matchmaking.py
```diff
@@ -413,10 +412,9 @@ async def begin_search(self, key_manager: GroupKeyManager, timeout: Optional[flo
             try:
                 yield self
             finally:
-                if not update_queue_task.done():
-                    update_queue_task.cancel()
-                if declare and not declare_averager_task.done():
-                    declare_averager_task.cancel()
+                await cancel_and_wait(update_queue_task)
+                if declare:
+                    await cancel_and_wait(declare_averager_task)
```
borzunov, line 417: Using `cancel_and_wait()` here resolves this frequent warning:

```
Task was destroyed but it is pending!
task: <Task pending name='Task-11' coro=<PotentialLeaders._declare_averager_periodically() done, defined at /home/jheuristic/Documents/exp/hivemind/hivemind/averaging/matchmaking.py:513> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7efcec662fd0>()]>>
```

hivemind/averaging/matchmaking.py
borzunov: I've reverted this change where I'm not sure, see the [comment](https://github.com/learning-at-home/hivemind/pull/357#discussion_r695271601).

justheuristic (approved):

hivemind/dht/__init__.py
```diff
@@ -261,18 +261,11 @@ def run_coroutine(
     async def _run_coroutine(
         self, coro: Callable[[DHT, DHTNode], Awaitable[ReturnType]], future: MPFuture[ReturnType]
     ):
-        main_task = asyncio.create_task(coro(self, self._node))
```
borzunov, line 264: Cancels here did not work since the new MPFuture implementation does not support asyncio `await`s in a child process.

hivemind/dht/__init__.py
```diff
@@ -261,18 +261,11 @@ def run_coroutine(
...
-            if future.cancelled():
-                main_task.cancel()
-            else:
-                future.set_result(await main_task)
+            future.set_result(await coro(self, self._node))
         except BaseException as e:
-            logger.exception(f"Caught an exception when running a coroutine: {e}")
-            if not future.done():
-                future.set_exception(e)
+            logger.exception("Caught an exception when running a coroutine:", exc_info=True)
```
mryab, line 267:
```suggestion
            logger.exception("Caught an exception when running a coroutine:")
```

mryab (approved):

PR #358, opened 2021-08-23
.gif mods and google.yml safety features

LaudateCorpus1:


deniskamazur:
image

justheuristic:
Seriously though... from what I can tell, this is some kind of automatic activity that suggests *generally* useful things and maybe something else *wink*.

While some of the suggested changes appear useful, they mostly duplicate our existing pipelines (e.g. container building, separate security discussion) or add compatibility to services that we do not use.

If you insist that they need to be merged, __please advocate for them on a case by case basis__.

PR #362, opened 2021-08-25
Disable elasticity for averaging

justheuristic:
Currently, dynamic averager nbits is broken and we deliberately parameterize training in such a way that it will not be triggered. This pull-request removes the broken code.

In future, we can revive elasticity based on @yhn112 's earlier attempt at fixing it: https://github.com/learning-at-home/hivemind/commit/dc0ea7580e9a8cce0d7663a38d8647c356f646b6

This PR also fixes several minor issues:
* change DHT code in such a way that can recover from file/memory errors
* reduce the size of shmem buffer based on our performance benchmarks

### DIFF
--- a/hivemind/averaging/key_manager.py
+++ b/hivemind/averaging/key_manager.py
@@ -25,31 +25,17 @@ class GroupKeyManager:
     Utility class that declares and fetches averaging-related keys using a DHT
     """
 
-    RESERVED_KEY_FOR_NBITS = "::NBITS"
-
     def __init__(
         self,
         dht: DHT,
         prefix: str,
         initial_group_bits: Optional[str],
         target_group_size: int,
-        insufficient_size: Optional[int] = None,
-        excessive_size: Optional[int] = None,
-        nbits_expiration: float = 60,
-        nbits_rewrite_grace_period: float = 15,
     ):
-        assert initial_group_bits is None or all(bit in "01" for bit in initial_group_bits)
-        if initial_group_bits is None:
-            search_result = dht.get(f"{prefix}.0b", latest=True)
-            initial_group_nbits = self.get_suggested_nbits(search_result) or 0
-            initial_group_bits = "".join(random.choice("01") for _ in range(initial_group_nbits))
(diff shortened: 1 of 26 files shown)


hivemind/averaging/key_manager.py
```diff
@@ -93,51 +78,16 @@ async def get_averagers(self, group_key: GroupKey, only_active: bool) -> List[Tu
...
-            and suggested_nbits != len(self.group_bits)
-            and suggested_nbits != self.suggested_nbits
-        ):
-            self.suggested_nbits = suggested_nbits
-            logger.warning(f"{self.peer_id} - another averager suggested {self.suggested_nbits}-bit keys")
-        elif num_active_averagers >= self.excessive_size:
-            self.suggested_nbits = max(suggested_nbits or 0, len(self.group_bits) + 1)
-            logger.warning(f"{self.peer_id} - too many peers in bucket, switching to {self.suggested_nbits}-bit keys")
+        averagers = []
+        for key, entry in result.value.items():
```
mryab, line 82: IMO `entry` is less clear than `looking_for_group`

hivemind/averaging/key_manager.py
```diff
@@ -93,51 +78,16 @@ async def get_averagers(self, group_key: GroupKey, only_active: bool) -> List[Tu
...
+            try:
+                if only_active and not entry.value:
+                    continue
+                averagers.append((PeerID(key), entry.expiration_time))
+            except Exception as e:
+                logger.warning(f"Could not parse group key {key} ({entry}, exc={e})")
```
mryab, line 88: logger.exception? Unparseable keys are not a sign of normal operation

mryab (approved):

hivemind/averaging/key_manager.py
justheuristic: fixd

hivemind/averaging/key_manager.py
justheuristic: Alas it isn't: it may happen if some peers are messing around, i would like to insist in logs that this is not critical for us

PR #369, opened 2021-08-29
Handle persistent connection writer exceptions

deniskamazur:
Writing to the persistent daemon connection may fail due to various reasons, which leads to the writing coroutine exiting. This PR implements handling of these errors.

### DIFF
--- a/hivemind/p2p/p2p_daemon_bindings/control.py
+++ b/hivemind/p2p/p2p_daemon_bindings/control.py
@@ -192,7 +192,10 @@ class ControlClient:
         with closing(writer):
             while True:
                 msg = await self._pending_messages.get()
-                await write_pbmsg(writer, msg)
+                try:
+                    await write_pbmsg(writer, msg)
+                except BaseException as e:
+                    logger.error(f"Failed to write to persistent connection, caught {e}.")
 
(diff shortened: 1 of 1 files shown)


justheuristic (changes requested):
As far as i remember, the main issue we're addressing popped out in a different spot:
https://github.com/learning-at-home/hivemind/blob/master/hivemind/p2p/p2p_daemon_bindings/control.py#L158-L161

but having this safeguard on the reading side is also appreciated

borzunov:
Thanks for the PR.

Usually, `await write_pbmsg(writer, msg)` fails due to a connection error, and subsequent writes would also fail. Therefore, the fact that `_write_to_persistent_conn()` exits after the first fail is a reasonable behavior, and I think we should keep it as is.

Other issues:

- We should catch `Exception` instead of `BaseException` (otherwise, `CancelledError` is not propagated, and the coroutine starts to occasionally ignore cancel).
- We should log exceptions with `log.exception()` instead of `log.error()` to show a traceback. For other loglevels (debug, warning, etc.), we should use `exc_info=True` to show a traceback.

I will close this for now, but feel free to continue discussion if you disagree.

PR #371, opened 2021-08-31
Fix "Too many open files" and load state freezing

justheuristic:


hivemind/optim/collaborative.py
```diff
@@ -185,7 +191,14 @@ def is_alive(self) -> bool:
     def load_state_from_peers(self, **kwargs):
         """Attempt to fetch the newest collaboration state from other peers"""
         with self.lock_collaboration_state:
-            self.averager.load_state_from_peers(**kwargs)
+            while True:
+                try:
+                    self.averager.load_state_from_peers(timeout=self.load_state_timeout, **kwargs)
+                    break
+                except BaseException as e:
```
borzunov, line 198:
```suggestion
                except Exception as e:
```

Otherwise, `KeyboardInterrupt` is ignored. Also, I doubt that we need to catch any exceptions not inherited from `Exception`.

hivemind/averaging/averager.py
```diff
@@ -554,7 +566,7 @@ def load_state_from_peers(self, wait=True) -> Optional[Tuple[Any, Sequence[torch
         """
         future = MPFuture()
         self._outer_pipe.send(("_load_state_from_peers", [], dict(future=future)))
-        return future.result() if wait else future
+        return future.result(timeout=timeout) if wait else future
```
borzunov, line 569: Is not it more practical to set the max time for loading the state _from one peer_?

I'm okay if this is addressed in a future PR.

hivemind/utils/asyncio.py
```diff
@@ -127,3 +127,14 @@ async def _put_items():
...
             task.cancel()
+
+
+async def aiter_with_timeout(iterable: AsyncIterable[T], timeout: float) -> AsyncIterator[T]:
+    """Iterate over an async iterable, raise TimeoutError if another portion of data does not arrive within timeout"""
+    # based on https://stackoverflow.com/a/50245879
+    iterator = iterable.__aiter__()
+    while True:
+        try:
+            yield await asyncio.wait_for(iterator.__anext__(), timeout=timeout)
```
borzunov, line 138: Can you please add a test for this? I am not sure cancelling `__anext__()` works fine.

Example: create an async generator that sleeps for 0.2 seconds before yielding a value, then iterate it with timeouts 0.1 and 0.3.

hivemind/averaging/averager.py
```diff
@@ -211,48 +215,56 @@ def _run_internal(self):
...
-                    if method == "_shutdown":
-                        await task
-                        break
-
-            loop.run_until_complete(_run())
+                    await asyncio.wait_for(pipe_semaphore.acquire(), timeout=self.request_timeout)
+                except asyncio.TimeoutError:
+                    pass
+                if not self._inner_pipe.poll():
+                    continue
```
borzunov, line 255:
```suggestion
                await pipe_semaphore.acquire()
```

As agreed, let's apply this in a future PR since it requires extra testing.

hivemind/dht/__init__.py
```diff
@@ -102,45 +102,52 @@ def __init__(
...
-                        await task
-                        break
-
-            coro = _run()
-            loop.run_until_complete(coro)
+                    await asyncio.wait_for(pipe_semaphore.acquire(), timeout=self._node.protocol.wait_timeout)
+                except asyncio.TimeoutError:
+                    pass
+                if not self._inner_pipe.poll():
+                    continue
```
borzunov, line 137:
```suggestion
                await pipe_semaphore.acquire()
```

As agreed, let's apply this in a future PR since it requires extra testing.

hivemind/dht/__init__.py
```diff
@@ -102,45 +102,52 @@ def __init__(
...
+                try:
+
```
borzunov, line 139:
```suggestion
```

nit: Let's remove the newline after `try:` (though I'm okay with a newline _before_ try).

borzunov (changes requested):

hivemind/averaging/averager.py
justheuristic: moved to https://github.com/learning-at-home/hivemind/issues/372

hivemind/averaging/averager.py
justheuristic: https://github.com/learning-at-home/hivemind/issues/372

hivemind/optim/collaborative.py
justheuristic: https://github.com/learning-at-home/hivemind/issues/372

hivemind/averaging/averager.py
```diff
@@ -554,7 +554,9 @@ async def _get_current_state_from_host_process(self):
         self._inner_pipe.send(("_TRIGGER_GET_CURRENT_STATE", future))
         return await future
 
-    def load_state_from_peers(self, wait=True, timeout: Optional[float] = None) -> Optional[Tuple[Any, Sequence[torch.Tensor]]]:
+    def load_state_from_peers(
+        self, wait=True, timeout: Optional[float] = None
```
borzunov, line 558:
```suggestion
        self, wait: bool = True, timeout: Optional[float] = None
```

borzunov (approved):

yhn112 (approved):
LGTM, if the fix is real

PR #376, opened 2021-09-01
Parametrize max message size for persistent connections

deniskamazur:
Go part: https://github.com/learning-at-home/go-libp2p-daemon/pull/15.


hivemind/p2p/p2p_daemon.py
```diff
@@ -95,6 +95,7 @@ async def create(
         use_relay: bool = True,
         use_relay_hop: bool = False,
         use_relay_discovery: bool = False,
+        persistent_conn_max_msg_size: int = 2 << 22,
```
borzunov, line 98: 1. The default `2 << 22` does not match the Go default `1 << 22` (seems like a typo).

2. Let's write it as `4 * 1024 ** 2`, so it is more clear that this is 4 MiB.

3. Having the default in 3 places is misleading: if a user changes one of them that they accidentally find, the behavior won't change as expected. Let's create the constant `DEFAULT_MAX_MSG_SIZE` in the most inner module (`control.py`), then set all defaults to this constant.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -107,8 +110,14 @@ async def create(
         daemon_connector: DaemonConnector,
         listen_maddr: Multiaddr = Multiaddr(DEFAULT_LISTEN_MADDR),
         use_persistent_conn: bool = True,
+        persistent_conn_max_msg_size=2 << 22,
```
borzunov, line 113:
```suggestion
        persistent_conn_max_msg_size: int = DEFAULT_MAX_MSG_SIZE,
```

The typo annotation is missing.

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -255,6 +270,9 @@ async def call_unary_handler(self, peer_id: PeerID, proto: str, data: bytes) ->
             callUnary=call_unary_req,
         )
 
+        if req.ByteSize() > self.persistent_conn_max_msg_size:
+            raise P2PHandlerError(f"Message size exceeds set limit {self.persistent_conn_max_msg_size}")
```
borzunov, line 274: I would say that `P2PHandlerError` is a error of the __remote__ handler. Let's use `P2PDaemonError`.

```suggestion
            raise P2PDaemonError(f"Message size exceeds set limit {self.persistent_conn_max_msg_size}")
```

borzunov (changes requested):

borzunov (approved):
Awesome, thanks for the contribution :)

Don't forget to change the daemon version in `setup.py`.

setup.py
```diff
@@ -14,9 +14,10 @@
 from setuptools.command.build_py import build_py
 from setuptools.command.develop import develop
 
-P2PD_VERSION = "v0.3.5"
-P2PD_CHECKSUM = "affea8ec63dbe2423ef7453718b5798d"
+P2PD_VERSION = "v0.3.6"
+P2PD_CHECKSUM = "627d0c3b475a29331fdfd1667e828f6d"
 LIBP2P_TAR_URL = f"https://github.com/learning-at-home/go-libp2p-daemon/archive/refs/tags/{P2PD_VERSION}.tar.gz"
```
borzunov, line 19:
```suggestion
P2PD_SOURCE_URL = f"https://github.com/learning-at-home/go-libp2p-daemon/archive/refs/tags/{P2PD_VERSION}.tar.gz"
```

Because it is still the p2p daemon like in `P2PD_BINARY_URL`, not the libp2p itself :)

PR #378, opened 2021-09-04
Make log handlers configurable, shorten entries

borzunov:
## Preliminaries

The current implementation of `hivemind.utils.logging` has several problems (they are not related to the colored logging and exist for a long time):

1. __Bugs:__ Sometimes, one message is logged multiple times. This due to the combination of the bugs:

    - If `name` stays the same, `logging.getLogger(name)` always returns the same logger instance. We expect the same behavior from hivemind's `get_logger(name)`, however it adds a new `logging.StreamHandler` to the logger instance every time. If it is called N times, you will have N stream handlers and each message will be repeated N times.
 
    - hivemind's `get_logger(name)` trims the first item in the module path. While this is intended to make the log line shorter by trimming the `hivemind.` string, this actually forces all modules at the root scope (e.g. `utils.py` and `huggingface_auth.py` at [tanmoyai/sahajbert](https://github.com/tanmoyio/sahajbert)) to use the same logger. Because of the previous bug, we end up logging the same message multiple times even if we use different `name`s.

2. It is not obvious how to force other libraries (such as `transformers`) to use our logging style, so we have __inconsistent log line styles__ in the example. Screenshot:

    Screen Shot 2021-09-04 at 6 35 08 AM

3. Since `hivemind` is a library, a developer may want to use it but keep the existing logging style in their application (i.e. force hivemind to follow it). Currently, there is no way to do it since `get_logger()` does not use message propagation to the application loggers.

## Solution

1. First, we fix the bugs above, making `get_logger()` be idempotent and trim the `hivemind.` prefix only while displaying lines.

2. Next, we note that there are 3 possible use cases:

    - _(default)_ A user tries importing hivemind. It uses its own logging style among the package.
    - A user likes the hivemind logging style and wants to use it among their application (e.g. force other libraries to follow it).
    - A user does not like the hivemind logging style and wants to force it to follow another existing style.

    We give a user a straightforward way to switch between these use cases via a special function:

    ```python
    use_hivemind_log_handler("in_hivemind")     # Option 1 (default)
    use_hivemind_log_handler("in_root_logger")  # Option 2, make the root logger to use hivemind style
    use_hivemind_log_handler("nowhere")         # Option 3, propagate hivemind logs to the existing root logger
    ```

    __Note:__ This approach is inspired by the `transformers.logging` module ([docs](https://huggingface.co/transformers/main_classes/logging.html), [source](https://github.com/huggingface/transformers/blob/master/src/transformers/utils/logging.py)). The module allows to enable/disable the propagation to the root logger and enable/disable the `transformers` default log style. However, our API is even higher-level.

3. We enable the `in_root_logger` mode in `examples/albert`, so that all messages (from `__main__`, `transformers`, and `hivemind` itself) consistently follow the hivemind style. Screenshot:

    Screen Shot 2021-09-04 at 6 33 03 AM

4. We change some log messages to improve their presentation.

### DIFF
--- a/examples/albert/run_trainer.py
+++ b/examples/albert/run_trainer.py
@@ -22,7 +22,7 @@ from hivemind.utils.logging import get_logger, use_hivemind_log_style
 import utils
 from arguments import AlbertTrainingArguments, AveragerArguments, CollaborationArguments, DatasetArguments
 
-use_hivemind_log_style('everywhere')
+use_hivemind_log_style("everywhere")
 logger = get_logger()
 
 LRSchedulerBase = getattr(torch.optim.lr_scheduler, "_LRScheduler", None)
--- a/examples/albert/run_trainer.py
+++ b/examples/albert/run_trainer.py
@@ -17,12 +17,12 @@ from transformers.trainer import Trainer
 from transformers.trainer_utils import is_main_process
 
 import hivemind
-from hivemind.utils.logging import get_logger, use_hivemind_log_style
+from hivemind.utils.logging import get_logger, use_hivemind_log_format
 
 import utils
 from arguments import AlbertTrainingArguments, AveragerArguments, CollaborationArguments, DatasetArguments
 
-use_hivemind_log_style("everywhere")
+use_hivemind_log_format("in_root_logger")
 logger = get_logger()
 
 LRSchedulerBase = getattr(torch.optim.lr_scheduler, "_LRScheduler", None)
--- a/examples/albert/run_trainer.py
+++ b/examples/albert/run_trainer.py
@@ -17,12 +17,12 @@ from transformers.trainer import Trainer
 from transformers.trainer_utils import is_main_process
 
 import hivemind
-from hivemind.utils.logging import get_logger, use_hivemind_log_format
+from hivemind.utils.logging import get_logger, use_hivemind_log_handler
 
 import utils
 from arguments import AlbertTrainingArguments, AveragerArguments, CollaborationArguments, DatasetArguments
 
-use_hivemind_log_format("in_root_logger")
+use_hivemind_log_handler("in_root_logger")
 logger = get_logger()
 
 LRSchedulerBase = getattr(torch.optim.lr_scheduler, "_LRScheduler", None)
--- a/examples/albert/run_trainer.py
+++ b/examples/albert/run_trainer.py
@@ -138,7 +138,7 @@ class CollaborativeCallback(transformers.TrainerCallback):
                     loss=self.loss,
                     mini_steps=self.steps,
                 )
-                logger.info(f"Step {self.collaborative_optimizer.local_step}")
+                logger.info(f"Step #{self.collaborative_optimizer.local_step}")
                 logger.info(f"Your current contribution: {self.total_samples_processed} samples")
                 logger.info(f"Performance: {samples_per_second} samples per second.")
                 if self.steps:
--- a/examples/albert/run_trainer.py
+++ b/examples/albert/run_trainer.py
@@ -18,32 +18,20 @@ from transformers.trainer import Trainer
 from transformers.trainer_utils import is_main_process
 
 import hivemind
+from hivemind.utils.logging import get_logger
 
 import utils
 from arguments import AlbertTrainingArguments, AveragerArguments, CollaborationArguments, DatasetArguments
 
-logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


examples/albert/run_trainer.py
```diff
@@ -18,32 +17,22 @@
...
-        format="%(asctime)s - %(levelname)s - %(name)s -   %(message)s",
-        datefmt="%m/%d/%Y %H:%M:%S",
-        level=logging.INFO if is_main_process(training_args.local_rank) else logging.WARN,
-    )
 
-    # Log on each process the small summary:
-    logger.warning(
-        f"Process rank: {training_args.local_rank}, device: {training_args.device}, n_gpu: {training_args.n_gpu}"
-        + f"distributed training: {bool(training_args.local_rank != -1)}, 16-bits training: {training_args.fp16}"
-    )
```
borzunov, line 40: This message is removed since `training_args` are already logged.

hivemind/utils/logging.py
```diff
@@ -60,23 +68,74 @@ def format(self, record: logging.LogRecord) -> str:
...
+    global _current_mode
+
+    if isinstance(where, str):
+        # We allow `where` to be a string, so a developer does not have to import the enum for one usage
+        where = StyleMode[where.upper()]
+
+    if _current_mode == StyleMode.AMONG_HIVEMIND:
+        _disable_default_handler(_PACKAGE_NAME)
+    elif _current_mode == StyleMode.EVERYWHERE:
+        _disable_default_handler(None)
```
mryab, line 131: I'm not sure if this is going to be a frequent enough use case, aside from our examples. Hivemind is first and foremost a library, and from my understanding, no other library implements this format hijacking mechanism: thus, I am not sure that this feature is actually required and expected to be present in a library about distributed DL. Probably, the reason for this is that it's usually expected that logs from the given library are consistent in their format in all applications for ease of support and issue reporting, and thus one has no direct incentive to implicitly adopt the logging format of other library

mryab (changes requested):

hivemind/utils/logging.py
```diff
@@ -60,23 +68,74 @@ def format(self, record: logging.LogRecord) -> str:
...
+
+
+def use_hivemind_log_style(where: Union[StyleMode, str]) -> None:
+    global _current_mode
+
+    if isinstance(where, str):
+        # We allow `where` to be a string, so a developer does not have to import the enum for one usage
+        where = StyleMode[where.upper()]
+
+    if _current_mode == StyleMode.AMONG_HIVEMIND:
```
mryab, line 128: Is it possible to have _current_mode uninitialized when calling this function? Won't that lead to an error?

hivemind/utils/logging.py
mryab: This isn't a blocker, I'm just afraid that adding unnecessary/rarely used features should not be our focus regarding logging

examples/albert/run_trainer.py
```diff
@@ -18,32 +17,22 @@
...
+logger = get_logger()
 
+LRSchedulerBase = getattr(torch.optim.lr_scheduler, "_LRScheduler", None)
 
-def setup_logging(training_args):
-    logging.basicConfig(
-        format="%(asctime)s - %(levelname)s - %(name)s -   %(message)s",
-        datefmt="%m/%d/%Y %H:%M:%S",
-        level=logging.INFO if is_main_process(training_args.local_rank) else logging.WARN,
-    )
```
borzunov, line 34: This `basicConfig()` setting had no effect:

_> This function does nothing if the root logger already has handlers configured, unless the keyword argument force is set to True._

hivemind/utils/logging.py
```diff
@@ -50,7 +55,10 @@ def format(self, record: logging.LogRecord) -> str:
             record.msecs = (record.created - int(record.created)) * 1000
 
         if not hasattr(record, "caller"):
-            record.caller = f"{record.name}.{record.funcName}:{record.lineno}"
+            module_path = record.name.split(".")
+            if module_path[0] == "hivemind":
+                module_path = module_path[1:]
```
borzunov, line 60: TODO: Consider removing this special case

hivemind/utils/logging.py
borzunov: Fixed

mryab (approved):

borzunov:
@justheuristic has verbally approved this.

PR #379, opened 2021-09-08
Implement CenteredClip in averager

borzunov:


hivemind/averaging/accumulators.py
```diff
@@ -0,0 +1,112 @@
...
+    n_clipped: torch.Tensor
+    last_step_delta: torch.Tensor
+
+
+def centered_clip(
+    input_tensors: torch.Tensor,
+    weights: torch.Tensor,
+    tau: float = 1.0,
+    n_iters: int = 20,
+    stop_delta: Optional[float] = None,
```
justheuristic, line 62: preference: let's default to some reasonable delta and a very large n steps

justheuristic:
Random thoughts:
- gotta ensure that we do not fail even if user sends nan/inf/near_inf values (both with compression and with raw values)
- would prefer to set delta to 1e-6 and increase n_iters dramatically (e.g. to 50) to make runtime adaptive
- let's print some warnings in case your values were more than e.g. 3 tau outside the range

hivemind/averaging/accumulators.py
```diff
@@ -0,0 +1,112 @@
...
+            coeffs = weights * torch.minimum(torch.tensor(1.0), tau / norms)
+
+            if stop_delta is not None:
+                prev_diff = result[...] = diff[0]  # Reuse memory from `result`
+
+            # We only need to update `diff` (not `result`) between iterations
+            diff.addmm_(-coeffs.repeat(n_peers, 1), diff)
+
+            if stop_delta is not None:
+                delta = prev_diff.sub_(diff[0]).max()
```
borzunov, line 105: TODO: .max() -> .abs().max()

borzunov:
```suggestion
                delta = prev_diff.sub_(diff[0]).abs().max()
```

hivemind/averaging/accumulators.py
```diff
@@ -0,0 +1,112 @@
...
+        result = sorted_tensors[n_peers // 2].clone()
+        delta = None
+
+        diff = torch.sub(input_tensors, result, out=sorted_tensors)  # Reuse memory from `sorted_tensors`
+        for _ in range(n_iters):
+            norms = diff.norm(dim=1)
+            coeffs = weights * torch.minimum(torch.tensor(1.0), tau / norms)
+
+            if stop_delta is not None:
+                prev_diff = result[...] = diff[0]  # Reuse memory from `result`
```
borzunov, line 99:
```suggestion
                result[...] = diff[0]  # Reuse memory from `result`
                prev_diff = result
```

hivemind/averaging/accumulators.py
```diff
@@ -0,0 +1,113 @@
...
+        for _ in range(n_iters):
+            norms = diff.norm(dim=1)
+            coeffs = weights * torch.minimum(torch.tensor(1.0), tau / norms)
+
+            if stop_delta is not None:
+                result[...] = diff[0]  # Reuse memory from `result`
+                prev_diff = result
+
+            # We only need to update `diff` (not `result`) between iterations
+            diff.addmm_(-coeffs.repeat(n_peers, 1), diff)
```
borzunov, line 103:
```suggestion
            diff -= coeffs @ diff
```

It seems like `addmm_()` doesn't work correctly if the destination is equal to one of the operands.

PR #380, opened 2021-09-10
Enable log handler in benchmarks and run_server

borzunov:
Currently, messages from the root logger are not displayed using hivemind logger (in particular, `LOGLEVEL` is ignored for the root logger, so the info messages are not visible). This PR fixes it:

Screen Shot 2021-09-10 at 11 49 43 PM

Also, it renames `LOGLEVEL` to `HIVEMIND_LOGLEVEL`.

### DIFF
--- a/benchmarks/benchmark_averaging.py
+++ b/benchmarks/benchmark_averaging.py
@@ -7,8 +7,11 @@ import torch
 
 import hivemind
 from hivemind.proto import runtime_pb2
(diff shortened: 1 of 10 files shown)


justheuristic (approved):
Looks good as far as i'm concerned, but i'm not an expert in the logging module

deniskamazur (approved):
Thanks for the fix!

mryab (approved):

PR #383, opened 2021-09-18
Fix step_tolerance in CollaborativeOptimizer

justheuristic:
This PR fixes a bug found by @yhn112 in our internal experiments

When a one peer have inadvertently ended up once step ahead of others, it would cause the following cycle:
- old peers would not load state from new ones because the new peer is still within step_tolerance
- old peers would not contribute to training minibatches because we did not account for step tolerance there
- new peers will maintain the 1-step distance from others because they made steps concurrently with the rest

In practice, this resulted in significant performance degradations for several (~10) steps infrequently during training

__new behavior:__ if a peer detects another peer who is ahead, but _within tolerance_, it will attempt to jump to that peer's step.

__Other changes:__
* changed default timeout from 30s to 1 minute to ensure that users will be able to run *something* even with a slow connection
* fixed a minor bug that caused wrong ETA to next step when collaboration consisted of exactly 1 peer

### DIFF
--- a/hivemind/optim/collaborative.py
+++ b/hivemind/optim/collaborative.py
@@ -451,7 +451,7 @@ class CollaborativeOptimizer(DecentralizedOptimizerBase):
 
         for state in valid_peer_states:
             total_samples_per_second += state.samples_per_second
-            if state.step >= self.local_step and current_time - state.time < self.staleness_timeout:
+            if state.step == self.local_step and current_time - state.time < self.staleness_timeout:
                 total_samples_accumulated += state.samples_accumulated
                 estimated_current_samples += (
                     state.samples_accumulated + max(0, current_time - state.time) * state.samples_per_second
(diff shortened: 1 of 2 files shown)


hivemind/optim/collaborative.py
```diff
@@ -149,7 +149,7 @@ def __init__(
 
         self.training_progress_key = f"{self.prefix}_progress"
         self.local_samples_accumulated = 0  # a number of local samples accumulated since last optimizer update
-        self.local_steps_accumulated = 0  # a number of calls to step() since last optimizer update
+        self.local_updates_accumulated = 0  # a number of calls to step() since last optimizer update
```
justheuristic, line 152: rationale: the term "step" is already used to represent optimizer steps

hivemind/optim/collaborative.py
```diff
@@ -401,9 +425,9 @@ def _fetch_state(self) -> CollaborationState:
 
         if not isinstance(response, dict) or len(response) == 0:
             logger.log(self.status_loglevel, f"Found no active peers: {response}")
-            local_eta_next_step = (
-                max(0, self.target_batch_size - self.local_steps_accumulated) / self.performance_ema.samples_per_second
```
justheuristic, line 405: this should have used "local_samples_accumulated"

yhn112 (approved):
Somewhat checked on a large scale. It works. Let's ship and see.

PR #384, opened 2021-09-18
Implement weight as part of the allreduce protocol, not matchmaking

justheuristic:
This PR allows specifying allreduce weights in AllReduceRunner, instead of gathering them during matchmaking.

This will allow peers to use their actual batch size in both DPU and advance matchmaking (aka @yhn112 -style matchmaking)

__[WIP]__ implement advance matchmaking as a working example

### DIFF
--- a/hivemind/averaging/allreduce.py
+++ b/hivemind/averaging/allreduce.py
@@ -182,7 +182,7 @@ class AllReduceRunner(ServicerBase):
             code=averaging_pb2.PART_FOR_AVERAGING,
             group_id=self.group_id,
             tensor_part=first_part,
-            metadata=self._weight_binary
+            metadata=self._weight_binary,
         )
         async for part in parts_aiter:
             yield averaging_pb2.AveragingData(tensor_part=part, metadata=self._weight_binary)
@@ -229,7 +229,7 @@ class AllReduceRunner(ServicerBase):
                 max_prefetch=self.tensor_part_container.prefetch,
             )
         ):
-            weight, = struct.unpack("d", part_metadata)
+            (weight,) = struct.unpack("d", part_metadata)
             averaged_part = await self.tensor_part_reducer.accumulate_part(
                 sender_index, part_index, tensor_part, weight=weight
             )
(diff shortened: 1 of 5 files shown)


hivemind/averaging/allreduce.py
```diff
@@ -37,13 +38,12 @@ class AllReduceRunner(ServicerBase):
     :param prefix: namespace for servicer's RPCs (typically, equal to prefix for group keys)
     :param group_id: unique identifier of this specific all-reduce run
     :param tensors: local tensors that should be averaged with groupmates
-    :param tensors: local tensors that should be averaged with groupmates
```
justheuristic, line 40: yeah, that was in the master code :)

hivemind/averaging/allreduce.py
```diff
@@ -180,9 +182,10 @@ async def _generate_input_for_peer(self, peer_index: int) -> AsyncIterator[avera
             code=averaging_pb2.PART_FOR_AVERAGING,
             group_id=self.group_id,
             tensor_part=first_part,
+            metadata=self._weight_binary,
```
borzunov, line 185: Can we create a new `double weight;` field instead and avoid encoding the weight to the binary format manually?

borzunov (changes requested):
I've left one comment, everything else is good :)

hivemind/averaging/allreduce.py
justheuristic: done so

hivemind/proto/averaging.proto
```diff
@@ -45,7 +45,7 @@ message AveragingData {
   bytes group_id = 2;       // a unique group identifier, same as in MessageFromLeader
   bytes peer_id = 3;        // sender's rpc peer_id, used for coordination
   Tensor tensor_part = 4;   // either peer's local tensor part (rpc input) or group average of this part (rpc output)
-  bytes metadata = 5;       // reserved user-extendable metadata
+  double weight = 5;         // peers will be averaged in proportion to these weights
```
borzunov, line 48:
```suggestion
  double weight = 5;        // peers will be averaged in proportion to these weights
```

nit: There are too many spaces currently :)

borzunov (approved):

PR #385, opened 2021-09-18
Simplified all-reduce for asymmetric tcp

justheuristic:
* [x] add an option to run sequential all-reduce, wherein a runner receives results only after it finishes sending tensors
* [x] ensure that calling stream RPC begins sending inputs right away, instead of when awaiting the first output
* [ ] right before merging, run a performance test on colab+nora

Note: on some systems (e.g. colab), one can get faster _bidirectional_ throughput by using BBR congestion control:
```
sudo cat <<EOF > /etc/sysctl.d/10-custom-kernel-bbr.conf
net.core.default_qdisc=fq
net.ipv4.tcp_congestion_control=bbr
EOF

sudo sysctl --system
```

However, this did not work for kaggle and looks kind of hacky in the script.


borzunov (approved):

borzunov (changes requested):
Before merging, I am curious to see it working on nora + Colab at least with `--averaging_timeout 120`.

Issue #381, opened 2021-09-11
Remote expert info is serialized with pickle


deniskamazur:
The hivemind server's `info` RPC [serializes data using pickle](https://github.com/learning-at-home/hivemind/blob/fcad8d0d63e0e0d04aa8016045916694be4132c0/hivemind/moe/server/connection_handler.py#L63). This is a critical vulnerability as it may lead to remote code execution.

borzunov:
Clarification for the readers: the vulnerability is present only in `hivemind.moe.server`. The current experiments (such as sahajBERT 2.0) do not use it, however future experiments may indeed start using MoE functionality.

borzunov:
Resolved in #386.

PR #386, opened 2021-09-19
Fix pickle vulnerability

deniskamazur:
This PR fixes a vulnerability in the hivemind server. For more information check out #381.

### DIFF
--- a/hivemind/utils/tensor_descr.py
+++ b/hivemind/utils/tensor_descr.py
@@ -94,15 +94,8 @@ class BatchTensorDescriptor(TensorDescriptor):
         obj_dict = asdict(self)
         obj_dict.pop("device")
 
-        if self.dtype is not None:
-            obj_dict["dtype"] = str(self.dtype)
-        else:
-            obj_dict.pop("dtype")
-
-        if self.layout is not None:
-            obj_dict["layout"] = str(self.layout)
-        else:
-            obj_dict.pop("layout")
+        obj_dict["dtype"] = str(self.dtype) if self.dtype is not None else None
+        obj_dict["layout"] = str(self.layout) if self.layout is not None else None
 
         if self.device is not None:
             obj_dict.update(
@@ -116,10 +109,10 @@ class BatchTensorDescriptor(TensorDescriptor):
     def unpackb(cls, raw: bytes) -> BatchTensorDescriptor:
         obj_dict = MSGPackSerializer.loads(raw)
 
-        if "dtype" in obj_dict:
+        if obj_dict["dtype"] is not None:
             obj_dict["dtype"] = _str_to_torch_type(obj_dict["dtype"], torch.dtype)
 
-        if "layout" in obj_dict:
+        if obj_dict["layout"] is not None:
             obj_dict["layout"] = _str_to_torch_type(obj_dict["layout"], torch.layout)
 
         if "device_type" in obj_dict:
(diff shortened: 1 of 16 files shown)


justheuristic (approved):
<:+1: />

It is strange, though, that the packb/unpackb is reported as not covered by tests

hivemind/utils/tensor_descr.py
```diff
@@ -70,13 +83,55 @@ def from_tensor(cls, tensor: torch.Tensor, compression=CompressionType.NONE) ->
...
         return super().make_empty(size=(*batch_size, *self.shape[1:]), **kwargs)
 
+    def packb(self) -> bytes:
+        obj_dict = asdict(self)
+        obj_dict.pop("device")
+
+        if self.dtype is not None:
+            obj_dict["dtype"] = str(self.dtype)
+        else:
+            obj_dict.pop("dtype")
```
borzunov, line 100:
```suggestion
```

I suggest to remove this (= keep `obj_dict["dtype"] = None` in this case) and replace `if "dtype" in obj_dict:` with the `if obj_dict["dtype"] is not None` line in `unpack()`.

If we do that for dtype, layout, etc., the code will become shorter and we will have a way to distinguish the None value from the absence of the field (may be useful if this class is extended in future).

borzunov (changes requested):
Thanks for this important contribution!

Let's write tests with running `MSGPackSerializer.dumps/loads` for `BatchTensorDescriptor`. Also, please consider the minor comment above.

hivemind/utils/tensor_descr.py
```diff
@@ -70,13 +83,48 @@ def from_tensor(cls, tensor: torch.Tensor, compression=CompressionType.NONE) ->
...
+            compression=compression if tensor.is_floating_point() else CompressionType.NONE,
         )
 
     def make_empty(self, *batch_size: int, **kwargs) -> torch.Tensor:
         assert self.shape[0] is None, "Make sure 0-th dimension is not specified (set to None)"
         return super().make_empty(size=(*batch_size, *self.shape[1:]), **kwargs)
 
+    def packb(self) -> bytes:
+        obj_dict = asdict(self)
+        obj_dict.pop("device")
```
borzunov, line 95: Let's move this line right below `if self.device is not None:` (currently, reading the code surprised me with removing the device, and I got the reasoning only when I finished reading this function).

borzunov (approved):

hivemind/utils/tensor_descr.py
```diff
@@ -70,13 +83,48 @@ def from_tensor(cls, tensor: torch.Tensor, compression=CompressionType.NONE) ->
...
+            obj_dict["dtype"] = _str_to_torch_type(obj_dict["dtype"], torch.dtype)
+
+        if obj_dict["layout"] is not None:
+            obj_dict["layout"] = _str_to_torch_type(obj_dict["layout"], torch.layout)
+
+        if obj_dict["device_type"] is not None:
+            obj_dict["device"] = torch.device(
+                obj_dict.pop("device_type"),
+                obj_dict.pop("device_index"),
+            )
```
borzunov, line 122:
```suggestion
        if obj_dict["device_type"] is not None:
            obj_dict["device"] = torch.device(obj_dict["device_type"], obj_dict["device_index"])
        else:
            obj_dict["device"] = None
        del obj_dict["device_type"], obj_dict["device_index"]
```

PR #389, opened 2021-09-25
Support different AMP & buffer configurations in one experiment, fix minor bugs

justheuristic:
New features:
- CollaborativeOptimizer can now combine fp16=True and reuse_grad_buffers=True with a special scaler
- CollaborativeOptimizer peers with reuse_grad_buffers=True and reuse_grad_buffers=False can now co-exist
- CollaborativeOptimizer peers with and without AMP can now co-exist

The new behavior of CollaborativeOptimizer with fp16 is:
* grad_scaler=None: regular fp32 behavior
* reuse_grad_buffers=False with GradScaler: works as usual, independently un-scales each tensor before accumulation, does not affect internal optimizer
* reuse_grad_buffers=True with GradScaler: when calling scaler.step(opt), it will raise error and complain that it requires HivemindGradScaler
* reuse_grad_buffers=False with HivemindGradScaler: applies unscale/update only around global optimizer step

### DIFF
--- a/hivemind/optim/collaborative.py
+++ b/hivemind/optim/collaborative.py
@@ -374,9 +374,7 @@ class CollaborativeOptimizer(DecentralizedOptimizerBase):
             return
 
         if self._grads is None:
-            self._grads = [
-                torch.zeros_like(grad, device=self.accumulate_grads_on) for grad in self._grad_buffers()
-            ]
+            self._grads = [torch.zeros_like(grad, device=self.accumulate_grads_on) for grad in self._grad_buffers()]
         yield from self._grads
 
     @torch.no_grad()
--- a/hivemind/optim/collaborative.py
+++ b/hivemind/optim/collaborative.py
@@ -99,7 +99,6 @@ class CollaborativeOptimizer(DecentralizedOptimizerBase):
     :note: If you are using CollaborativeOptimizer with lr_scheduler, it is recommended to pass this scheduler
       explicitly into this class. Otherwise, scheduler may not be synchronized between peers.
     """
-    _step_supports_amp_scaling = True  # pytorch amp support
 
     def __init__(
         self,
@@ -149,6 +148,8 @@ class CollaborativeOptimizer(DecentralizedOptimizerBase):
         self.client_mode, self.step_tolerance = client_mode, step_tolerance
         self.status_loglevel = logging.INFO if verbose else logging.DEBUG
         self.averager = self._make_averager(**kwargs)
+
+        self._step_supports_amp_scaling = self.reuse_grad_buffers  # enable custom execution with torch GradScaler
 
         self.training_progress_key = f"{self.prefix}_progress"
         self.local_samples_accumulated = 0  # a number of local samples accumulated since last optimizer update
@@ -218,7 +219,8 @@ class CollaborativeOptimizer(DecentralizedOptimizerBase):
         :param grad_scaler: if amp is enabled, this **must** be a hivemind-aware gradient scaler
         :note: this .step is different from normal pytorch optimizers in several key ways. See __init__ for details.
         """
-        assert grad_scaler is None or isinstance(grad_scaler, HivemindGradScaler)
+        if grad_scaler is not None and not isinstance(grad_scaler, HivemindGradScaler):
+            raise ValueError("CollaborativeOptimizer requires a hivemind-aware gradient scaler (HivemindGradScaler).")
         if self.batch_size_per_step is None:
             if batch_size is None:
                 raise ValueError("Please either set batch_size_per_step parameter at init or when calling .step")
@@ -373,7 +375,9 @@ class CollaborativeOptimizer(DecentralizedOptimizerBase):
             return
         else:
             if self._grads is None:
-                self._grads = [torch.zeros_like(grad, device=self.accumulate_grads_on) for grad in self._grad_buffers()]
+                self._grads = [
+                    torch.zeros_like(grad, device=self.accumulate_grads_on) for grad in self._grad_buffers()
+                ]
             yield from self._grads
 
     @torch.no_grad()
@@ -381,7 +385,7 @@ class CollaborativeOptimizer(DecentralizedOptimizerBase):
         """add current gradients to grad accumulators (if any)"""
         if self.reuse_grad_buffers:
             # user is responsible for accumulating gradients in .grad buffers
-            assert batch_size == self.batch_size_per_step, "Custom batch size is not implemented for reuse_grad_buffers"
+            assert batch_size == self.batch_size_per_step, "Custom batch size is not supported if reuse_grad_buffers"
         else:
             alpha = float(batch_size) / self.batch_size_per_step
             for grad_buf, grad_acc in zip(self._grad_buffers(), self.accumulated_grads()):
--- a/hivemind/optim/collaborative.py
+++ b/hivemind/optim/collaborative.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 from dataclasses import dataclass
 from threading import Event, Lock, Thread
-from typing import Dict, Iterator, Optional
+from typing import Dict, Iterator, Optional, Any, Callable
 
 import numpy as np
 import torch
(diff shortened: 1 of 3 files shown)


hivemind/optim/collaborative.py
```diff
@@ -344,38 +370,36 @@ def accumulated_grads(self) -> Iterator[torch.Tensor]:
...
+            # user is responsible for accumulating gradients in .grad buffers
+            assert batch_size == self.batch_size_per_step, "Custom batch size is not implemented for reuse_grad_buffers"
+        else:
+            alpha = float(batch_size) / self.batch_size_per_step
+            for grad_buf, grad_acc in zip(self._grad_buffers(), self.accumulated_grads()):
+                grad_acc.add_(grad_buf.to(grad_acc.device), alpha=alpha)
 
     @torch.no_grad()
     def apply_accumulated_grads_(self, scale_by: Optional[float] = None):
-        if self.reuse_grad_buffers:
```
justheuristic, line 365: This previously caused a bug where reuse=True peers would not be scaled by `scale_by`. As a result, they would have larger gradients and dominate the reuse=False peers.

hivemind/optim/collaborative.py
```diff
@@ -344,38 +370,36 @@ def accumulated_grads(self) -> Iterator[torch.Tensor]:
         """local gradient accumulators"""
         if self.reuse_grad_buffers:
             yield from self._grad_buffers()
-        elif self._grads is None:
-            with torch.no_grad():
-                self._grads = [
-                    torch.zeros_like(grad, device=self.accumulate_grads_on) for grad in self._grad_buffers()
-                ]
-        yield from self._grads
```
justheuristic, line 352: This wold actually an error with reuse_grad_buffers=True, but it worked because noone asked for more than len(grad_buffers) elements

hivemind/optim/collaborative.py
```diff
@@ -197,6 +200,8 @@ def load_state_from_peers(self, **kwargs):
                 try:
                     self.averager.load_state_from_peers(timeout=self.load_state_timeout, **kwargs)
                     break
+                except KeyboardInterrupt:
+                    raise
```
justheuristic, line 204: This would previously cause a deadlock if user would KeyboardInterrupt CollaborativeOptimizer during load_state_from_peers

hivemind/optim/grad_scaler.py
```diff
@@ -0,0 +1,76 @@
...
+        assert isinstance(optimizer, DecentralizedOptimizerBase)
+        if self._is_running_global_step:
+            super().unscale_(optimizer.opt)
+            return True
+        else:
+            self._check_inf_per_device(optimizer.opt)
+            self._optimizer_states_to_reset.add(id(optimizer))
+            return False
+
+    def step(self, optimizer, *args, **kwargs):
```
borzunov, line 41: nit:

```suggestion
    def step(self, optimizer: Optimizer, *args, **kwargs) -> bool:
```

hivemind/optim/grad_scaler.py
```diff
@@ -0,0 +1,76 @@
...
+        else:
+            for opt_id in self._optimizer_states_to_reset:
+                self._per_optimizer_states[opt_id] = _refresh_per_optimizer_state()
+            self._optimizer_states_to_reset.clear()
+            return False
+
+    def _unscale_grads_(
+        self, optimizer: Optimizer, inv_scale: torch.Tensor, found_inf: torch.Tensor, allow_fp16: bool
+    ) -> Dict[torch.device, torch.Tensor]:
+        return super()._unscale_grads_(optimizer, inv_scale, found_inf, allow_fp16=True)
```
borzunov, line 72: Why does it ignore the `allow_fp16` value always setting it to True?

hivemind/optim/collaborative.py
borzunov: Maybe replace `except BaseException` -> `except Exception` instead of this change?

hivemind/optim/collaborative.py
```diff
@@ -251,6 +266,10 @@ def step(self, batch_size: Optional[int] = None, **kwargs):
 
             # divide accumulators by local steps to recover the true average grad w.r.t. local_samples_accumulated
             self.apply_accumulated_grads_(scale_by=1.0 / self.local_updates_accumulated)
+            if grad_scaler:
```
borzunov, line 269: nit:
```suggestion
            if grad_scaler is not None:
```

hivemind/optim/collaborative.py
```diff
@@ -279,13 +298,21 @@ def step(self, batch_size: Optional[int] = None, **kwargs):
                     f"Skipped averaging: collaboration consists of " f"{self.collaboration_state.num_peers} peer(s).",
                 )
 
-            self.opt.step()
+            if grad_scaler:
```
borzunov, line 301: nit:
```suggestion
            if grad_scaler is not None:
```

hivemind/optim/collaborative.py
```diff
@@ -279,13 +298,21 @@ def step(self, batch_size: Optional[int] = None, **kwargs):
...
+            else:
+                self.opt.step()
+
             self.reset_accumulated_grads_()
             self.local_samples_accumulated = self.local_updates_accumulated = 0
             self.collaboration_state.register_step(current_step + 1)
             self.averager.local_step = current_step + 1
             self.collaboration_state_updated.set()
             self.update_scheduler()
+            if grad_scaler:
```
borzunov, line 313: nit:
```suggestion
            if grad_scaler is not None:
```

hivemind/optim/grad_scaler.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def are_grads_finite(self, optimizer: DecentralizedOptimizerBase):
```
borzunov, line 74:
```suggestion
    def are_grads_finite(self, optimizer: DecentralizedOptimizerBase) -> bool:
```

hivemind/optim/grad_scaler.py
```diff
@@ -0,0 +1,76 @@
...
+                super().step(optimizer.opt, *args, **kwargs)
+            else:
+                logger.warning("Skipping global step due to gradient over/underflow")
+            return True
+        else:
+            super().step(optimizer)
+            self._optimizer_states_to_reset.add(optimizer)
+            return False
+
+    def update(self, new_scale=None):
```
borzunov, line 54: nit:

```suggestion
    def update(self, new_scale: Optional[float] = None) -> bool:
```

borzunov (changes requested):

hivemind/optim/grad_scaler.py
justheuristic: that's the same trick that fairscale uses to allow training without master fp32 weights
https://github.com/facebookresearch/fairscale/blob/main/fairscale/optim/grad_scaler.py
(got referred there by @TimDettmers)

Added a quick comment explaining that

hivemind/optim/collaborative.py
```diff
@@ -344,38 +371,38 @@ def accumulated_grads(self) -> Iterator[torch.Tensor]:
         """local gradient accumulators"""
         if self.reuse_grad_buffers:
             yield from self._grad_buffers()
-        elif self._grads is None:
-            with torch.no_grad():
+            return
+        else:
```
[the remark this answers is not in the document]
justheuristic: I respectfully disagree here: it does reduce indentation, but:
- the code in the else clause is quite simple as it is
- crucially, last time the code was wrong because i've accidentally removed return and didn't think about the ramifications

If you do insist, please state that explicitly, i'll remove the else clause anyway

hivemind/optim/collaborative.py
justheuristic: Sometime in future, yes, but currently i'm afraid that it may accidentally forward CancelledError or something like that.

hivemind/optim/collaborative.py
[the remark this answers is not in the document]
borzunov: In this case, I do not insist :)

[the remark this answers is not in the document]
borzunov: I would say that `else:` after `return` is obviously redundant and seems like a bug, so I insist on choosing one of the two options:

1. `return` without `else:` (preferred since it implies less indentation, more clarity)
2. `else:` without `return`

hivemind/optim/collaborative.py
borzunov: Okay

borzunov (approved):

PR #398, opened 2021-10-28
Add an option to pre-schedule averaging

justheuristic:
This if the first of 3 pull-requests that aim to unify hivemind optimizers under one class:
```(python)
hivemind.Optimizer(
  params=model.parameters(), make_opt=partial(Adam, betas=(0.9, 0.95)),  # alternative: opt=my_optimizer_instance
  average_gradients=True, average_parameters=True, average_optimizer_statistics=['exp_avg_sq'],
  async_state_averaging=True, async_gradient_averaging=True, async_optimizer_step=True, offload_optimizer=True,
)
```

Part 1:
- [x] implement a way to pre-schedule averager step and add a trigger that manually enables averaging
- [x] add tests for pre-scheduling and averaging trigger

Part 2:
- [ ] create HivemindOptimizer that combines CollabroativeOptimizer, DecentralizedSGD, DecentralizedAdam, OffloadOptimizer, AdaptiveCollaborativeOptimizer functionalities as options
- [ ] HivemindGradScaler -> hivemind.GradScaler
- [ ] HivemindOptimizer -> hivemind.Optimizer
- [ ] deprecate hivemind.optim.*
- [ ] add a test for the new optimizer
- [ ] switch quickstart.html to the new optimizer

Part 3:
- [ ] implement averaging every k steps
- [ ] implement asynchronous gradient/state averaging
- [ ] implement offloaded optimizer
- [ ] implement EF21-style gradient averager

v1.1.0:
- [ ] remove the old optimizers


hivemind/averaging/matchmaking.py
```diff
@@ -78,8 +81,16 @@ def __init__(
...
-        self.potential_leaders = PotentialLeaders(self.peer_id, averaging_expiration, target_group_size)
-        self.data_for_gather: Optional[bytes] = None
+        self.potential_leaders = PotentialLeaders(self.peer_id, min_matchmaking_time, target_group_size)
+        self.step: Optional[StepControl] = None
+
+    @contextlib.asynccontextmanager
+    async def looking_for_group(self, step: StepControl):
+        async with self.lock_looking_for_group:
+            assert self.step is None
+            self.step = step
```
borzunov, line 91:
```suggestion
            self.step_control = step_control
```

I'd suggest to rename it to `step_control`, so it does not remind a step number.

hivemind/averaging/control.py
```diff
@@ -0,0 +1,133 @@
...
+    :returns: an assembled group if successful, None if failed; does NOT perform the actual averaging
+    """
+
+    def __init__(
+        self, scheduled_time: DHTExpiration, deadline: float, allow_retries: bool, weight: float, gather_binary: bytes
+    ):
+        super().__init__()
+        self._gather_binary, self._deadline, self._allow_retries = gather_binary, deadline, allow_retries
+        self._trigger: Optional[MPFuture] = None
+        self._shared_buffer = torch.zeros([18], dtype=torch.uint8).share_memory_()
```
borzunov, line 39:
```suggestion
        # Buffer contents: [scheduled_time: double | weight: double | stage: AveragingStage (1 byte) | began_allreduce: bool (1 byte)]
        self._shared_buffer = torch.zeros([18], dtype=torch.uint8).share_memory_()
```

hivemind/averaging/control.py
```diff
@@ -0,0 +1,133 @@
...
+        return self._deadline
+
+    def get_timeout(self) -> Optional[DHTExpiration]:
+        return max(0.0, self.deadline - get_dht_time())
+
+    @property
+    def allow_retries(self) -> bool:
+        return self._allow_retries
+
+    def __getstate__(self):
```
borzunov, line 117: Why do we need `__getstate__()` and `__setstate__()` here? They seem to do same things as automatic pickling/unpickling.

hivemind/optim/base.py
```diff
@@ -1,6 +1,9 @@
 import torch
 
 from hivemind.dht import DHT
+from hivemind.utils import get_logger
+
+logger = get_logger(__name__)
```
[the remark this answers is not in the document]
justheuristic: thanks! fixed it

tests/test_averaging.py
```diff
@@ -420,6 +422,52 @@ def test_getset_bits():
...
+        controls.append(
+            averager.step(
+                wait=False,
+                scheduled_time=hivemind.get_dht_time() + 1,
+                weight=1.0,
+                require_trigger=i in (1, 2),
+            )
+        )
+
+    time.sleep(1.1)
```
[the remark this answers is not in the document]
justheuristic: will do

hivemind/averaging/matchmaking.py
justheuristic: [applied]

hivemind/averaging/control.py
justheuristic: A: this ensures that the part of state that is stored in shared_buffer will be shared correctly in all edge cases (for the same reason as in MPFuture)

hivemind/averaging/averager.py
```diff
@@ -104,8 +106,9 @@ def __init__(
         target_group_size: int,
         min_group_size: int = 2,
         initial_group_bits: str = "",
-        averaging_expiration: float = 15,
-        request_timeout: float = 3,
+        averaging_expiration: float = None,
```
mryab, line 109:
```suggestion
        averaging_expiration: Optional[float] = None,
```

hivemind/averaging/averager.py
```diff
@@ -116,6 +119,7 @@ def __init__(
         min_vector_size: int = 0,
         auxiliary: bool = False,
         allow_state_sharing: Optional[bool] = None,
+        declare_state_period: float = 30,
```
mryab, line 122: It's seemingly not documented

hivemind/averaging/averager.py
```diff
@@ -316,68 +328,80 @@ def __del__(self):
...
+        :param wait: if True (default), return when finished. Otherwise return StepControl and run in background.
         :returns: on success, update averaged_tensors and return group info; on failure, return None
         """
         if self.mode == AveragingMode.AUX and weight is not None:
             logger.warning("Averager is running in auxiliary mode, weight is unused.")
+        if scheduled_time is None:
+            scheduled_time = get_dht_time() + self.matchmaking_kwargs["min_matchmaking_time"]
         if weight is None:
             weight = float(self.mode != AveragingMode.AUX)
+        deadline = get_dht_time() + timeout if timeout else float("inf")
```
mryab, line 358:
```suggestion
        deadline = get_dht_time() + timeout if timeout is not None else float("inf")
```

hivemind/averaging/averager.py
```diff
@@ -316,68 +328,80 @@ def __del__(self):
...
         Set up the averager to look for a group and run one round of averaging, return True on success, False on failure
 
         :param gather: optionally send this informaton to all peers in the next group and gather it from every groupmate
           (this operation is known as all-gather). The gathered data will be available as the output of this function.
+        :param scheduled_time: when matchmaking, assume that all-reduce will begin at this moment
         :param weight: averaging weight for this peer, int or float, must be strictly positive
         :param allow_retries: if averager fails to run one round of allreduce, this option will allow it to try again
           within the specified timeout
+        :param require_trigger: if True, await for user to call .allow_allreduce() before running all-reduce
         :param timeout: if averager was unable to *find* a group in this many seconds, consider allreduce failedK
```
mryab, line 348:
```suggestion
        :param timeout: if averager was unable to *find* a group in this many seconds, consider allreduce failed
```

hivemind/averaging/averager.py
```diff
@@ -316,68 +328,80 @@ def __del__(self):
...
-                    weight=weight,
-                    allow_retries=allow_retries,
-                    timeout=timeout,
-                ),
-            )
+        assert not (wait and require_trigger), "Non-asynchronous step cannot wait for trigger (use wait=False)"
+        assert scheduled_time < deadline, "Scheduled start time does not fit within timeout"
+
+        user_gather_bytes = self.serializer.dumps(gather)  # serialize here to avoid imports in the averager process
+        gather_binary = self.serializer.dumps([self.bandwidth, self.mode.value, user_gather_bytes])
```
mryab, line 364: I think naming the data to be gathered with a verb here is a bit misleading (it may be a function or at least a boolean flag, but hardly an object). What do you think about `serialized_data_to_gather` and `payload_to_gather`/`gather_data`, so that gather/average refers to an adjective instead of a verb?

mryab: (And also for the arguments/fields of StepControl)

hivemind/averaging/averager.py
```diff
@@ -316,68 +328,80 @@ def __del__(self):
...
-                        timeout=timeout, data_for_gather=data_for_gather
-                    )
+                    step.stage = AveragingStage.LOOKING_FOR_GROUP
+                    group_info = await self._matchmaking.look_for_group(step)
                     if group_info is None:
                         raise AllreduceException("Averaging step failed: could not find a group.")
 
-                    future.set_result(
+                    if not step.done():
+                        step.stage = AveragingStage.AWAITING_TRIGGER
```
mryab, line 396: In which case is it possible to have `step.done()` here? If it happens when the trigger is already called, will it be safe to switch the state anyway and then change it to  `RUNNING_ALLREDUCE` right away?

hivemind/averaging/averager.py
```diff
@@ -393,20 +417,20 @@ async def _step(
                     asyncio.InvalidStateError,
                     P2PHandlerError,
                 ) as e:
-                    time_elapsed = get_dht_time() - start_time
-                    if not allow_retries or (timeout is not None and timeout < time_elapsed):
-                        logger.exception(f"Averager caught {repr(e)}")
-                        future.set_exception(e)
+                    if not step.allow_retries or get_dht_time() >= step.deadline:
+                        logger.exception(f"{self.__class__.__name__} caught {repr(e)}")
```
mryab, line 421: I'm not sure if any of this is necessary, since we'll have a traceback with the name of the class and repr() of the exception anyway

hivemind/averaging/averager.py
```diff
@@ -517,19 +540,28 @@ async def _declare_for_download_periodically(self):
...
                         ),
-                        timeout=self._matchmaking.averaging_expiration,
+                        timeout=expiration_time - self.request_timeout,
                     )
                 )
-            await asyncio.sleep(self._matchmaking.averaging_expiration)
+            try:
+                await asyncio.wait_for(self._state_updated.wait(), self.declare_state_period - self.request_timeout)
+            except asyncio.TimeoutError:
+                pass
```
mryab, line 560: Should we issue a warning at least in the debug mode? Completely ignoring the exception is probably not the best strategy here

hivemind/averaging/averager.py
```diff
@@ -517,19 +540,28 @@ async def _declare_for_download_periodically(self):
...
+
+    def _mark_state_updated(self):
+        self.last_updated = get_dht_time()
+        self._state_updated.set()
```
mryab, line 564: If it's only used once, maybe it isn't necessary to extract these two lines to a function?

hivemind/averaging/control.py
```diff
@@ -0,0 +1,136 @@
...
+    :param scheduled_time: estimated time when averaging should begin. Will be used for scheduling
+    :param deadline: if averaging is still in progress at this time, it should be stopped due to TimeoutError
+    :param allow_retries: if True, allow running matchmaking and all-reduce again if previous attempt fails
+    :param weight: averaging weight, can be changed afterwards
+    :param gather_binary: optionally send this data to all peers in the next group and gather it from groupmates
+    :returns: an assembled group if successful, None if failed; does NOT perform the actual averaging
+    """
+
+    def __init__(
+        self, scheduled_time: DHTExpiration, deadline: float, allow_retries: bool, weight: float, gather_binary: bytes
```
mryab, line 34:
```suggestion
        self, scheduled_time: DHTExpiration, deadline: float, allow_retries: bool, weight: float, gather_binary: Optional[bytes]
```

hivemind/averaging/control.py
```diff
@@ -0,0 +1,136 @@
...
+        # Buffer contents:
+        # scheduled_time (double) | weight (double) | stage (AveragingStage, 1 byte) | began_allreduce: (bool, 1 byte)
+        self._shared_buffer = torch.zeros([18], dtype=torch.uint8).share_memory_()
+        self.stage = AveragingStage.IDLE
+        self.scheduled_time = scheduled_time
+        self.weight = weight
+        self.began_allreduce = False
+
+    def attach_trigger(self, trigger: MPFuture):
+        assert self._trigger is None, "trigger is already attached"
```
mryab, line 49:
```suggestion
        assert self._trigger is None, "Trigger is already attached"
```

hivemind/averaging/control.py
```diff
@@ -0,0 +1,136 @@
...
+        self._trigger = trigger
+
+    def allow_allreduce(self):
+        """Allow averager to begin allreduce when it finds a group. Meant to be triggered by user."""
+        assert self._trigger is not None, "StepControl does not have an attached trigger (not properly initialized)"
```
mryab, line 54:
```suggestion
        assert self._trigger is not None, "StepControl does not have an attached trigger"
```

mryab: The trigger is not attached during initialization, so it might mislead a user into thinking that they did something wrong at init time

hivemind/averaging/control.py
```diff
@@ -0,0 +1,136 @@
...
+        else:
+            self._trigger.set_result(None)
+
+    async def wait_for_trigger(self):
+        assert self._trigger is not None, "StepControl does not have an attached trigger (not properly initialized)"
+        await self._trigger
+
+    @property
+    def scheduled_time(self) -> DHTExpiration:
+        return struct.unpack("d", self._shared_buffer[0:8].numpy().data)[0]
```
mryab, line 66: Is it possible to just use four `mp.Value`s instead of playing around with byte representations? What are the constraints for _shared_buffer here that necessitate the shared-array part?

Furthermore, maybe pytorch isn't necessary and something like [SharedMemory](https://docs.python.org/3/library/multiprocessing.shared_memory.html#multiprocessing.shared_memory.SharedMemory) can help us do the same task with less levels of abstraction

mryab: If having this array is absolutely necessary, I'd create a dict/namedtuple of slices for its different parts to avoid indexing bugs

hivemind/averaging/matchmaking.py
```diff
@@ -78,8 +81,16 @@ def __init__(
 
         self.current_leader: Optional[PeerID] = None  # iff i am a follower, this is a link to my current leader
         self.current_followers: Dict[PeerID, averaging_pb2.JoinRequest] = {}  # my current followers excluding myself
-        self.potential_leaders = PotentialLeaders(self.peer_id, averaging_expiration, target_group_size)
-        self.data_for_gather: Optional[bytes] = None
+        self.potential_leaders = PotentialLeaders(self.peer_id, min_matchmaking_time, target_group_size)
+        self.step_control: Optional[StepControl] = None
+
+    @contextlib.asynccontextmanager
+    async def looking_for_group(self, step_control: StepControl):
```
mryab, line 88: Methods are generally verbs and participles are attributes/properties, is it safe to rename this to `enable_looking_for_group`?

mryab: And maybe if this method is used once, it's okay to just inline it (look_for_group is not that complicated)

hivemind/averaging/matchmaking.py
```diff
@@ -166,26 +175,25 @@ async def _request_join_potential_leaders(self, timeout: Optional[float]) -> Gro
                         self.assembled_group.set_exception(e)
                     raise e
 
-    async def request_join_group(self, leader: PeerID, expiration_time: DHTExpiration) -> Optional[GroupInfo]:
+    async def request_join_group(self, leader: PeerID) -> Optional[GroupInfo]:
```
mryab, line 178: Can we make this method a private one?

tests/test_averaging.py
```diff
@@ -420,6 +422,52 @@ def test_getset_bits():
...
+        )
+
+    time.sleep(0.6)
+
+    c0, c1, c2, c3 = controls
+    assert not any(c.done() for c in controls)
+    assert c0.stage == AveragingStage.RUNNING_ALLREDUCE
+    assert c1.stage == AveragingStage.AWAITING_TRIGGER
+    assert c2.stage == AveragingStage.AWAITING_TRIGGER
+    assert c3.stage == AveragingStage.RUNNING_ALLREDUCE
```
mryab, line 459: Can we also create a test that somehow checks that the matchmaking step is done in the background? Not necessary, but would be nice to have

mryab:
Thanks for the PR, especially for keeping the changelist short and focused!

BTW, don't forget to replace `[Optimizer overhaul 1/3] Add an option to pre-schedule averaging` with `Add an option to pre-schedule averaging` in the commit message, so that it fits the GH 50-character limit

docs/conf.py
```diff
@@ -203,7 +203,7 @@
 
 
 def setup(app):
-    app.add_stylesheet("fix_rtd.css")
```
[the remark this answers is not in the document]
justheuristic: RTFD would stop building without it -- sphinx has updated and this call is now deprecated

hivemind/averaging/averager.py
```diff
@@ -303,11 +315,11 @@ def shutdown(self) -> None:
         else:
             logger.exception("Averager shutdown has no effect: the process is already not alive")
 
-    async def _shutdown(self, timeout: Optional[float] = None) -> None:
+    async def _shutdown(self, timeout: Optional[float]) -> None:
```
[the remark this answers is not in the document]
justheuristic: It is, user can pass None to shutdown (without underscore) and it will be passed here

hivemind/averaging/matchmaking.py
```diff
@@ -499,13 +507,18 @@ async def _update_queue_periodically(self, key_manager: GroupKeyManager) -> None
             )
             self.update_triggered.clear()
 
-    async def _declare_averager_periodically(self, key_manager: GroupKeyManager) -> None:
+    async def _declare_averager_periodically(self, step: StepControl, key_manager: GroupKeyManager) -> None:
```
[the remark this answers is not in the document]
justheuristic: Not true, scheduled time can change between calls, passing it here freezes the time

hivemind/averaging/matchmaking.py
```diff
@@ -237,6 +245,14 @@ async def request_join_group(self, leader: PeerID, expiration_time: DHTExpiratio
             if stream is not None:
                 await stream.aclose()
 
+    def get_request_expiration_time(self) -> float:
```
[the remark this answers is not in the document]
justheuristic: It changes over time, therefore it is misleading to make it a property

hivemind/averaging/matchmaking.py
justheuristic: Not the case, this is a contextmanager.

with looking_for_group(...):

tests/test_averaging.py
justheuristic: Yes, in part2, please see the roadmap

hivemind/averaging/matchmaking.py
justheuristic: We can, but i'd rather not: it would cause dissonance with get_request_expiration_time, rpc_join_group, etc

hivemind/averaging/control.py
justheuristic: MP Value cannot be passed ahead without inheritance
shared memory is incompatible with py37
if you have any ideas, please apply them to MPFuture first

hivemind/averaging/averager.py
justheuristic: fixd

hivemind/averaging/averager.py
justheuristic: renamed everywhere to data_for_gather

hivemind/averaging/averager.py
justheuristic: that was a bug, i've intended to check if trigger was already enabled.

If so, it would be misleading report AWAITING_TRIGGER to user, even if momentarily

fixed now

hivemind/averaging/averager.py
justheuristic: fixd

hivemind/averaging/averager.py
justheuristic: fixd

hivemind/averaging/averager.py
justheuristic: Here, TimeoutError is a normal behavior.
It is the standard way to await for EITHER for event/future to be triggered or for a certain number of seconds to pass

hivemind/averaging/control.py
justheuristic: this argument is not optional: even if user does not supply data, it will be serialized as bytes alongside (bandwidth, averaging_mode) 

https://github.com/learning-at-home/hivemind/blob/pre_scheduling_again/hivemind/averaging/averager.py#L364-L365

mryab: Got it, the reason why I suggested it is that it's described as optional in the docstring

hivemind/averaging/control.py
mryab: What about four shared pytorch tensors then? My only concern is that stacking all values in a single array might be the easiest solution storage-wise, but more difficult to understand, maintain and wrap one's head around of

hivemind/averaging/matchmaking.py
mryab: Not sure about the dissonance: we have methods that are intended to be parts of the class public interface and those that are only meant to be used internally; the naming distinction simply helps reduce the interaction surface

hivemind/averaging/control.py
justheuristic: Four pytorch tensors is very expensive w.r.t. the number of open files, that's the same reason we have byte shifts in MPFuture

hivemind/averaging/matchmaking.py
```diff
@@ -499,13 +507,18 @@ async def _update_queue_periodically(self, key_manager: GroupKeyManager) -> None
...
                     await self.running.wait()
-
-                    new_expiration_time = min(get_dht_time() + self.averaging_expiration, self.search_end_time)
+                    new_expiration_time = float(
+                        np.clip(
+                            step.scheduled_time,
+                            a_min=get_dht_time() + self.min_matchmaking_time,
+                            a_max=self.search_end_time,
+                        )
+                    )
```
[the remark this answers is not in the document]
justheuristic: fixd

hivemind/averaging/control.py
justheuristic: fixd

hivemind/averaging/matchmaking.py
justheuristic: There are just as "private" (or more private) methods such as get_request_expiration_time rpc_join_group

We won't be able to rename rpc_join_group

hivemind/averaging/averager.py
```diff
@@ -129,6 +133,12 @@ def __init__(
         assert all(bit in "01" for bit in initial_group_bits)
         assert not client_mode or not auxiliary, "auxiliary peers must accept incoming connections"
 
+        if averaging_expiration is not None:
+            logger.warning("averaging_expiration is deprecated and will be removed soon, use min_matchmaking_time")
+            assert min_matchmaking_time == 5.0, "can't set both averaging_expiration and min_matchmaking_time"
```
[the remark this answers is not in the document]
justheuristic: i seemingly convinced you that it would be best to avoid global constants, lest we forget to remove them when the deprecation time comes

hivemind/averaging/control.py
```diff
@@ -0,0 +1,147 @@
...
+    def began_allreduce(self, value: bool):
+        self._shared_buffer[StepControl._BEGAN_ALLREDUCE] = int(value)
+
+    @property
+    def data_for_gather(self) -> bytes:
+        return self._data_for_gather
+
+    @property
+    def triggered(self) -> bool:
+        return self._trigger.done()
```
yhn112, line 118: I'm not sure if I like your `assert`s in general, but for consistency seems like we need `assert self._trigger is not None` here

yhn112 (changes requested):

hivemind/averaging/control.py
justheuristic: [fixd]

yhn112 (approved):

hivemind/averaging/averager.py
```diff
@@ -54,8 +55,7 @@ class DecentralizedAverager(mp.Process, ServicerBase):
     :param prefix: a shared prefix for all group keys
     :param target_group_size: attempts to form groups with up to this many peers (recommended: a power of 2, e.g. 16)
     :param initial_group_bits: a string of bits ('0' and '1') that define the initial group key (bucket index)
-    :param averaging_expiration: attempt to find a group for this many seconds, otherwise try again
-      note - this expiration time only applies to looking for group, passing tensors in allreduce may take more time
```
mryab, line 58: If it's deprecated, the `request_timeout` docstring should no longer reference it

hivemind/averaging/averager.py
```diff
@@ -393,20 +418,20 @@ async def _step(
...
-                    time_elapsed = get_dht_time() - start_time
-                    if not allow_retries or (timeout is not None and timeout < time_elapsed):
-                        logger.exception(f"Averager caught {repr(e)}")
-                        future.set_exception(e)
+                    if not step.allow_retries or get_dht_time() >= step.deadline:
+                        logger.exception(e)
+                        step.set_exception(e)
                     else:
-                        logger.warning(f"Averager caught {repr(e)}, retrying")
+                        logger.warning(f"{self.__class__.__name__} caught {repr(e)}, retrying")
```
mryab, line 425: Maybe use logger.exception here as well? (same as L422)

hivemind/averaging/control.py
```diff
@@ -0,0 +1,148 @@
...
+class StepControl(MPFuture):
+    """
+    An auxiliary data structure that allows user to control stages and track progress in a single averaging step
+
+    :param scheduled_time: estimated time when averaging should begin. Will be used for scheduling
+    :param deadline: if averaging is still in progress at this time, it should be stopped due to TimeoutError
+    :param allow_retries: if True, allow running matchmaking and all-reduce again if previous attempt fails
+    :param weight: averaging weight, can be changed afterwards
+    :param data_for_gather: send this data to all peers in the next group and gather it from groupmates
+    :returns: an assembled group if successful, None if failed; does NOT perform the actual averaging
```
mryab, line 30: It doesn't seem to return the group (and the class isn't callable, so it is unclear which method this refers to anyway)

hivemind/averaging/control.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+
+    _SCHEDULED_TIME, _WEIGHT, _STAGE, _BEGAN_ALLREDUCE = slice(0, 8), slice(8, 16), 16, 17
```
mryab, line 33: It'd be great to have a comment that these values denote indices for the shared buffer (right now, the names are nearly the same as the variables kept in the buffer). Also, I think it's safe to have them as module-level constants instead of class-level ones, but it's your call

hivemind/averaging/matchmaking.py
```diff
@@ -43,16 +46,16 @@ def __init__(
         prefix: str,
         target_group_size: int,
         min_group_size: int,
+        min_matchmaking_time: float,
         request_timeout: float,
         client_mode: bool,
         initial_group_bits: str = "",
-        averaging_expiration: float = 15,
```
mryab, line 49: At some point, it'd be great to document this class as well, but you can do this in future PRs

mryab (approved):

hivemind/averaging/averager.py
justheuristic: Here the important part is that it's retrying and ignoring the error

hivemind/averaging/control.py
justheuristic: good catch!

hivemind/averaging/matchmaking.py
justheuristic: yep, i've been wanting to make something like the DHT structure scheme, but for averaging and optimizers

hivemind/averaging/control.py
justheuristic: done

hivemind/averaging/averager.py
justheuristic: ya got it :)
image


id say it looks better off with the new name as well

PR #399, opened 2021-10-31
Move DHT to dht/dht.py, update DHT figure

justheuristic:
This is a chore PR, it **does not change any functionality**

- moved DHT to a separate sub-class https://github.com/learning-at-home/hivemind/issues/275
- updated DHT scheme (new paths, endpoint->peerID, 20bit -> 20byte)
- checked that benchmarks/exampels still work


hivemind/dht/routing.py
```diff
@@ -10,7 +10,7 @@
 from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, Union
 
 from hivemind.p2p import PeerID
-from hivemind.utils import MSGPackSerializer, get_dht_time
+from hivemind.utils import DHTExpiration, MSGPackSerializer, get_dht_time
```
borzunov, line 13: nit: DHTExpiration is not needed in `routing.py`. It would be better to remove this change and import `DHTExpiration` directly from `hivemind.utils` in `dht/__init__.py`.

borzunov (approved):

PR #400, opened 2021-11-01
[RFC] hivemind.Optimizer overhaul

justheuristic:
### Current experience with hivemind optimizers

Where it hurts:
- it should be clear which optimizer you should use (e.g. DecentralizedSGD vs CollaborativeOptimizer)
- optimizers should work fine with default parameters
- it hurts that we need to re-implement the same features for each optimizer (e.g. #252)
- implementing complex features (e.g. DPU) in CollaborativeOptimizer makes it into a huge blob of code that is difficult to maintain
- OffloadOptimizerWrapper (in sahajbert-xl) was painful to maintain and inefficient w.r.t. cpu-gpu communication

What worked:
- mimicry for PytorchOptimizer still seems to be a good idea
- tracking the global batch size in CollaborativeOptimizer works good enough for practical use
- wrapping arbitrary optimizer and scheduler proved convenient

Additional constraints:
- we need to finalize the interface by NeurIPS demo (not necessarily all functionality)
- we must be able to maintain features: PowerSGD/EF21, etc

### Proposal: interface

Replace the 4 existing optimizers with a single one
```(python)
hivemind.Optimizer(
  params=model.parameters(), optim_cls=partial(Adam, betas=(0.9, 0.95)),  # alternative: opt=my_optimizer_instance
  average_gradients=True, average_parameters=True, average_statistics=['exp_avg_sq'],
  async_state_averaging=True, async_gradient_averaging=True, async_optimizer_step=True, offload_optimizer=True,
)
```

The new optimizer would cover all 5 existing ones as special cases:
- DecentralizedSGD: `Optimizer(opt=SGD(...), average_parameters=True)`
- DecentralizedAdam: `Optimizer(opt=Adam(...), average_parameters=True, average_statistics=("exp_avg_sq",))`
- CollaborativeOptimizer: `Optimizer(opt=..., average_gradients=True, target_batch_size=4096, average_parameters=True)`
- CollaborativeAdaptiveOptimizer: `Optimizer(opt=..., average_gradients=True, target_batch_size=4096, average_parameters=True, average_statistics=("exp_avg_sq",))`
- DPUOptimizer: `Optimizer(..., average_gradients=True, async_gradient_averaging=True, **same_as_above)`

### Proposal: internals

img

### TODOs

- create hivemind.opt.Optimizer
   - [ ] test convergence rate for GPT2-like
   - [ ] add a CI test that checks the correctness after several optimizer steps
[...]

Issue #394, opened 2021-10-14
Delayed Parameter Update when step(wait=False)


bgyoon:
**Is your feature request related to a problem? Please describe.**

Eh, this could be a question. I'm trying to use `TrainingAverager` with `step(wait=False)`. That requires `data_lock` and `use_old_local_tensor=True` follows.

When `use_old_local_tensor=True`, is it correct to simply add the weight difference between local model and all-reduced model to the new model parameters? The gradients calculated from the old model parameter is being added to the new model parameters. That doesn't seem quite right.

**Describe the solution you'd like**

https://arxiv.org/abs/2101.06840 proposes Delayed Parameter Update. Parameter update is delayed by one step. Apparently, it makes little difference in the training curve if DPU is applied after 40 iterations in BERT-large training.

I think to implement DPU, you simply have to copy back the averaged tensor back to the model in the beginning of `step()`.

**Describe alternatives you've considered**

I understand that if the weight difference is not added back, the local steps taken before  the asynchronous all-reduce completes are being wasted. Not only it defeats purpose of asynchronous all-reduce(if local updates are going to be wasted until async completes, why not just go sync) but it also skips over input data which could trouble training.

justheuristic:
Hi!

What you're saying is perfectly correct, and with proper application, it will be equivalent to [2101.06840](https://arxiv.org/abs/2101.06840).

As you've said, ZeRO-offload is equivalent to at most 1 step of staleness. Therefore, there won't be any asynchronous **parameter** updates in the background for all-reduce. Instead, here's what should happen:

- first, peers accumulate gradients until they reach the target batch size
- then, IN PARALLEL
    - THREAD A - average currently accumulated gradients and run optimizer step
    - THREAD B - begin accumulating gradients for the next step - but do not run the next optimizer step yet

Therefore, we implemented DPU without using wait=False, but with two threads:
- thread A runs entirely on CPU, runs TrainingAverager and performs the optimizer steps  
- thread B uses GPU, accumulates gradients into a separate on-gpu buffer

These threads have only two points of interaction: A receives gradients from B before each all-reduce and sends back updated parameters after optimizer step. 

<details>
 <summary>More specifically, this (overly complicated) algorithm</summary>
- [ thread A, without GPU, with TrainingAverager ] 
  1. record local parameters & gradients from GPU as old_local_tensors
  2. run all_reduce over the network, get averaged gradients
[...]

PR #404, opened 2021-11-07
Add GradientAverager with support for delayed averaging

justheuristic:
This PR implements GradientAverager - a subclass of DecentralizedAverager that supports accumulating and aggregating gradients. This class supports pre-scheduling and delayed averaging (for DPU, #394 ) for use in hivemind.Optimizer ( #400 )


hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+      This is more memory efficient, but it requires that the user does *not* call zero_grad or clip_by_whatever at all
+    :param accumulate_grads_on: if specified, accumulate gradients on this device. By default, this will use the same
+     device as model parameters. One can specify a different device (e.g. 'cpu' vs 'cuda') to save device memory at
+     the cost of extra time per step. If reuse_grad_buffers is True, this parameter has no effect.
+    :param kwargs: see DecentralizedAverager keyword arguments for additional parameters
+
+
+    Example:
+
+    >>> model = SuchModelMuchLayers()
```
justheuristic, line 40: i insist

mryab: Alright, have it your way :P But shouldn't it be `ManyLayers` then?

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
+import contextlib
+from typing import Iterable, Iterator, Optional
+
+import torch
+
+import hivemind.dht.dht
```
mryab, line 6: Right now this import seems unused: it's better to directly `from hivemind.dht import DHT` if you need it to avoid import scope creep and side effects

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+class GradientAverager(DecentralizedAverager):
+    """
+    An auxiliary averaging class that is responsible for accumulating gradients and aggregating them with peers.
+    GradientAverager is meant to be used within hivemind.Optimizer, but it can be used standalone (see example below).
+
+    GradientAverager manages three sets of buffers:
+    (1) grads from model parameters - the gradients associated with local model parameters by pytorch.
+      - these tensors are typically stored in device memory
```
mryab, line 21:
```suggestion
    (1) gradients from model parameters - the gradients associated with local model parameters by PyTorch.
        These tensors are typically stored in device memory
```

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+    >>> grad_averager = GradientAverager(model.parameters(), dht=hivemind.DHT(...))
+    >>> next_step_time = hivemind.get_dht_time() + 60   # runs global steps every 60 seconds
+    >>> next_step_control = None
+    >>> while True:
+    >>>    # accumulate as many gradients as you can before next_step_time
+    >>>    loss = compute_loss(model, batch_size=32)
+    >>>    loss.backward()
+    >>>    grad_averager.accumulate_grads_(batch_size=32)
+    >>>    # [optional] next step in 5 seconds, start looking for peers in advance
+    >>>    if next_step_time - hivemind.get_dht_time() <= 5.
```
mryab, line 51:
```suggestion
    >>>    if next_step_time - hivemind.get_dht_time() <= 5
```

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+    def __init__(
+        self,
+        parameters: Iterable[torch.nn.Parameter],
+        *,
+        dht: hivemind.DHT,
+        prefix: str,
+        reuse_grad_buffers: bool = False,
+        accumulate_grads_on: Optional[torch.device] = None,
+        client_mode: bool = None,
+        warn: bool = True,
```
mryab, line 73: These two options are not documented. Also, do we actually need the `warn` option, which only appears once in the code?

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def _grads_from_parameters(self) -> Iterator[torch.Tensor]:
+        """gradient buffers associated with parameters"""
+        for param in self._parameters:
+            if param.grad is None:
+                param.grad = torch.zeros_like(param)
+            yield param.grad
+
+    @torch.no_grad()
+    def _grad_acumulators(self) -> Iterator[torch.Tensor]:
```
mryab, line 107:
```suggestion
    def _grad_accumulators(self) -> Iterator[torch.Tensor]:
```

mryab: Also, isn't this a property in disguise? If it's a method, we should rename it to something like `_get_gradient_accumulators`

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+        """averager-based gradient accumulators"""
+        assert (self._local_accumulators is None) == self.reuse_grad_buffers
+        yield from self._grads_from_parameters() if self.reuse_grad_buffers else self._local_accumulators
+
+    @torch.no_grad()
+    def accumulate_grads_(self, batch_size: int):
```
mryab, line 113: Perhaps the trailing underscore isn't necessary

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+        """add current gradients to local grad accumulators (if used)"""
+        if self._accumulators_used_in_step:
+            logger.warning(
+                "[warn=True] Gradient accumulators were not reset since the last averaging round. Please "
+                "call .reset_accumulated_grads_ after every step or use .step(reset_accumulators=True)."
+            )
+            self._accumulators_used_in_step = False  # warn once per round
+        if self._anchor_batch_size is None:
+            self._anchor_batch_size = batch_size
```
mryab, line 122: I'm not sure that you need this field in the class: it is only meaningfully used to compute `alpha` below, and since `batch_size` is not optional, you always know the actual batch size for the current gradients and can use simply this argument

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+    ) -> StepControl:
+        """
+        Begin matchmaking: look for a group of peers and prepare for averaging gradients at a specified time.
+
+        :param scheduled_time: expected time when to perform all-reduce. Can be changed using control.scheduled_time
+        :param gather: optionally send this informaton to all peers in the next group and gather it from every groupmate
+          (this operation is known as all-gather). The gathered data will be available as the output of this function.
+        :param allow_retries: if averager fails to run one round of allreduce, this option will allow it to try again
+          within the specified timeout
+        :param weight: setting weight at this stage is not supported, please leave this parameter as None
```
mryab, line 143: Can't we simply remove this argument or pass it as a note?

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+        self, scheduled_time: Optional[DHTExpiration] = None, weight: Optional[float] = None, **kwargs
+    ) -> StepControl:
+        """
+        Begin matchmaking: look for a group of peers and prepare for averaging gradients at a specified time.
+
+        :param scheduled_time: expected time when to perform all-reduce. Can be changed using control.scheduled_time
+        :param gather: optionally send this informaton to all peers in the next group and gather it from every groupmate
+          (this operation is known as all-gather). The gathered data will be available as the output of this function.
+        :param allow_retries: if averager fails to run one round of allreduce, this option will allow it to try again
+          within the specified timeout
```
mryab, line 142: These options aren't actually present in the list of method arguments. You can probably refer the user to kwargs of DecentralizedAverager to avoid duplication

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+        :param weight: by default, peer's weight is equal to the number of accumulated samples. Use this to override.
+        :param reset_accumulators: by default, set local gradient accumulators to zeros after averaging succeeds
+        :param control: reuse a pre-arranged group of peers (or a matchmaking in progress) from averager.schedule_step
+        :param wait: if True, await for the step to finish (or fail), otherwise run all-reduce in background
+        """
+        if control is None:
+            control = self.schedule_step(**kwargs)
+        else:
+            logger.warning(f"Averaging with a pre-scheduled group, parameters {kwargs} will have no effect.")
+        assert not control.triggered, f"this {type(control)} instance was already used."
```
mryab, line 170:
```suggestion
        assert not control.triggered, f"This {type(control)} instance was already used."
```

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+
+        control.weight = self.local_samples_accumulated
+        if reset_accumulators:
+            self.reset_accumulated_grads_()
+
+        control.allow_allreduce()
+        return control.result() if wait else control
+
+    @torch.no_grad()
+    def _load_accumulators_into_averager_(self):
```
mryab, line 183: Here, the trailing underscore is also unnecessary. If it signifies the inplace aspect of a function, I'd advise against it, because 1) the "load ... into ..." name already assumes that the changes happen inplace 2) this convention is not used in other parts of the repository

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,219 @@
...
+        control: Optional[StepControl] = None,
+        wait: bool = True,
+        **kwargs,
+    ):
+        """
+        Average accumulated gradients with peers, optionally load averaged gradients and reset accumulators
+
+        :param weight: by default, peer's weight is equal to the number of accumulated samples. Use this to override.
+        :param reset_accumulators: by default, set local gradient accumulators to zeros after averaging succeeds
+        :param control: reuse a pre-arranged group of peers (or a matchmaking in progress) from averager.schedule_step
```
mryab, line 163: The descriptions of arguments here are somewhat backwards now: they should first name the object/argument, then state what it's used for. Like "weight: the weight of this peer to be considered during gradient averaging, equal to the number of accumulated samples by default"

Also, the weight argument appears unused

tests/test_optimizer.py
```diff
@@ -0,0 +1,66 @@
+import multiprocessing as mp
```
mryab, line 1: Unused import

tests/test_optimizer.py
```diff
@@ -0,0 +1,66 @@
...
+
+import hivemind
+from hivemind.averaging.control import AveragingStage
+from hivemind.optim.experimental.grad_averager import GradientAverager
+
+
+@pytest.mark.forked
+def test_grad_averager():
+    dht1 = hivemind.DHT(start=True)
+    model1 = nn.ParameterDict(dict(w=nn.Parameter(torch.zeros(3))))
```
mryab, line 17: Maybe declare the dict in a regular manner here? It's even a bit more compact:
```suggestion
    model1 = nn.ParameterDict({"w": nn.Parameter(torch.zeros(3))})
```

tests/test_optimizer.py
```diff
@@ -0,0 +1,66 @@
...
+    model2 = nn.ParameterDict(dict(w=nn.Parameter(torch.zeros(3))))
+    averager2 = GradientAverager(
+        model2.parameters(), dht=dht2, prefix="test", target_group_size=2, reuse_grad_buffers=True, start=True
+    )
+
+    control1 = averager1.schedule_step(hivemind.get_dht_time() + 5)
+    control2 = averager2.schedule_step(hivemind.get_dht_time() + 5)
+
+    for i in range(10):
+        time.sleep(0.1)
```
mryab, line 32: This adds up to a second, can we reduce/remove it?

tests/test_optimizer.py
```diff
@@ -0,0 +1,66 @@
...
+    assert averager2.local_samples_accumulated == 18 and averager2.local_times_accumulated == 6
+    ref_grads2 = torch.full((3,), 2 * 1 / 3 * averager2.local_times_accumulated)
+    assert torch.allclose(next(averager2._grad_acumulators()), ref_grads2)
+
+    averager1.step(control=control1, wait=False)
+    averager2.step(control=control2, wait=False)
+    for step in (control1, control2):
+        step.result()  # wait for all-reduce to finish
+
+    ref_average = 8 / 26 * (ref_grads1 / 4) + 18 / 26 * (ref_grads2 / 6)
```
mryab, line 58: Instead of hardcoding the constants, it's better to compute them (8/18/26) in the code given the batch size per averager and the number of iterations. Basically, transform your comments from above :)

hivemind/optim/experimental/grad_averager.py
justheuristic: fixed

hivemind/optim/experimental/grad_averager.py
justheuristic: image

hivemind/optim/experimental/grad_averager.py
justheuristic: [added docstr for both]

@warn 
I missed one spot when it was needed, fixed now

hivemind/optim/experimental/grad_averager.py
justheuristic: It is, but the current behavior matches the one torch.nn.Module.parameters().

Willing to change that if you insist

hivemind/optim/experimental/grad_averager.py
justheuristic: This underscore signifies that the operation is in-place, resembling torch.add_ or clip_grad_norm_ and agrees with the rest of the optimizer

hivemind/optim/experimental/grad_averager.py
justheuristic: This argument is needed when accumulating several batches of different size, it will ensure that the batches are properly scaled when accumulated.

hivemind/optim/experimental/grad_averager.py
justheuristic: done

hivemind/optim/experimental/grad_averager.py
justheuristic: kk

hivemind/optim/experimental/grad_averager.py
justheuristic: As we agreed, im not changing the interface until we have the full optimizer

hivemind/optim/experimental/grad_averager.py
justheuristic: good catch :)

tests/test_optimizer.py
justheuristic: no, doing so would make the test flaky. The VM needs time to predictably run matchmaking

tests/test_optimizer.py
justheuristic: fixd

hivemind/optim/experimental/grad_averager.py
```diff
@@ -0,0 +1,218 @@
...
+                grad_acc.add_(grad_buf.to(grad_acc.device), alpha=alpha)
+
+    def schedule_step(self, scheduled_time: Optional[DHTExpiration] = None, **kwargs) -> StepControl:
+        """
+        Begin matchmaking: look for a group of peers and prepare for averaging gradients at a specified time.
+
+        :param scheduled_time: expected time when to perform all-reduce. Can be changed using control.scheduled_time
+        :param kwargs: any additional keyword args from DecentralizedAverager.step, such as gather, allow_retries, etc
+        :note: setting weight at this stage is not supported, please leave this parameter as None
+        :returns: a step_control - handle that can be passed into GradientAverager.step to use the pre-scheduled group
```
mryab, line 144:
```suggestion
        :returns: step_control - a handle that can be passed into GradientAverager.step to use the pre-scheduled group
```

mryab (approved):

hivemind/optim/experimental/grad_averager.py
```diff
@@ -165,8 +165,8 @@ def step(
         """
         if control is None:
             control = self.schedule_step(**kwargs)
-        else:
-            logger.warning(f"Averaging with a pre-scheduled group, parameters {kwargs} will have no effect.")
+        elif len(kwargs) > 0:
```
mryab, line 168: `elif kwargs:`

PR #405, opened 2021-11-07
move PerformanceEMA to utils, TrainingAverager to optim, update utils

justheuristic:
* implement and test async wrapper for ContextManager (used in DecentralizedAverager and ProgressTracker)
* implement .reset_timer in PerformanceEMA (used when progress was reset, e.g. with fp16 gradient overflow, which should not affect samples per second)
* move PerformanceEMA to hivemind.utils (rationale: will be used in hivemind.moe in @mryab 's pipelining exps)
* move TrainingAverager to hivemind.optim (for compliance with hivemind.Optimizer and future deprecation in favour of Training StateAverager)
* fix process-wide RSA keys in the validator


hivemind/utils/asyncio.py
```diff
@@ -147,3 +148,24 @@ async def attach_event_on_finished(iterable: AsyncIterable[T], event: asyncio.Ev
...
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        return self._context.__exit__(exc_type, exc_value, traceback)
+
+
+@asynccontextmanager
+async def enter_asynchronously(context: AbstractContextManager):
+    """Wrap a non-async context so that it can be entered asynchronously"""
+    async with _AsyncContextWrapper(context) as ret_value:
+        yield ret_value
```
justheuristic, line 171: note: we can't simply
```
try:
    yield await loop.run_in_executor(context.__enter__(...))
finally:
    context.__exit__(None, None, None)
```
b/c this option does not correctly propagate exceptions into the inner context manager

tests/conftest.py
```diff
@@ -33,6 +33,9 @@ def event_loop():
 def cleanup_children():
     yield
 
+    with RSAPrivateKey._process_wide_key_lock:
+        RSAPrivateKey._process_wide_key = None
```
mryab, line 37: Didn't see this change in the description and right now it seems to have no effect, do we need this?

hivemind/averaging/averager.py
```diff
@@ -453,7 +461,7 @@ async def _run_allreduce(self, group_info: GroupInfo, min_vector_size: int, **kw
                 None, load_balance_peers, self.total_size, download_bandwidths, min_vector_size
             )
 
-            async with self.get_tensors_async() as local_tensors:
+            async with enter_asynchronously(self.get_tensors()) as local_tensors:
```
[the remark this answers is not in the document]
justheuristic: It does, inside the get_tensors

hivemind/utils/performance_ema.py
```diff
@@ -37,15 +37,19 @@ def update(self, task_size: float, interval: Optional[float] = None) -> float:
         self.samples_per_second = 1 / max(adjusted_seconds_per_sample, self.eps)
         return self.samples_per_second
 
+    def reset_timer(self):
```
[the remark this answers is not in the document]
justheuristic: Yes, it will :)

tests/conftest.py
justheuristic: This has no effect, but it will if you run existing tests in different order.

TL;DR here is how it breaks things:
- you create **something** that instantiates RSAPrivateKey.process_wide
- and then all subsequent tests use the same key
- crucially, if you create two DHT instances with RSA validator **AFTER** instantiating private key, everything breaks because they both inherit your key

Future PRs introduce tests that instantiate validators before the sensitive tests and break everything in bizarre ways.

mryab (approved):

PR #411, opened 2021-11-17
Patch recursive cancel in StepControl

justheuristic:
Previously, cancelling StepControl would result in infinite recursion. This PR fixes this.

Log [discussion with @borzunov]
- Let's create a test case for this
- Test A: test averaging when one of the peers cancels during matchmaking
  - the non-cancelled peers should average successfully
- Test B: test averaging when one of the peers cancels during allreduce
  - test that no freezing occurs

### DIFF
--- a/hivemind/averaging/control.py
+++ b/hivemind/averaging/control.py
@@ -43,6 +43,7 @@ class StepControl(MPFuture):
         super().__init__()
         self._data_for_gather, self._deadline, self._allow_retries = data_for_gather, deadline, allow_retries
         self._trigger: Optional[MPFuture] = None
+        self._cancel: Optional[MPFuture] = None
 
         # Buffer contents:
         # scheduled_time (double) | weight (double) | stage (AveragingStage, 1 byte) | began_allreduce: (bool, 1 byte)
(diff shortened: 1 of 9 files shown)


borzunov (approved):

hivemind/averaging/control.py
```diff
@@ -52,13 +53,13 @@ def __init__(
...
-        self._trigger = trigger
+    def attach(self, trigger: MPFuture, cancel: MPFuture):
+        assert self._trigger is None and self._cancel is None, "Futures are already attached"
+        self._trigger, self._cancel = trigger, cancel
 
     def allow_allreduce(self):
-        """Allow averager to begin allreduce when it finds a group. Meant to be triggered by user."""
-        assert self._trigger is not None, "StepControl does not have an attached trigger"
+        """Allow averager to begin all-reduce when it finds a group. Meant to be triggered by user."""
+        assert self._trigger is not None and self._cancel is not None, "StepControl was not"
```
borzunov, line 62: The assert message is unfinished

borzunov (approved):

PR #417, opened 2021-12-01
Hotfix: load_state_from_peers with offload_optimizer

justheuristic:
hivemind.optim.experimental.Optimizer with offload_optimizer=True behaved incorrectly when loading state from peers.

It would load the state into local parameters, and then it was **meant to** write new parameters into the offloaded optimizer, but actually overriden the newly loaded parameters with old offloaded ones. The PR fixes this.

Demonstration 1: peers with uneven performance are now handled correctly (same as in CollaborativeOptimizer)
image

Demonstration 2: peers that join late are now handled correctly even with offload_optimizer=True. In the demo below, the **orange peer** started late.
image

Before fix, late peers with offload_optimizer would have correct main parameters, but their optimizer would actually have wrong parameters
image

### DIFF
--- a/hivemind/optim/experimental/state_averager.py
+++ b/hivemind/optim/experimental/state_averager.py
@@ -631,7 +631,8 @@ class TrainingStateAverager(DecentralizedAverager):
         Attempt to download the latest optimizer state from peers and update trainer parameters/statistics.
         :returns: whether or the averager succeeded in loading parameters
         """
-        main_parameters_and_extras = tuple(chain(self.main_parameters, self.extra_tensors))
+        opt_parameters = tuple(param for param_group in self.optimizer.param_groups for param in param_group["params"])
+        main_parameters_and_extras = tuple(chain(opt_parameters, self.extra_tensors))
         num_parameters_and_extras = len(main_parameters_and_extras)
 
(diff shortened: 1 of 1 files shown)


hivemind/optim/experimental/state_averager.py
```diff
@@ -631,7 +631,8 @@ def load_state_from_peers(self, **kwargs):
         Attempt to download the latest optimizer state from peers and update trainer parameters/statistics.
         :returns: whether or the averager succeeded in loading parameters
         """
-        main_parameters_and_extras = tuple(chain(self.main_parameters, self.extra_tensors))
+        opt_parameters = tuple(param for param_group in self.optimizer.param_groups for param in param_group["params"])
+        main_parameters_and_extras = tuple(chain(opt_parameters, self.extra_tensors))
```
borzunov, line 635: What does it mean? If `self.main_parameters` are invalid at this stage, shouldn't we replace them?

borzunov (approved):

justheuristic:
@SeanNaren JFYI: I see that you've reacted to this, but just in case, you will need to switch to newer hivemind version after this gets merged :)

PR #418, opened 2021-12-02
Do not log caller for INFO messages

borzunov:
The caller info for INFO log messages is often long but rarely useful. Such long lines confuse new users.

This PR hides the caller info for INFO messages **only**, unless a user specifies env variable `HIVEMIND_ALWAYS_LOG_CALLER=True` (`True` is case-insensitive).

I think it is important that this behavior **does not** depend on `HIVEMIND_LOGLEVEL`. Example: a user could create a Colab notebook parsing the logs and change the `HIVEMIND_LOGLEVEL` in the last minute, thus breaking the parser (since the INFO messages start to have the other format).

**Before:**

Screenshot 2021-12-02 at 13 22 53

**After:**

Screenshot 2021-12-02 at 13 22 30

### DIFF
--- a/hivemind/utils/logging.py
+++ b/hivemind/utils/logging.py
@@ -15,6 +15,9 @@ if _env_colors is not None:
 else:
     use_colors = sys.stderr.isatty()
 
+_env_log_caller = os.getenv("HIVEMIND_ALWAYS_LOG_CALLER")
+always_log_caller = _env_log_caller is not None and _env_log_caller.lower() == "true"
+
 
 class HandlerMode(Enum):
     NOWHERE = 0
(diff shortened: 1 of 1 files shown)


hivemind/optim/experimental/optimizer.py
```diff
@@ -461,7 +461,7 @@ def _update_global_epoch(self, grad_scaler: Optional[GradScaler]) -> None:
                 if not self.client_mode:
                     self.grad_averager.state_sharing_priority = self.local_epoch
 
-            logger.log(self.status_loglevel, f"Transitioning to epoch {self.local_epoch}.")
+            logger.debug(f"Transitioning to epoch {self.local_epoch}")
```
borzunov, line 464: This duplicates the `Step N` message (see the "Above" screenshot).

justheuristic: This message is useful because
- it signifies when a (potentially long) optimizer step is finished
- it is the only message that will be printed when use_local_updates=True (i.e. there is no global step)

justheuristic (approved):
Good job!

Please revert the epoch message though (or continue the discussion)

PR #419, opened 2021-12-02
Remove trailing dots in log messages and errors

borzunov:
This is important to make the output consistent.


justheuristic (approved):
Let's standardize

PR #422, opened 2021-12-11
Add minor tweaks learned from the NeurIPS demo run

justheuristic:
- if matchmaking encountered ControlFailure (dial failed), it would catch an error and restart completely. A more preferable behavior is to treat it as any other network errror
- awaiting for delayed updates to finish will no longer freeze forever if underlying update hanged
- rollback default max_refresh_timeout to 10 seconds

Previously, a dead peer would cause others to print errors incessently for 20-30 seconds trying to dial it
image

### DIFF
--- a/hivemind/optim/experimental/state_averager.py
+++ b/hivemind/optim/experimental/state_averager.py
@@ -376,17 +376,20 @@ class TrainingStateAverager(DecentralizedAverager):
         if wait_for_delayed_updates:
             for pending_update in self.pending_updates:
                 try:
+                    timeout = (averaging_opts or {}).get("averaging_timeout", self._allreduce_timeout)
                     logger.log(self.status_loglevel, "Waiting for delayed updates to finish...")
-                    output = pending_update.result()
+                    output = pending_update.result(timeout)
                 except BaseException:
-                    pass  # exception will be reported below
+                    # exception will be reported below
+                    if not pending_update.done():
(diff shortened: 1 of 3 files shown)


hivemind/optim/experimental/state_averager.py
```diff
@@ -376,17 +376,20 @@ def step(
...
+                        pending_update.cancel()
 
         # remove finished updates, log any exceptions
         finished_updates = {pending_update for pending_update in self.pending_updates if pending_update.done()}
         self.pending_updates = {pending_update for pending_update in self.pending_updates if not pending_update.done()}
         for finished_update in finished_updates:
-            if finished_update.exception():
-                logger.log(self.status_loglevel, f"Background update failed with {finished_update.exception()}")
+            if finished_update.cancelled() or finished_update.exception():
+                logger.log(self.status_loglevel, f"Background update failed: {finished_update}")
```
borzunov, line 392: Will the traceback for that be posted somewhere?

borzunov (approved):

hivemind/optim/experimental/state_averager.py
justheuristic: Depending on what happened, it will be printed by state_averager or ignored (e.g. if we cancel, it will just be ignored)

borzunov: Okay.

PR #423, opened 2021-12-12
Improve All-Reduce fault-tolerance

justheuristic:
- [x] allow AllreduceRunner to tolerate clients that
   - [x] do not send some of their local tensors
   - [x] do not show up at at all after matchmaking is over
- [x] allow AllreduceRunner to tolerate full/aux peers that do not send some or all results
- [x] introduce timeout after which sender/reducer is considered failed
- [x] AllreduceRunner & DecentralizedAverager will no longer _send_error_to_peer
   - log spam is gone!
- report allreduce integrity
   - [x] TensorPartReducer will report the fraction of expected parts received if that fraction is not 1
   - [x] TensorPartContainer will report the fraction of parts that did not fail if that fraction is not 1
- miscellaneous improvements to Optimizer
  - [x] set good default sender/reducer timeouts
  - [x] pre-schedule state averaging ahead of time
  - [x] no longer block the entire peer if it is time to pre-schedule gradients but background state averaging is still underway

Test cases:
- [x] test with peers that fail early
- [x] test with peers that fail to send a certain part
- [x] test with peers that fail to reduce their part
- [x] test cancelling

Sanity checks:
- [x] run tests 100 times
- [x] benchmark_optimizer
- [x] test env 64+ nodes 4+ hours

### DIFF
--- a/hivemind/averaging/allreduce.py
+++ b/hivemind/averaging/allreduce.py
@@ -1,6 +1,6 @@
 import asyncio
 from enum import Enum
-from typing import Any, AsyncIterator, Dict, Optional, Sequence, Tuple, Type
+from typing import Any, AsyncIterator, Dict, Optional, Sequence, Tuple, Type, Set
 
 import torch
 
@@ -16,7 +16,7 @@ from hivemind.utils.asyncio import (
     amap_in_executor,
     anext,
     as_aiter,
-    attach_event_on_finished,
+    attach_event_on_finished, aiter_with_timeout,
 )
 
 # flavour types
@@ -71,6 +71,7 @@ class AllReduceRunner(ServicerBase):
         peer_fractions: Tuple[float, ...],
         modes: Optional[Sequence[AveragingMode]] = None,
         gathered: Optional[Dict[PeerID, Any]] = None,
+        next_chunk_timeout: Optional[float] = None,
         **kwargs,
     ):
         self._p2p = p2p
@@ -102,6 +103,11 @@ class AllReduceRunner(ServicerBase):
             if mode != AveragingMode.AUX:
                 self.sender_peer_ids.append(peer_id)
 
+        self.next_chunk_timeout = next_chunk_timeout
+        self.active_senders: Set[PeerID] = {self.peer_id}  # peers that began sending data via rpc_aggregate_part
+        self.banned_senders: Set[PeerID] = set()  # peers that did not send data by next_chunk_timeout
+        self.banlock = asyncio.Lock()
+
         peer_id_index = self.ordered_peer_ids.index(self.peer_id)
         self.tensor_part_container = TensorPartContainer(tensors, peer_fractions, **kwargs)
         self.parts_for_local_averaging = self.tensor_part_container.get_raw_input_parts(peer_id_index)
@@ -132,6 +138,8 @@ class AllReduceRunner(ServicerBase):
     async def run(self) -> AsyncIterator[torch.Tensor]:
         """Run all-reduce, return differences between averaged and original tensors as they are computed"""
         pending_tasks = set()
+        if self.next_chunk_timeout is not None:
+            asyncio.create_task(self._handle_missing_senders())
         try:
             if len(self.sender_peer_ids) == 0:
                 logger.debug(f"{self} - finished all-reduce early: all peers are auxiliaries ({self.modes})")
@@ -151,10 +159,23 @@ class AllReduceRunner(ServicerBase):
                 self.finalize()
 
         except BaseException as e:
-            self.finalize(exception=e)
             for task in pending_tasks:
                 task.cancel()
+                if task.done() and not task.cancelled():
+                    logger.debug(f"Task {task} failed with {task.exception()}.")
+            self.finalize(exception=e)
             raise
+
+    async def _handle_missing_senders(self):
+        """Detect senders that should have sent tensors for averaging, but did not send anything within timeout"""
+        assert self.next_chunk_timeout is not None
+        await asyncio.sleep(self.next_chunk_timeout)
+        async with self.banlock:
+            for peer_id in self.sender_peer_ids:
+                if peer_id not in self.active_senders:
+                    sender_index = self.sender_peer_ids.index(peer_id)
+                    self.banned_senders.add(peer_id)
+                    self.tensor_part_reducer.on_sender_failed(sender_index)
 
     async def _communicate_with_peer(self, peer_id: PeerID):
         """Send a part of local tensors and metadata to a single peer, receive the average for that part of tensors"""
@@ -168,25 +189,38 @@ class AllReduceRunner(ServicerBase):
                 self.tensor_part_container.register_processed_part(peer_index, part_index, averaged_part - tensor_part)
 
         else:
-            code = None
-            stream = await self._get_peer_stub(peer_id).rpc_aggregate_part(self._generate_input_for_peer(peer_index))
-            async for part_index, (averaged_part_delta, msg) in aenumerate(
-                amap_in_executor(
-                    lambda msg: (deserialize_torch_tensor(msg.tensor_part), msg),
-                    stream,
-                    max_prefetch=self.tensor_part_container.prefetch,
-                )
-            ):
-                if code is None:
-                    code = msg.code
-                self.tensor_part_container.register_processed_part(peer_index, part_index, averaged_part_delta)
+            try:
+                code = None
+                done_sending = asyncio.Event()
+                inputs_aiter = attach_event_on_finished(self._generate_input_for_peer(peer_index), done_sending)
+                stream = await self._get_peer_stub(peer_id).rpc_aggregate_part(inputs_aiter)
 
-            if code != averaging_pb2.AVERAGED_PART:
-                raise AllreduceException(
-                    f"peer {peer_id} returned {averaging_pb2.MessageCode.Name(code)} "
-                    f"instead of {averaging_pb2.MessageCode.Name(averaging_pb2.AVERAGED_PART)}"
-                    f", allreduce failed"
-                )
+                if self.should_delay_results(self.peer_id):
+                    await done_sending.wait()
+                if self.next_chunk_timeout is not None:
+                    stream = aiter_with_timeout(stream, self.next_chunk_timeout)
+
+                async for part_index, (averaged_part_delta, msg) in aenumerate(
+                    amap_in_executor(
+                        lambda msg: (deserialize_torch_tensor(msg.tensor_part), msg),
+                        stream,
+                        max_prefetch=self.tensor_part_container.prefetch,
+                    )
+                ):
+                    if code is None:
+                        code = msg.code
+                    self.tensor_part_container.register_processed_part(peer_index, part_index, averaged_part_delta)
+                if code != averaging_pb2.AVERAGED_PART:
+                    raise AllreduceException(
+                        f"peer {peer_id} returned {averaging_pb2.MessageCode.Name(code)} "
+                        f"instead of {averaging_pb2.MessageCode.Name(averaging_pb2.AVERAGED_PART)}"
+                        f", allreduce failed"
+                    )
+
+            except BaseException as e:
+                logger.warning(f"Caught {repr(e)} when communicating to {peer_id}")
+                self.tensor_part_container.register_failed_reducer(peer_index)
+                raise
 
     async def _generate_input_for_peer(self, peer_index: int) -> AsyncIterator[averaging_pb2.AveragingData]:
         parts_aiter = self.tensor_part_container.iterate_input_parts_for(peer_index)
@@ -205,15 +239,15 @@ class AllReduceRunner(ServicerBase):
     ) -> AsyncIterator[averaging_pb2.AveragingData]:
         """a peer sends us a part of his tensor; we should average it with other peers and return the difference"""
         request: averaging_pb2.AveragingData = await anext(stream)
-        reason_to_reject = self._check_reasons_to_reject(request)
+        reason_to_reject = self._check_reasons_to_reject(request, context)
         if reason_to_reject:
             yield reason_to_reject
             return
 
         elif request.code == averaging_pb2.PART_FOR_AVERAGING:
+            sender_index = self.sender_peer_ids.index(context.remote_id)
             try:
-                sender_index = self.sender_peer_ids.index(context.remote_id)
-
+                self.active_senders.add(context.remote_id)
                 if not self.should_delay_results(context.remote_id):
                     async for msg in self._accumulate_parts_streaming(achain(as_aiter(request), stream), sender_index):
                         yield msg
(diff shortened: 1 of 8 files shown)


hivemind/averaging/allreduce.py
```diff
@@ -132,6 +148,8 @@ def should_delay_results(self, peer_id: PeerID) -> bool:
     async def run(self) -> AsyncIterator[torch.Tensor]:
         """Run all-reduce, return differences between averaged and original tensors as they are computed"""
         pending_tasks = set()
+        if self.sender_timeout is not None:
+            asyncio.create_task(self._handle_missing_senders())
```
borzunov, line 152: Such fire-and-forget calls always lead to "Task destroyed but is pending". Please fix that, e.g., save the task to a class field and `await` when finalizing.

hivemind/averaging/allreduce.py
```diff
@@ -205,28 +262,32 @@ async def rpc_aggregate_part(
     ) -> AsyncIterator[averaging_pb2.AveragingData]:
         """a peer sends us a part of his tensor; we should average it with other peers and return the difference"""
         request: averaging_pb2.AveragingData = await anext(stream)
```
borzunov, line 264: Please handle `self.sender_timeout` for this first part as well.

hivemind/averaging/allreduce.py
```diff
@@ -240,62 +301,62 @@ async def _accumulate_parts():
...
         else:
             error_code = averaging_pb2.MessageCode.Name(request.code)
-            logger.debug(f"{self} - peer {context.remote_id} sent {error_code}, allreduce cannot continue")
-            self.finalize(exception=AllreduceException(f"Peer {context.remote_id} sent {error_code}"))
+            logger.debug(f"{self} - peer {context.remote_id} sent {error_code}")
+            sender_index = self.sender_peer_ids.index(context.remote_id)
+            async with self.banlock:
+                if context.remote_id not in self.banned_senders:
+                    self.banned_senders.add(context.remote_id)
+                    self.tensor_part_reducer.on_sender_failed(sender_index)
```
borzunov, line 317: Please fix this code duplication (L305-308, L314-317, L352-355), this is likely to lead to a bug if someone decides to change the ban procedure (but changes only 1-2 of 3 copies).

borzunov (changes requested):
Thanks for the PR! I've went through everything but tests and left my comments.

hivemind/utils/asyncio.py
```diff
@@ -114,23 +114,30 @@ async def amap_in_executor(
...
+                raise future_or_exception
+            yield await future_or_exception
+            future_or_exception = await queue.get()
         await task
     finally:
-        if not task.done():
-            task.cancel()
+        task.cancel()
+        if task.done() and not task.cancelled():
+            task.exception()
```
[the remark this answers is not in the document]
justheuristic: I specifically mean "if task did not send result or throw exception but we died anyway, silence the "task ... was never retrieved""

hivemind/utils/asyncio.py
```diff
@@ -114,23 +114,30 @@ async def amap_in_executor(
...
-        future = await queue.get()
-        while future is not None:
-            yield await future
-            future = await queue.get()
+        future_or_exception = await queue.get()
+        while future_or_exception is not None:
+            if isinstance(future_or_exception, BaseException):
+                raise future_or_exception
+            yield await future_or_exception
+            future_or_exception = await queue.get()
```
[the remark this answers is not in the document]
justheuristic: Previously, if input iterator threw an exception, this code would freeze forever

hivemind/optim/experimental/optimizer.py
```diff
@@ -427,6 +432,8 @@ def _update_global_epoch(self, grad_scaler: Optional[GradScaler]) -> None:
 
             if self.use_gradient_averaging:
                 logger.log(self.status_loglevel, f"Beginning optimizer step #{self.local_epoch}")
+                self.state_averager.step(wait_for_delayed_updates=True)
```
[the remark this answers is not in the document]
justheuristic: If previous optimizer step is still running, we wait to avoid modifying its gradients in-place while it may still need them

hivemind/averaging/partition.py
```diff
@@ -209,23 +219,38 @@ async def accumulate_part(
...
+                self.reset_accumulators()
+
     def finalize(self):
         if not self.finished.is_set():
             if hasattr(self, "current_part_future"):
                 self.current_part_future.cancel()
                 del self.accumulator
             self.finished.set()
 
+            if self.num_parts != 0 and self.num_senders != 0:
```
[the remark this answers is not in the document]
justheuristic: it protects from zero division error

hivemind/averaging/allreduce.py
justheuristic: GREAT catch: i've meant to add it to pending_tasks (and i just did so)

hivemind/averaging/allreduce.py
justheuristic: replaced with a unified `await self._ban_sender(peer_id)` call

hivemind/averaging/partition.py
```diff
@@ -91,21 +92,16 @@ def get_raw_input_parts(self, peer_index: int) -> Tuple[torch.Tensor, ...]:
...
 
     @torch.no_grad()
     async def iterate_input_parts_for(self, peer_index: int) -> AsyncIterator[runtime_pb2.Tensor]:
         """iterate serialized tensor parts for a peer at a given index. Run serialization in background."""
         assert not self._inputs_consumed_by_peer[peer_index], "input parts of a given peer are already deallocated."
         self._inputs_consumed_by_peer[peer_index] = True
-
-        async def _aiterate_parts():
-            for _ in range(self.num_parts_by_peer[peer_index]):
-                yield self._input_parts_by_peer[peer_index].popleft()
```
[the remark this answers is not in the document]
justheuristic: These are all links that reuse tensors' memory, we fixed it a couple of months ago

The additional memory is only in the metadata

hivemind/averaging/partition.py
```diff
@@ -155,8 +160,6 @@ def finalize(self):
         if not self.finished.is_set():
             for peer_index in range(self.group_size):
                 self._inputs_consumed_by_peer[peer_index] = True
-                self._input_parts_by_peer[peer_index].clear()
-                self._output_parts_by_peer[peer_index].clear()
```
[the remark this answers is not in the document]
justheuristic: done. will introduce 16 such peers in test env and test if they're fine

hivemind/averaging/partition.py
```diff
@@ -178,18 +181,24 @@ def __init__(self, part_shapes: Sequence[torch.Size], num_senders: int):
         self.denominator = 0.0  # total weight accumulated from all peers for current part
         self.current_part_future = asyncio.Future()
         self.finished = asyncio.Event()
+
+        self.num_parts_received = [0 for _ in range(self.num_senders)]
+        self.sender_failed_after = [float("inf") for _ in range(self.num_senders)]
+        self.current_senders = self.num_senders
```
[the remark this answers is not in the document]
justheuristic: done

hivemind/averaging/allreduce.py
justheuristic: done

hivemind/averaging/partition.py
```diff
@@ -209,23 +219,38 @@ async def accumulate_part(
...
         return await current_part_future
 
+    def on_sender_failed(self, sender_index: int):
+        """Exclude that sender's data for averaging any parts that it did not submit yet."""
+        self.sender_failed_after[sender_index] = self.num_parts_received[sender_index]
+        if self.current_part_index == self.num_parts_received[sender_index]:
+            self.current_senders -= 1
+            if self.current_part_accumulated_from == self.current_senders:
+                self.current_part_future.set_result(self.accumulator.div_(self.denominator))
+                self.reset_accumulators()
```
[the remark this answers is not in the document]
justheuristic: done

hivemind/averaging/partition.py
```diff
@@ -178,18 +183,25 @@ def __init__(self, part_shapes: Sequence[torch.Size], num_senders: int):
         self.denominator = 0.0  # total weight accumulated from all peers for current part
         self.current_part_future = asyncio.Future()
         self.finished = asyncio.Event()
+
+        self.num_parts_received = [0 for _ in range(self.num_senders)]
+        self.sender_failed_after = [float("inf") for _ in range(self.num_senders)]
+        self.\
+            num_current_senders = self.num_senders
```
borzunov:
```suggestion
        self.num_current_senders = self.num_senders
```

justheuristic: thx

hivemind/utils/asyncio.py
```diff
@@ -114,23 +114,30 @@ async def amap_in_executor(
...
     async def _put_items():
-        async for args in azip(*iterables):
-            await queue.put(loop.run_in_executor(executor, func, *args))
-        await queue.put(None)
+        try:
+            async for args in azip(*iterables):
+                await queue.put(loop.run_in_executor(executor, func, *args))
+            await queue.put(None)
+        except BaseException as e:
+            await queue.put(e)  # note: there is no chance that iterables
```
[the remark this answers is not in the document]
justheuristic: fixd

hivemind/utils/asyncio.py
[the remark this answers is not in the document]
borzunov: Let's remove L136-137 because `.cancel()` already suppresses this message.

Indeed, the message is printed only when `task.__log_traceback == True` ([source](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/asyncio/futures.py#L94)) and `task.cancel()` sets it to `False` ([source](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/asyncio/futures.py#L151)), exactly as `task.exception()` does ([source](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/asyncio/futures.py#L217)).

hivemind/utils/asyncio.py
```diff
@@ -114,23 +114,30 @@ async def amap_in_executor(
...
     async def _put_items():
-        async for args in azip(*iterables):
-            await queue.put(loop.run_in_executor(executor, func, *args))
-        await queue.put(None)
+        try:
+            async for args in azip(*iterables):
+                await queue.put(loop.run_in_executor(executor, func, *args))
+            await queue.put(None)
+        except BaseException as e:
+            await queue.put(e)
```
borzunov, line 122: Suggestion: let's do it like this and revert the change below. This will remove this edge case as well as unnecessary `isinstance()` check.

```suggestion
            future = asyncio.Future()
            future.set_exception(e)
            await queue.put(future)
```

tests/test_allreduce_fault_tolerance.py
```diff
@@ -0,0 +1,188 @@
...
+from hivemind.proto import averaging_pb2
+from hivemind.utils.asyncio import as_aiter, azip, enter_asynchronously
+from hivemind.utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+class Fault(Enum):
+    NONE = auto()
+    FAIL_BEFORE = auto()
```
borzunov, line 25:
```suggestion
    FAIL_AWOL = auto()
```

[Cambridge Dictionary](https://dictionary.cambridge.org/dictionary/english/awol)

tests/test_allreduce_fault_tolerance.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class FaultyAllReduceRunner(AllReduceRunner):
+    def __init__(self, *args, fault: Fault, **kwargs):
+        self.fault = fault
+        super().__init__(*args, **kwargs)
+
+    async def rpc_aggregate_part(self, stream, context) -> AsyncIterator[averaging_pb2.AveragingData]:
+        if self.fault == Fault.FAIL_REDUCING:
+            yield averaging_pb2.AveragingData(code=averaging_pb2.INTERNAL_ERROR)
```
borzunov, line 91: Can we also test the case when `INTERNAL_ERROR` and `CANCEL` are the 2nd message? Should be useful since the 1st and all other messages are processed separately.

tests/test_allreduce_fault_tolerance.py
```diff
@@ -0,0 +1,188 @@
...
+    [
+        (Fault.NONE, Fault.FAIL_BEFORE),
+        (Fault.FAIL_BEFORE, Fault.FAIL_BEFORE),
+        (Fault.FAIL_SENDING, Fault.FAIL_SENDING),
+        (Fault.FAIL_SENDING, Fault.FAIL_BEFORE),
+        (Fault.FAIL_SENDING, Fault.FAIL_REDUCING),
+        (Fault.NONE, Fault.CANCEL),
+    ],
+)
+def test_fault_tolerance(fault0: Fault.NONE, fault1: Fault.NONE):
```
borzunov, line 128:
```suggestion
def test_fault_tolerance(fault0: Fault, fault1: Fault):
```

tests/test_allreduce_fault_tolerance.py
```diff
@@ -0,0 +1,188 @@
...
+
+    dht = hivemind.DHT(start=True)
+
+    averagers = []
+    for i in range(5):
+        averager = FaultyAverager(
+            _make_tensors(),
+            hivemind.DHT(initial_peers=dht.get_visible_maddrs(), start=True),
+            prefix="test",
+            request_timeout=0.3,
```
borzunov, line 140: Can we also test timeouts?

tests/test_allreduce_fault_tolerance.py
```diff
@@ -0,0 +1,188 @@
...
+
+        if all(fault == Fault.FAIL_SENDING for fault in (fault0, fault1)):
+            assert fault0 != Fault.FAIL_REDUCING and fault1 != Fault.FAIL_REDUCING
+            assert abs(diff[: len(diff) // 2]).max() < 1e-5
+        elif any(fault == Fault.CANCEL for fault in (fault0, fault1)):
+            pass  # late cancel may result in an arbitrary mix of averaging results with and without the cancelled peer
+        elif fault0 == Fault.NONE:  # only peer1 in client mode may have failed
+            assert abs(diff).max() < 1e-5
+        else:
+            assert (abs(diff) < 1e-5).numpy().mean() > 0.5
```
borzunov, line 185: Can we test the expected behavior on `FAIL_REDUCING`? It would be to be ensure that the averager uses a correct local part in this case (not some another part that will break the model completely).

justheuristic: added that case

hivemind/averaging/allreduce.py
```diff
@@ -168,25 +195,51 @@ async def _communicate_with_peer(self, peer_id: PeerID):
...
+
+                part_index = 0
+
+                def _try_deserialize(msg):
+                    if msg.code != averaging_pb2.AVERAGED_PART:
+                        return AllreduceException(f"{peer_id} sent {averaging_pb2.MessageCode.Name(msg.code)}"), msg
+                    try:
+                        return deserialize_torch_tensor(msg.tensor_part), msg
+                    except Exception as e:
+                        return e, msg
```
borzunov, line 215:
```suggestion
                    if msg.code != averaging_pb2.AVERAGED_PART:
                        raise AllreduceException(f"{peer_id} sent {averaging_pb2.MessageCode.Name(msg.code)}")
                    return deserialize_torch_tensor(msg.tensor_part), msg
```

Can we do that? Since `amap_in_executor()` is tolerant to exceptions

borzunov (approved):

tests/test_allreduce_fault_tolerance.py
justheuristic: [added two test conditions: timeout while sending and while reducing]

PR #425, opened 2021-12-15
Simplify getting latest checkpoint in examples/albert

mryab:
**From Discord:**

Based on https://github.com/huggingface/transformers/pull/10334, I think it's safe for us to rely on transformers for getting the last checkpoint from output_dir and get rid of it in our code. I've drafted https://github.com/learning-at-home/hivemind/pull/425, will test later this week.

### DIFF
--- a/examples/albert/run_trainer.py
+++ b/examples/albert/run_trainer.py
@@ -40,23 +40,6 @@ def setup_transformers_logging(process_rank: int):
         transformers.utils.logging.enable_propagation()
 
 
-def get_model(training_args, config, tokenizer):
-    # Find latest checkpoint in output_dir
-    output_dir = Path(training_args.output_dir)
-    logger.info(f'Checkpoint dir {output_dir}, contents {list(output_dir.glob("checkpoint*"))}')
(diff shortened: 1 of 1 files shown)


mryab:
Tested it with the checkpoint present in the directory

Before:
```
Dec 27 15:12:14.244 [INFO] Checkpoint dir outputs, contents [PosixPath('outputs/checkpoint-125000'), PosixPath('outputs/checkpoint-124500')]
Dec 27 15:12:14.244 [INFO] Loading model from outputs/checkpoint-125000
```

After (https://github.com/huggingface/transformers/blob/master/src/transformers/trainer.py#L1096):

```
Dec 27 15:13:20.843 [INFO] max_steps is given, it will override any value given in num_train_epochs
Dec 27 15:13:20.843 [INFO] Using amp fp16 backend
Dec 27 15:13:20.844 [INFO] Loading model from outputs/checkpoint-125000).
```

mryab:
Unfortunately, without any checkpoint in the directory, it gives

```
    raise ValueError(f"No valid checkpoint found in output directory ({args.output_dir})")
ValueError: No valid checkpoint found in output directory (outputs)
```

So I think I'll close this for now, since removing the code will not make it possible to launch the same script in both cases and expect it to work

PR #426, opened 2021-12-15
Use hivemind.Optimizer in examples/albert

mryab:
This PR shifts the ALBERT example from CollaborativeOptimizer to hivemind.Optimizer and deprecates the former. Also, it exposes LocalTrainingProgress from Optimizer and adds the `is_synchronized_with_peers` method to this class.


examples/albert/run_trainer.py
```diff
@@ -124,40 +124,43 @@ def on_step_end(
...
                 if self.steps:
                     logger.info(f"Local loss: {self.loss / self.steps}")
-                if self.collaborative_optimizer.local_step % self.backup_every_steps == 0:
+                if self.optimizer.local_epoch % self.backup_every_steps == 0:
                     self.latest_backup = self.backup_state()
 
                 self.loss = 0
                 self.steps = 0
-                if self.collaborative_optimizer.is_synchronized:
+                if self.optimizer.is_synchronized_with_peers():
```
mryab, line 154: This is the main reason why I needed to rename the method

examples/albert/run_trainer.py
```diff
@@ -124,40 +124,43 @@ def on_step_end(
...
                     mini_steps=self.steps,
                 )
-                logger.info(f"Step #{self.collaborative_optimizer.local_step}")
+                logger.info(f"Step #{self.optimizer.local_epoch}")
                 logger.info(f"Your current contribution: {self.total_samples_processed} samples")
                 logger.info(f"Performance: {samples_per_second} samples per second.")
                 if self.steps:
                     logger.info(f"Local loss: {self.loss / self.steps}")
-                if self.collaborative_optimizer.local_step % self.backup_every_steps == 0:
+                if self.optimizer.local_epoch % self.backup_every_steps == 0:
```
mryab, line 149: Actually, this is very confusing for me: should I use `local_epoch` from the optimizer class (basically, from averager) or rely on `local_progress.epoch` here? What's the recommended usage for other situations?

hivemind/optim/experimental/optimizer.py
```diff
@@ -626,9 +629,9 @@ def zero_grad(self, set_to_none: bool = False):
                 else:
                     param.grad.zero_()
 
-    def should_load_state_from_peers(self) -> bool:
+    def is_synchronized_with_peers(self) -> bool:
         """
-        If true, peer will discard local progress and attempt to download state from peers.
+        If false, peer will discard local progress and attempt to download state from peers.
```
mryab, line 634: I'm likely to move this to a comment before loading the state from peers and rewrite the docstring

tests/test_optimizer.py
```diff
@@ -374,12 +374,12 @@ def run_trainer(batch_size: int, batch_time: float, client_mode: bool):
         peer.join()
 
     assert isinstance(optimizer, Optimizer)
-    assert optimizer.local_epoch == optimizer.tracker.global_epoch == total_epochs
+    assert optimizer.local_epoch == optimizer._tracker.global_epoch == total_epochs
```
mryab, line 377: I believe that using private fields in tests in fine

hivemind/optim/collaborative.py
```diff
@@ -101,6 +97,9 @@ class CollaborativeOptimizer(DecentralizedOptimizerBase):
     :param kwargs: additional parameters forwarded to DecentralizedAverager
     :note: If you are using CollaborativeOptimizer with lr_scheduler, it is recommended to pass this scheduler
       explicitly into this class. Otherwise, scheduler may not be synchronized between peers.
+    :note: **For new projects, please use hivemind.Optimizer**. CollaborativeOptimizer is an older version of that.
```
borzunov, line 100: I think it is important to keep that in the beginning of the corresponding docs section: https://learning-at-home.readthedocs.io/en/latest/modules/optim.html#module-hivemind.optim.collaborative

Otherwise, people are likely to miss this warning while taking a quick look on the docs.

tests/test_optimizer.py
borzunov: I agree.

examples/albert/run_trainer.py
```diff
@@ -246,19 +247,18 @@ def main():
...
-        batch_size_per_step=total_batch_size_per_step,
-        bandwidth=collaboration_args.bandwidth,
+        run_id=collaboration_args.experiment_prefix,
         target_batch_size=adjusted_target_batch_size,
+        batch_size_per_step=total_batch_size_per_step,
+        optimizer=opt,
+        scheduler=scheduler,
+        grad_compression=Float16Compression(),
+        state_averaging_compression=Float16Compression(),
+        averager_opts={"bandwidth": collaboration_args.bandwidth, **asdict(averager_args)},
```
borzunov, line 259: Can you please reorder these args to the same order as in the API reference?
https://learning-at-home.readthedocs.io/en/latest/modules/optim.html#module-hivemind.optim.collaborative

I'm trying to fix this order throughout the project, so it's easier to compare the sets of Optimizer args in different places.

borzunov (changes requested):
Thanks for the PR! I've left a few comments.

examples/albert/run_trainer.py
```diff
@@ -246,19 +247,18 @@ def main():
...
         client_mode=collaboration_args.client_mode,
         verbose=True,
-        start=True,
-        **asdict(averager_args),
```
borzunov, line 261: Let's pass other args (such as various timeouts) here as well, as in dalle-hivemind: https://github.com/learning-at-home/dalle-hivemind/blob/main/task.py#L134

examples/albert/run_trainer.py
borzunov: Could you also please check whether every argument from `arguments.py` goes to the right place?

tests/test_optimizer.py
justheuristic: These fields are *not* private and are meant to be exposed to the user.

hivemind/optim/experimental/optimizer.py
justheuristic: This method specifically checks if you should download the state from peers. If that returns True, you **must call .load_state_from_peers**. If you call this method and do not load state from peers, this will result in worse synchronization guarantees.

I'd suggest **creating a new method is_synchronized_with_peers** that simply checks if local_epoch is equal or at most 1 behind the global epoch -- without changing the existing method. Furthermore,

justheuristic (changes requested):
* keep averagers and the tracker public, they are meant to be exposed to the user
* please read the comment on load_state_from_peers
Otherwise a good PR!

examples/albert/run_trainer.py
justheuristic: These are the same. local_epoch is just a shortcut.

examples/albert/run_trainer.py
mryab: AFAIK they're already in the correct order: the link you attached refers to CollaborativeOptimizer, not Optimizer. As for checking the arguments, I launched the script and it appears to train correctly (at least no errors about unrecognized arguments)

hivemind/optim/collaborative.py
mryab: Moved it higher, but below the overview of the class

tests/test_optimizer.py
mryab: Perhaps it is indeed a significant change given current usage: however, I must point out that Optimizer was an experimental class up to this PR and didn't appear in any of the releases, so its API was subject to change at any moment :)

Overall, I agree with not changing these fields to private right away, but we should do it in the next major release due to the reasons specified in the PR message. By that time, we should have a pretty good understanding of what internals of tracker/averager need to be made available by Optimizer

hivemind/optim/experimental/optimizer.py
mryab: Created a method and made the current one private

examples/albert/run_trainer.py
mryab: Actually, there were some errors and inconsistencies; fixed them now

justheuristic (approved):

borzunov (approved):

PR #428, opened 2021-12-18
Improve user-friendliness and fix misc errors

justheuristic:
This PR introduces two quality-of-life changes found with the help of community members

1. Warn if there are too many no-grad params (Found by mr_seeker@ from [KoboldAI](https://github.com/KoboldAI))

The problem manifested itself when he created an optimizer over GPT-J-6B with most parameters set to requires_grad_(False).

Currently, TrainingStateAverager will still average these parameters with peers to properly average batchnorm EMAs and similar statistics. As a result, TrainingStateAverager would OOM after trying to initialize buffers for all gpt-j parameters.

This PR checks if the majority of parameters are non-trainable and prints a warning with instructions.

2. Do not count metric loggers as training peers (found with the help of CALM collaborators)

Currently, non-gpu peers that are responsible for metric logging periodically report having accumulated 0 samples at eps samples/sec. This causes confusion because these peers do not actually compute minibatches and are NOT counted in "averaged gradients with X peers".

3. found in @bawr 's logs
image

4. found by @pr-mais : setting ANNOUNCE_MADDRS with port 0 would silently break matchmaking with some probability
The reason behind this is that P2P does not auto-resolve port 0 in announce_maddrs. We will now raise an error if user sets up a P2P with this configuration error

5. found by @borzunov  - Reducer error when handling network error
image

This is a non-fatal error that happens on reducer when some senders take too long to send their data and are killed -- but their last message does make it.

This is a non-fatal error that causes peers to skip one optimizer step if it experienced gradient overflow in the next training batch.


borzunov (approved):

hivemind/averaging/matchmaking.py
```diff
@@ -242,7 +240,7 @@ async def _request_join_group(self, leader: PeerID) -> Optional[GroupInfo]:
             logger.debug(f"{self} - potential leader {leader} did not respond within {self.request_timeout}")
             return None
         except (P2PHandlerError, ControlFailure, DispatchFailure, StopAsyncIteration) as e:
-            logger.debug(f"{self} - failed to request potential leader {leader}:")
+            logger.debug(f"{self} - failed to request potential leader {leader}: {repr(e)}", exc_info=True)
```
borzunov, line 243:
```suggestion
            logger.debug(f"{self} - failed to request potential leader {leader}:", exc_info=True)
```

No need for the repr, the full traceback will be printed right away.

hivemind/p2p/p2p_daemon.py
```diff
@@ -140,6 +140,11 @@ async def create(
         socket_uid = secrets.token_urlsafe(8)
         self._daemon_listen_maddr = Multiaddr(cls._UNIX_SOCKET_PREFIX + f"p2pd-{socket_uid}.sock")
         self._client_listen_maddr = Multiaddr(cls._UNIX_SOCKET_PREFIX + f"p2pclient-{socket_uid}.sock")
+        if announce_maddrs is not None:
+            for addr in announce_maddrs:
+                if addr.startswith("/ip4/") or addr.startswith("/ip6/"):
+                    if addr.endswith("/tcp/0") or addr.endswith("/udp/0/quic"):
+                        raise ValueError("Please specify an explicit port in announce_maddrs: port 0 is not supported")
```
borzunov, line 147:
```suggestion
                addr = Multiaddr(addr)
                if ("tcp" in addr and addr["tcp"] == "0") or ("udp" in addr and addr["udp"] == "0"):
                    raise ValueError("Please specify an explicit port in announce_maddrs: port 0 is not supported")
```

This is a better way to work with multiaddrs. It accounts for some tricky cases where TCP/UDP is not in the end of the addr/doesn't use QUIC/etc.

hivemind/optim/state_averager.py
```diff
@@ -152,6 +152,15 @@ def _check_params(
         parameter_names = tuple(nested_flatten(parameter_names))
         assert len(parameters) == len(parameter_names), f"Expected {len(parameters)} names, got {len(parameter_names)}"
         assert len(set(parameters)) == len(parameters), "Found duplicate parameters in param_groups"
+        params_with_grad = sum(p.numel() for p in parameters if p.requires_grad)
+        params_no_grad = sum(p.numel() for p in parameters if not p.requires_grad)
+        if params_no_grad >= params_with_grad:
+            logging.warning(
```
mryab, line 158:
```suggestion
            logger.warning(
```

justheuristic: good catch!

hivemind/averaging/allreduce.py
```diff
@@ -343,10 +343,13 @@ async def _accumulate_parts_streaming(self, stream: AsyncIterator[averaging_pb2.
...
-                    sender_index, part_index, tensor_part, weight=weight
-                )
-                part_index += 1
+                try:
+                    averaged_part = await self.tensor_part_reducer.accumulate_part(
+                        sender_index, part_index, tensor_part, weight=weight
+                    )
+                    part_index += 1
+                except BannedException:
+                    break  # sender was banned, we no longer need to aggregate it
```
borzunov, line 352: Let's print a debug log message here if it's not already printed anywhere else :)

borzunov (approved):
I've reviewed the latest changes.

hivemind/averaging/allreduce.py
justheuristic: done

PR #432, opened 2021-12-25
Add PowerSGD for compressed gradient averaging

artek0chumak:
This PR implements the PowerSGD algorithm (https://arxiv.org/abs/1905.13727) for compressed gradient averaging.


hivemind/averaging/averager.py
```diff
@@ -109,7 +109,7 @@ class DecentralizedAverager(mp.Process, ServicerBase):
     """
 
     _matchmaking: Matchmaking
-    _pending_group_assembled: asyncio.Event
+    _pending_groups_registered: asyncio.Event
```
justheuristic, line 112: Note: changes to this file are required so that we can run multiple AllreduceRunners per step (for powerSGD and more in future)

hivemind/optim/grad_averager.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         accumulate_grads_on: Optional[torch.device] = None,
         client_mode: bool = None,
         warn: bool = True,
+        grad_extra_tensors: Sequence[torch.Tensor] = (),
```
justheuristic, line 78:
```suggestion
        averaged_grads: Optional[Sequence[torch.Tensor]] = None,
```

hivemind/optim/optimizer.py
```diff
@@ -187,11 +188,13 @@ def __init__(
         client_mode: bool = None,
         auxiliary: bool = False,
         grad_compression: CompressionBase = NoCompression(),
+        grad_rank_averager: Optional[str] = None,
```
justheuristic, line 191:
```suggestion
        grad_averager: Optional[Type[GradientAverager]] = None,
```

justheuristic: (instead of str AND instead of grad_averager_opts that would be passed via partial)

hivemind/optim/optimizer.py
```diff
@@ -241,6 +244,18 @@ def __init__(
...
+            grad_extra_tensors = [
+                torch.zeros_like(param, device="cpu")
+                for param_group in optimizer.param_groups
+                for param in param_group["params"]
+            ]
+            for tensor in grad_extra_tensors:
+                if tensor is not None:
+                    tensor.share_memory_()
+            grad_averager_opts["grad_extra_tensors"] = grad_extra_tensors
+            extra_tensors = [e for e in extra_tensors] + [eg for eg in grad_extra_tensors]
```
justheuristic, line 258: Let's add some boolean property of GradAverager that determines if we need to add_grads_to_state -- so that it is False for default averager but True for PowerSGD

hivemind/optim/power_ef_averager.py
```diff
@@ -0,0 +1,227 @@
...
+        averager_rank: int,
+        *,
+        dht: hivemind.DHT,
+        prefix: str,
+        reuse_grad_buffers: bool = False,
+        accumulate_grads_on: Optional[torch.device] = None,
+        client_mode: bool = None,
+        warn: bool = True,
+        min_comprasion_ratio: float = 0.5,
+        grad_extra_tensors: Sequence[torch.Tensor] = (),
```
justheuristic, line 61:
```suggestion
        averaged_grads: Optional[Sequence[torch.Tensor]] = None,
```

hivemind/optim/power_ef_averager.py
```diff
@@ -0,0 +1,227 @@
...
+        self.parameters = tuple(parameters)
+        self._uncompressed_gradients = set(
+            i
+            for i, grad in enumerate(self._grads_from_parameters())
+            if len(tuple(grad.size())) == 1
+            or (
+                self.rank * (grad.size(0) + np.prod(grad.size()[1:])) / np.prod(grad.size()) > 1 - min_comprasion_ratio
+            )
+        )
+        self._gradient_rests = list(torch.zeros_like(grad, device="cpu") for grad in self._grads_from_parameters())
```
justheuristic, line 74: the proper term is residual / residue

hivemind/optim/power_ef_averager.py
```diff
@@ -0,0 +1,227 @@
...
+        self._qs = list(
+            torch.rand((grad.reshape((grad.size(0), -1)).size(1), self.rank), device="cpu")
```
justheuristic, line 76: I agree with codecov: let's make a basic test for that. Please see [test_optimizer.py](https://github.com/learning-at-home/hivemind/blob/master/tests/test_optimizer.py#L14) for examples

OPTIONAL: if you have time, let's also add an option to use power EF to `benchmarks/benchmark_optimizer.py`

hivemind/optim/power_ef_averager.py
```diff
@@ -0,0 +1,227 @@
...
+        try:
+            self._running_groups[group_info.group_id + b".phase1"] = asyncio.Future()
+            self._running_groups[group_info.group_id + b".phase2"] = asyncio.Future()
+            self._pending_groups_registered.set()
+            yield
+        finally:
+            maybe_future = self._running_groups.pop(group_info.group_id + b".phase1", None)
+            if maybe_future and not maybe_future.done():
+                logger.warning(f"All-reduce group {group_info.group_id + b'.phase1'} did not finish.")
+            maybe_future = self._running_groups.pop(group_info.group_id + b".phase2", None)
```
justheuristic, line 109: Weak suggestion: use for loop

tests/test_allreduce_fault_tolerance.py
```diff
@@ -66,16 +66,19 @@ async def _run_allreduce(self, group_info: GroupInfo, min_vector_size: int, **kw
                     **kwargs,
                 )
 
-                with self.register_allreduce_group(group_info.group_id, allreduce):
-                    if modes[group_info.peer_ids.index(self.peer_id)] != AveragingMode.AUX:
-                        async for tensor, update in azip(as_aiter(*local_tensors), allreduce):
-                            # all-reduce is performed asynchronously while iterating
-                            tensor.add_(update, alpha=self._averaging_alpha)
-                        self._state_updated.set()
+                self._running_groups[group_info.group_id].set_result(allreduce)
```
justheuristic, line 69: note to others: this is to adjust for the changes into averagar.py for creating multiple groups

hivemind/optim/optimizer.py
```diff
@@ -241,6 +244,18 @@ def __init__(
         self.tracker = self._make_progress_tracker(
             target_batch_size, performance_ema_alpha=performance_ema_alpha, **tracker_opts or {}
         )
+        if grad_rank_averager == "power_ef" and not use_local_updates:
+            assert len(extra_tensors) == 0
+            grad_extra_tensors = [
+                torch.zeros_like(param, device="cpu")
+                for param_group in optimizer.param_groups
```
justheuristic, line 251:
```suggestion
                for param_group in param_groups
```

wherein
```python
params = list(params) if params is not None else optimizer.param_groups
if all(isinstance(p, torch.Tensor) for p in param_groups):
    param_groups = (dict(params=param_groups),)
```

justheuristic:
__TODO__ consider adding an option that disables return_deltas because that would result in slightly different $g$s for each peer due to them compressing different deltas (with different quantization erros)

justheuristic:
TWIMC Current project status:
- @artek0chumak implemented vanilla powerSGD and it works well on the XXL-scale model (equiv. to GPT-J ~2.7B)
image

- during last sync, we decided to adapt vanilla powerSGD as a separate PR to make sure it gets to master quicker
  - and then be used immediately in both language models; probably-but-not-certainly also in vision experiments
- after that: powerSGD + EF21 integration

tests/test_optimizer.py
```diff
@@ -305,7 +316,11 @@ def test_optimizer(
 
     def run_trainer(batch_size: int, batch_time: float, client_mode: bool):
         nonlocal optimizer
-        model = nn.Linear(5, 1)
+        model = nn.Sequential(
```
justheuristic, line 319: Please explain this change

docs/user/quickstart.md
```diff
@@ -58,7 +58,6 @@ opt = hivemind.Optimizer(
     batch_size_per_step=32,   # each call to opt.step adds this many samples towards the next epoch
     target_batch_size=10000,  # after peers collectively process this many samples, average weights and begin the next epoch 
     optimizer=opt,            # wrap the SGD optimizer defined above
-    use_local_updates=True,   # perform optimizer steps with local gradients, average parameters in background
```
justheuristic, line 61: Perhaps we should roll this one back? I'm not sure how it connects to powerSGD

hivemind/optim/grad_averager.py
```diff
@@ -11,6 +11,10 @@
 logger = get_logger(__name__)
 
 
+TGradientAverager = TypeVar("TGradientAverager", bound="GradientAverager")
```
justheuristic, line 14: Question to Max: what's our naming convention for this? personally I'm not sure and fine with either option, but this looks like something that was standardized by more cultured devs :)

hivemind/optim/grad_averager.py
```diff
@@ -11,6 +11,10 @@
...
+GradientAveragerFactory = Callable[[Type[TGradientAverager], Any], TGradientAverager]
```
justheuristic, line 15: Question to Max: what's our naming convention for this? In other parts, i've seen Callable[Any, GradientAveragerType].
Personally I'm not sure and fine with either option, but this looks like something that was standardized by more cultured devs :)

hivemind/optim/grad_averager.py
```diff
@@ -224,3 +230,10 @@ def use_averaged_gradients(self):
     def notify_used_averaged_gradients(self):
         """Notify averager that the results of a previous averaging round are accounted for"""
         self._new_averaged_grads = False
+
+    @classmethod
+    def get_factory(cls, **kwargs1) -> GradientAveragerFactory:
```
justheuristic, line 235: Nitpicking: consider functools.partial

hivemind/optim/optimizer.py
```diff
@@ -34,7 +35,7 @@ class Optimizer(torch.optim.Optimizer):
 
     By default, Optimizer is configured to be exactly **equivalent to synchronous training** with target_batch_size.
     There are advanced options make training semi-asynchronous (delay_optimizer_step and delay_gradient_averaging)
-    or even fully asynchronous (use_local_updates=True).
+    or even fully asynchronous (grad_averager=None).
```
justheuristic, line 38: Weak reminder: whatever we decide on use_local_updates, keep this updated

hivemind/optim/optimizer.py
```diff
@@ -139,10 +140,6 @@ class Optimizer(torch.optim.Optimizer):
       hardly ever skip averaging rounds, they can average state less frequently. In turn, network failures, lossy
       gradient compression and local_updates cause parameters to diverge faster and requires more frequent averaging.
 
-    :param use_local_updates: if enabled, peers will update parameters on each .step using local gradients;
-      if not enabled (default), accumulate gradients to target_batch_size, and then call .step with averaged gradients.
-      Even if use_local_updates=True, learning rate scheduler will still be called once per target_batch_size.
```
justheuristic, line 144: Weak reminder: whatever we decide on use_local_updates, keep this updated

hivemind/optim/optimizer.py
```diff
@@ -183,10 +180,10 @@ def __init__(
         delay_grad_averaging: bool = False,
         delay_state_averaging: bool = True,
         average_state_every: int = 1,
-        use_local_updates: bool = False,
         client_mode: bool = None,
         auxiliary: bool = False,
         grad_compression: CompressionBase = NoCompression(),
+        grad_averager: Optional[GradientAveragerFactory] = PowerSGDGradientAverager.get_factory(averager_rank=32),
```
justheuristic, line 186: Optional: i would recommend renaming it to grad_averager_cls to match [other similar situations](https://github.com/learning-at-home/hivemind/blob/master/hivemind/moe/server/server.py#L101)

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+        dht: hivemind.DHT,
+        prefix: str,
+        reuse_grad_buffers: bool = False,
+        accumulate_grads_on: Optional[torch.device] = None,
+        client_mode: bool = None,
+        warn: bool = True,
+        min_comprasion_ratio: float = 0.5,
+        averaged_grads: Optional[Sequence[torch.Tensor]] = None,
+        **kwargs,
+    ):
```
justheuristic, line 64: [Let's write a docstring here]

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+
+    @contextlib.contextmanager
+    def _register_allreduce_group(self, group_info: GroupInfo):
+        """registers a given all-reduce runner to listen for incoming connections"""
+        try:
+            for phase in self.all_reduce_phases:
+                self._running_groups[group_info.group_id + phase] = asyncio.Future()
+            self._pending_groups_registered.set()
+            yield
+        finally:
```
justheuristic, line 104: Note: codecov seemingly lies here

tests/test_optimizer.py
justheuristic: Resolved: this is likely due to the fact that powersgd needs 2+-dim tensors

tests/test_optimizer.py
```diff
@@ -286,7 +292,12 @@ def run_worker(index: int, batch_size: int, period: float, **kwargs):
 
 
 @pytest.mark.forked
+@pytest.mark.parametrize(
+    "grad_averager",
+    [GradientAverager.get_factory(), PowerSGDGradientAverager.get_factory(averager_rank=1)],
```
justheuristic, line 297:
```suggestion
    [GradientAverager, partial(PowerSGDGradientAverager, averager_rank=1)],
```

justheuristic (approved):
The only serious concern is backward compatibility for optimizer (synchronous by default, use_local_steps param does not get forwarded-through), the rest is totally up for discussion

[Let's wait for Max tho]

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
+import asyncio
+import contextlib
+import faulthandler
```
mryab, line 3: Please check for unused imports, this module appears to have plenty of them

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+import math
+import multiprocessing as mp
+from typing import Any, Iterable, Optional, Sequence
+
+import numpy as np
+import torch
+
+import hivemind
```
mryab, line 11: Better to avoid broad imports inside the library, since it makes refactoring more difficult (due to the possibility of circular imports) and unnecessarily increases the scope of the import statement

hivemind/optim/power_sgd_averager.py
mryab: +1, we shouldn't merge this to master until the key class introduced in the PR is properly documented

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+    as_aiter,
+    azip,
+    enter_asynchronously,
+    switch_to_uvloop,
+)
+from hivemind.utils.grpc import combine_from_streaming, split_for_streaming
+from hivemind.utils.serializer import MSGPackSerializer, SerializerBase
+from hivemind.utils.timed_storage import DHTExpiration, ValueWithExpiration, get_dht_time
+
+from .grad_averager import GradientAverager
```
mryab, line 43: Also, please refrain from relative imports [link1](https://stackoverflow.com/a/54218868) [link2](https://peps.python.org/pep-0008/#imports)

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+        parameters: Iterable[torch.nn.Parameter],
+        averager_rank: int,
+        *,
+        dht: hivemind.DHT,
+        prefix: str,
+        reuse_grad_buffers: bool = False,
+        accumulate_grads_on: Optional[torch.device] = None,
+        client_mode: bool = None,
+        warn: bool = True,
+        min_comprasion_ratio: float = 0.5,
```
mryab, line 61:
```suggestion
        min_compression_ratio: float = 0.5,
```

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+        self.parameters = tuple(parameters)
+        self._uncompressed_gradients = set(
+            i
+            for i, grad in enumerate(self._grads_from_parameters())
+            if len(tuple(grad.size())) == 1
+            or (
+                self.rank * (grad.size(0) + np.prod(grad.size()[1:])) / np.prod(grad.size()) > 1 - min_comprasion_ratio
+            )
+        )
+        self._ms = list(torch.zeros_like(grad, device="cpu").share_memory_() for grad in self._grads_from_parameters())
```
mryab, line 75: It's a matter of preference, but why not just use list comprehension instead of `list()`?

hivemind/optim/optimizer.py
```diff
@@ -222,10 +219,14 @@ def __init__(
                 "(A) hivemind.Optimizer(..., params=params, optimizer=lambda params: create_opt(params)\n"
                 "(B) hivemind.Optimizer(..., optimizer=pre_initialize_optimizer)"
             )
-        if use_local_updates:
+        if grad_averager is None:
             assert not reuse_grad_buffers, "if local_updates is True, gradients will not be accumulated"
             assert not delay_grad_averaging, "if local_updates is True, gradients will not be averaged"
 
+        params = list(params) if params is not None else optimizer.param_groups
```
mryab, line 226: Would a tuple work in this case?

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+        accumulate_grads_on: Optional[torch.device] = None,
+        client_mode: bool = None,
+        warn: bool = True,
+        min_comprasion_ratio: float = 0.5,
+        averaged_grads: Optional[Sequence[torch.Tensor]] = None,
+        **kwargs,
+    ):
+        self.rank = averager_rank
+        self.parameters = tuple(parameters)
+        self._uncompressed_gradients = set(
```
mryab, line 67: This is not a collection of gradients, but a collection of their indices when iterating over parameters. It's better to reflect that in the naming.

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+            i
+            for i, grad in enumerate(self._grads_from_parameters())
+            if len(tuple(grad.size())) == 1
```
mryab, line 70: `grad.numel` or `grad.ndim`?

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+            or (
+                self.rank * (grad.size(0) + np.prod(grad.size()[1:])) / np.prod(grad.size()) > 1 - min_comprasion_ratio
+            )
+        )
+        self._ms = list(torch.zeros_like(grad, device="cpu").share_memory_() for grad in self._grads_from_parameters())
+        self._qs = list(
+            torch.rand((grad.reshape((grad.size(0), -1)).size(1), self.rank), device="cpu").share_memory_()
```
mryab, line 77: Reshaping the tensor just to compute the number of elements of the reshaped form seems quite inefficient (especially since `reshape` triggers a memory copy), can't you just compute it from the number of elements?

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+        **kwargs,
+    ):
+        self.rank = averager_rank
+        self.parameters = tuple(parameters)
+        self._uncompressed_gradients = set(
+            i
+            for i, grad in enumerate(self._grads_from_parameters())
+            if len(tuple(grad.size())) == 1
+            or (
+                self.rank * (grad.size(0) + np.prod(grad.size()[1:])) / np.prod(grad.size()) > 1 - min_comprasion_ratio
```
mryab, line 72: It's not trivial to understand why we use this expression to compute a ratio of data left after compression, a comment would be helpful

tests/test_optimizer.py
mryab: Just to make it clear, do you mean "tensors with >1 dimensions" or "tensors with each dimension having a size of >1"? Because I've tested the first one, and it appears to be true for a (5, 1) linear:

```
>>> a=torch.nn.Linear(5,1)
>>> a.weight.ndim
2
```

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+            )
+        )
+        self._ms = list(torch.zeros_like(grad, device="cpu").share_memory_() for grad in self._grads_from_parameters())
+        self._qs = list(
+            torch.rand((grad.reshape((grad.size(0), -1)).size(1), self.rank), device="cpu").share_memory_()
+            for idx, grad in enumerate(self._grads_from_parameters())
+            if idx not in self._uncompressed_gradients
+        )
+
+        self.all_reduce_phases = (b".phase1", b".phase2")
```
mryab, line 82: Can we give more meaningful names to each of the phases? 

Also, I noticed that you're accessing the elements of this field by index: it would be probably better to make it a `Enum` outside of the class

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+            peer_fractions = await asyncio.get_event_loop().run_in_executor(
+                None, load_balance_peers, self.total_size, download_bandwidths, min_vector_size
+            )
+
+            async with enter_asynchronously(self.get_tensors()) as averaged_grads:
+                for grad, m in zip(averaged_grads, self._ms):
+                    m.add_(grad.to(m.device))
+
+                averaged_sgd_ms = [m for idx, m in enumerate(self._ms) if idx not in self._uncompressed_gradients]
+                averaged_sgd_grad = [
```
mryab, line 130:
```suggestion
                averaged_ms = [m for idx, m in enumerate(self._ms) if idx not in self._uncompressed_gradients]
                averaged_grads = [
```

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+                    grad for idx, grad in enumerate(averaged_grads) if idx not in self._uncompressed_gradients
+                ]
+                ps = [
+                    torch.zeros((grad.size(0), self.rank), device="cpu")
+                    for idx, grad in enumerate(averaged_grads)
+                    if idx not in self._uncompressed_gradients
+                ]
+                for p, q, m in zip(ps, self._qs, averaged_sgd_ms):
+                    torch.matmul(m.reshape(-1, q.size(0)), q, out=p)
```
mryab, line 139: Maybe add a comment that we need to reshape M for non-2D parameter matrices

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+
+                if modes[group_info.peer_ids.index(self.peer_id)] != AveragingMode.AUX:
+                    async for tensor, update in azip(as_aiter(*first_all_reduced), allreduce1):
+                        # all-reduce is performed asynchronously while iterating
+                        tensor.add_(update, alpha=self._averaging_alpha)
+                else:
+                    async for _ in allreduce1:  # trigger all-reduce by iterating
+                        raise ValueError("aux peers should not receive averaged tensors")
+
+                # orth ps
```
mryab, line 163: The comment is redundant: we can either remove it or replace with a reference for a line 5 of Algorithm 1 from the paper

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+                        self._state_updated.set()
+                else:
+                    async for _ in allreduce2:  # trigger all-reduce by iterating
+                        raise ValueError("aux peers should not receive averaged tensors")
+
+                # recompute grads
+                for p, q, m, grad in zip(ps, self._qs, averaged_sgd_ms, averaged_sgd_grad):
+                    new_m = torch.matmul(p, q.t())
+                    m.sub_(new_m.reshape(m.size()))
+                    grad.copy_(new_m.reshape(grad.size()))
```
mryab, line 199: Reshaping a large tensor twice can be expensive, you can do it once in L197

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+
+                for idx, (m, grad) in enumerate(zip(self._ms, averaged_grads)):
+                    if idx in self._uncompressed_gradients:
+                        grad.copy_(m)
+                        m.data[...] = 0
```
mryab, line 204:
```suggestion
                        m.data.zero_()
```

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+
+                return allreduce1.gathered
+        except BaseException as e:
+            logger.exception(e)
+            raise MatchmakingException(f"Unable to run All-Reduce: {e}")
+        finally:
+            pass
```
mryab, line 211: Does this statement have any effect?

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+                        raise ValueError("aux peers should not receive averaged tensors")
+
+                # recompute grads
+                for p, q, m, grad in zip(ps, self._qs, averaged_sgd_ms, averaged_sgd_grad):
+                    new_m = torch.matmul(p, q.t())
+                    m.sub_(new_m.reshape(m.size()))
+                    grad.copy_(new_m.reshape(grad.size()))
+
+                for idx, (m, grad) in enumerate(zip(self._ms, averaged_grads)):
+                    if idx in self._uncompressed_gradients:
```
mryab, line 202: We have this pattern several times in the class, maybe it would be worth having a separate list of gradients that are not compressed

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+            raise MatchmakingException(f"Unable to run All-Reduce: {e}")
+        finally:
+            pass
+
+    def get_current_state(self):
+        with torch.no_grad(), self.lock_averaged_tensors:
+            grad_averager_buffers = list(q for q in self._qs)
+            grad_averager_buffers_infos = [
+                CompressionInfo.from_tensor(buffer, key=f"buffer_q_{key}", role=TensorRole.GRADIENT)
+                for buffer, key in zip(grad_averager_buffers, range(len(grad_averager_buffers)))
```
mryab, line 218: `enumerate`?

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+                return allreduce1.gathered
+        except BaseException as e:
+            logger.exception(e)
+            raise MatchmakingException(f"Unable to run All-Reduce: {e}")
+        finally:
+            pass
+
+    def get_current_state(self):
+        with torch.no_grad(), self.lock_averaged_tensors:
+            grad_averager_buffers = list(q for q in self._qs)
```
mryab, line 215: We can also use a list comprehension here

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,249 @@
...
+    def load_state_from_peers(self, **kwargs):
+        loaded_state = super().load_state_from_peers(**kwargs)
+        if loaded_state is None:
+            return
+
+        metadata, flat_tensors = loaded_state
+        logger.info("Starting loading gradient averager buffers from peers")
+
+        if len(flat_tensors) != len(self._qs):
+            logger.error("Failed to load state from peer, received parameters, extras or metadata")
```
mryab, line 233: The error seems to not correspond to the condition which is checked: do we verify here that the number of tensors is the same?

mryab (changes requested):
Also, please make sure that the core functionality of PowerSGDAverager is covered by the tests, since right now codecov appears not to notice it

hivemind/optim/grad_averager.py
borzunov: Yes, the factories look overcomplicated to me, please try to use `partial()` instead (and remove the factories) if possible

borzunov (changes requested):
Thanks for the PR!

Unfortunately, the interface with `get_factory()` looks overcomplicated to me, please try to use `partial()` instead. I'm okay to keep the word `factory` in the parameter names though :)

hivemind/optim/grad_averager.py
```diff
@@ -75,6 +79,7 @@ def __init__(
         accumulate_grads_on: Optional[torch.device] = None,
         client_mode: bool = None,
         warn: bool = True,
+        averaged_grads: Sequence[torch.Tensor] = (),
```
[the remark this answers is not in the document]
artek0chumak: This argument used to create tensor outside the GradAverager,
e.g. you need tensors to be in StateAverager and GradAverager at the same time.
That why you cann't use only parameters, even we always have same set of them.

tests/test_optimizer.py
artek0chumak: PowerSGD requires one of the dimensions tobe greater than factorization rank. I added additional layer to create possible case to test the averager.

hivemind/optim/optimizer.py
```diff
@@ -405,7 +409,7 @@ def step(
             self._maybe_schedule_state_averaging()
 
         else:
-            # use_local_updates=True: update parameters on every step independently of other peers
+            # grad_averager=None: update parameters on every step independently of other peers
```
justheuristic, line 412:
```suggestion
            # use_local_updates=True: update parameters on every step independently of other peers
```

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,273 @@
...
+    azip,
+    enter_asynchronously,
+    switch_to_uvloop,
+)
+from hivemind.utils.grpc import combine_from_streaming, split_for_streaming
+from hivemind.utils.math import orthogonalize_
+from hivemind.utils.serializer import MSGPackSerializer, SerializerBase
+from hivemind.utils.timed_storage import DHTExpiration, ValueWithExpiration, get_dht_time
+
+from .grad_averager import GradientAverager
```
justheuristic, line 43:
```suggestion
from hivemind.optim.grad_averager import GradientAverager
```

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,273 @@
...
+        min_compression_ratio: float = 0.5,
+        averaged_grads: Optional[Sequence[torch.Tensor]] = None,
+        **kwargs,
+    ):
+        self.rank = averager_rank
+        self.parameters = tuple(parameters)
+        self._uncompressed_gradients_indexes = set(
+            i
+            for i, grad in enumerate(self._grads_from_parameters())
+            if len(tuple(grad.size())) == 1
```
justheuristic, line 107:
```suggestion
            if grad.ndim <= 1
```

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,273 @@
...
+                ]
+                for grad, m in zip(averaged_grads_via_sgd, self._ms):
+                    m.add_(grad.to(m.device))
+
+                ps = [
+                    torch.zeros((grad.size(0), self.rank), device="cpu")
+                    for idx, grad in enumerate(averaged_grad_via_sgd)
+                ]
+                for p, q, m in zip(ps, self._qs, self._ms):
+                    # we use reshape for all matrixes because sgd works only with 2d tensors
```
justheuristic, line 179:
```suggestion
                    # we use reshape for all matrixes because power-sgd works only with 2d tensors
```

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,273 @@
...
+    def load_state_from_peers(self, **kwargs):
+        loaded_state = super().load_state_from_peers(**kwargs)
+        if loaded_state is None:
+            return
+
+        metadata, flat_tensors = loaded_state
+        logger.info("Starting loading gradient averager buffers from peers")
+
+        if len(flat_tensors) != len(self._qs):
+            logger.error("Failed to load state from peer, received parameters, extras or metadata")
```
justheuristic, line 268:
```suggestion
            logger.error("Failed to load state from peer, received invalid parameters, extras or metadata")
```

borzunov (approved):

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,246 @@
...
+            accumulate_grads_on=accumulate_grads_on,
+            client_mode=client_mode,
+            warn=warn,
+            averaged_grads=None,
+            **kwargs,
+        )
+
+    @contextlib.contextmanager
+    def _register_allreduce_group(self, group_info: GroupInfo):
+        """registers a given all-reduce runner to listen for incoming connections"""
```
mryab, line 114: The docstring appears to be incorrect

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,248 @@
...
+        **kwargs,
+    ):
+        self.rank = averager_rank
+        self.parameters = tuple(parameters)
+        self._uncompressed_gradients_indexes = set(
+            i
+            for i, grad in enumerate(self._grads_from_parameters())
+            if grad.ndim <= 1
+            or (1 - self.rank * sum(get_flatten_greedy_dims(grad)) / grad.numel()) < min_compression_ratio
+            # compute how much parameters can we left via factorization
```
mryab, line 87:
```suggestion
            # compute how much parameters are left after factorization
```

hivemind/averaging/averager.py
```diff
@@ -519,47 +536,39 @@ async def _run_allreduce(self, group_info: GroupInfo, min_vector_size: int, **kw
...
-            self._running_groups[group_id] = allreduce
-            self._pending_group_assembled.set()
-            yield
-        finally:
-            self._running_groups.pop(group_id, None)
-            self._pending_group_assembled.set()
+    async def _run_allreduce_inplace_(
+        self, tensors: Sequence[torch.Tensor], group_info: GroupInfo, group_id: Optional[bytes] = None, **kwargs
+    ):
+        """Run one allreduce process to average tensors inplace. Can be called more a few times in one aggergate process"""
```
mryab, line 549:
```suggestion
        """Run one allreduce process to average tensors inplace. Can be called more than a few times in one aggregation process"""
```

hivemind/optim/power_sgd_averager.py
```diff
@@ -0,0 +1,223 @@
...
+
+    As a result, PowerSGD only needs to aggregate O((m + n) * r) tensors instead of O(m * n).
+    High r, e.g. sqrt(max(m, n)) typically reduce communication by 2-8x without affecting convergence.
+    Low r, e.g. 1-8, further accelerate communication, but may converge worse depending on the task.
+
+    To maintain convergence with low r, this averager uses the error feedback strategy. Put simply,
+    if some part of the gradient is "lost in compression", it will be added to the next iteration.
+    This has two implications: (a) it needs more RAM in order to store the "feedback buffers"
+    and (b) if devices stay alive only for one step, training with small rank may converge slower.
+    This is because error feedback takes multiple step to kick in.
```
mryab, line 44:
```suggestion
    This is because error feedback takes multiple steps to kick in.
```

mryab (approved):

Issue #387, opened 2021-09-21
Simple mistakes trigger unclear error messages in the ALBERT example


borzunov:
Simple mistakes trigger unclear error messages in the ALBERT example, that is:

- Absence of the unpacked data for trainer (currently triggers `requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://huggingface.co/api/models/data/tokenizer`)
- Running all peers in `--client_mode` (currently triggers `AllReduce failed: could not find a group`)

It would be great to show a clear error message in these cases.

soodoshll:
Hi Alexander,
I find that running with only one regular trainer(+training monitor) also triggers the second problem. I guess it should skip the averaging when there's only one trainer?

borzunov:
@soodoshll Thanks for the report. In this case, training proceeds normally, so you can proceed to connecting more peers despite the error.

To be honest, I was not sure if we should remove it, since hivemind is designed to be used with several trainers, and a trainer finding itself to be alone is rather a symptom of connection issues in real training runs.

However, I now agree that this message is confusing for people running our example for the first time, so we should remove it.

PR #440, opened 2022-01-04
Skip gradient averaging if there are no other peers

justheuristic:
This resolves the bug found by @soodoshll here: https://github.com/learning-at-home/hivemind/issues/387#issuecomment-1001821504

### DIFF
--- a/hivemind/optim/optimizer.py
+++ b/hivemind/optim/optimizer.py
@@ -528,7 +528,7 @@ class Optimizer(torch.optim.Optimizer):
                 logger.log(self.status_loglevel, f"Tagging along for a pre-scheduled gradient averaging round")
                 self._tag_along_with_zero_weight(self.scheduled_grads)
             else:
-                logger.log(self.status_loglevel, f"Cancelling pre-scheduled averaging round: there are no other peers")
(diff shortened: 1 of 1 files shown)


hivemind/optim/optimizer.py
```diff
@@ -524,8 +524,12 @@ def _begin_averaging_gradients(self, grad_scaler: Optional[GradScaler]) -> bool:
                 logger.exception(e)
 
         if not began_averaging_gradients and self.scheduled_grads is not None and not self.scheduled_grads.done():
-            logger.log(self.status_loglevel, f"Tagging along for a pre-scheduled gradient averaging round")
-            self._tag_along_with_zero_weight(self.scheduled_grads)
+            if self.tracker.global_progress.num_peers > 1:
+                logger.log(self.status_loglevel, f"Tagging along for a pre-scheduled gradient averaging round")
+                self._tag_along_with_zero_weight(self.scheduled_grads)
+            else:
+                logger.log(self.status_loglevel, f"Cancelling pre-scheduled averaging round: there are no other peers")
```
borzunov, line 531:
```suggestion
                logger.log(self.status_loglevel, f"Skipping pre-scheduled averaging round: there are no other peers")
```

Because `Cancelling` sounds like we cancel an already started process abruptly.

borzunov (approved):

PR #441, opened 2022-01-08
Move hivemind.Server from init, streamline imports

mryab:
This is a chore PR that mainly moves hivemind.Server and related functions into a separate file to keep `server/__init__.py` simple and avoid exposing unintended methods at the top level of the library. 

Also, I replaced broad `import hivemind` statements with more specific ones in relevant parts of the project to narrow down their scope and avoid possible circular imports in the future. 

Finally, I moved `generate_uids_from_pattern` to `server.py`, since the refactoring from the first step exposed a circular dependency in the module. This can be improved in the future via better encapsulation of ExpertUID (right now its constants are used all over `hivemind.moe`); for now, I believe the solution with a private function is fine.


hivemind/moe/server/__init__.py
```diff
@@ -1,356 +1,4 @@
...
-
-class Server(threading.Thread):
-    """
-    Server allows you to host "experts" - pytorch sub-networks used by Decentralized Mixture of Experts.
-    After creation, a server should be started: see Server.run or Server.run_in_background.
-
-    A working server does 3 things:
-     - processes incoming forward/backward requests via Runtime (created by the server)
-     - publishes updates to expert status every :update_period: seconds
-     - follows orders from HivemindController - if it exists
```
mryab, line 43: image

I guess this was something that we planned to implement, but never did. Stumbled upon it by pure chance

justheuristic: Curiously, the doc statement is still technically correct :D

justheuristic (approved):

PR #442, opened 2022-01-08
Change make_empty to make_zeros for TensorDescriptor

mryab:
Right now, `make_empty` will fail when trying to make a dummy forward pass with an expert that accepts LongTensor objects as an input (e.g., embeddings), since usually not all integer values are valid inputs. Since on the client side we actually initialize those empty tensors with zeros too, the simplest solution appears to be just making zero tensors in each case by default.


justheuristic (approved):
Good point. I believe that our original motivation for empty was because it is faster. However, it appears that they are not on a critical path, so that motivation is no longer valid.

PR #444, opened 2022-01-13
Fix "too many open files" issue

yhn112:
Using memory sharing strategy not relying on file descriptors. See https://pytorch.org/docs/stable/multiprocessing.html#sharing-strategies

(Sorry, didn't think a lot about prettifying this PR)

### DIFF
--- a/examples/albert/requirements.txt
+++ b/examples/albert/requirements.txt
@@ -4,4 +4,4 @@ torch_optimizer==0.1.0
 wandb==0.10.26
 sentencepiece
 requests
-nltk==3.6.5
+nltk==3.6.7
--- a/hivemind/averaging/averager.py
+++ b/hivemind/averaging/averager.py
@@ -178,7 +178,7 @@ class DecentralizedAverager(mp.Process, ServicerBase):
         else:
             self.mode = AveragingMode.NODE
         self.daemon = daemon
-
+        torch.multiprocessing.set_sharing_strategy(os.environ.get("HIVEMIND_MEMORY_SHARING_STRATEGY", "file_system"))
         self._averaged_tensors = tuple(averaged_tensors)
         self.lock_averaged_tensors = mp.Lock()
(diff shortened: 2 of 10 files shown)


justheuristic (approved):
To summarize our earlier discussion
- awesome work!
- i'd recommend keeping set_sharing_strategy only in mpfuture.py. However, i'm ready to be outvoted :)
  - Rationale 1: some of the methods where it is added (e.g. _make_host_tensor) are called thousand(s) of times
  - Rationale 2: if user changes/resets the environment variable, it will introduce weird behavior with a mix of both strategies. The current behavior ("Look once during init" is, while not perfect, fairly standard among other libraries such as pytorch and numpy)
- it would be great to reduce test timeouts to ~1 second

borzunov (changes requested):
Thanks for investigating this!

Please either keep reading the env var only in `mpfuture.py` (preferred given the @justheuristic 's rationale), or extract this line to a separate function in `hivemind.utils` and call it everywhere.

Also, please address @mryab's questions in Discord.

justheuristic:
Just a reminder: this is a very important issue. It aint urgent, but please don't forget about it.

yhn112:
> 
@justheuristic, I thought it's obvious that the delay with this issue has something to do with non-working CI
Check #448 first, please

yhn112:
@borzunov 
> Also, please address @mryab's questions in Discord.

We didn't face any performance degradation but will keep an eye on it.
Unfortunately, there is very little information with the compassion of the sharing strategies. But kinda everything I met, mentioned `'file_system'` as a solution to some problem, not the source of it. Let's try and see. 🤷‍♂️

borzunov (approved):
Thanks! Two minor comments:

- Please use `time.sleep(0.5)` if it's enough. We have a lot of tests with sleep, and these delays are accumulated.
- Please clarify the purpose of sleep in a comment.

Issue #447, opened 2022-01-18
[BUG] Loss did not decrease in Albert example after 125000 max step.


elricwan:
**Describe the bug**
I run the albert example with wikitext data. I use one peer, default settings (target_batch_size=4096, train_batch_size=4, max_step=125000, lr=0.00176), but the loss did not decrease after training, it start as 11 and finish as 11.

Jan 15 10:30:14.734 [INFO] Step #1      loss = 11.04938
Jan 15 10:32:14.842 [INFO] Step #2      loss = 11.05589
Jan 15 10:34:14.975 [INFO] Step #3      loss = 11.06803
Jan 15 10:36:15.093 [INFO] Step #4      loss = 11.06271
Jan 15 10:38:15.228 [INFO] Step #5      loss = 11.06433
Jan 15 10:40:15.337 [INFO] Step #6      loss = 11.05447
Jan 15 10:41:45.401 [INFO] Step #7      loss = 11.06115
Jan 15 10:43:45.541 [INFO] Step #8      loss = 11.06025
..........
Jan 15 18:09:13.117 [INFO] Step #238    loss = 11.05597
Jan 15 18:11:13.233 [INFO] Step #239    loss = 11.06724
Jan 15 18:13:13.369 [INFO] Step #240    loss = 11.06289
Jan 15 18:15:13.494 [INFO] Step #241    loss = 11.05922
Jan 15 18:16:43.577 [INFO] Step #242    loss = 11.05226
Jan 15 18:18:43.691 [INFO] Step #243    loss = 11.05418
Jan 15 18:20:43.843 [INFO] Step #244    loss = 11.05638

**To Reproduce**
Run the script in albert example.
For monitor, I run:

python run_training_monitor.py \
--experiment_prefix albert_experiment \
--wandb_project albert_wandb

For trainer, I run:

IP=/ip4/192.168.0.188/tcp/45731/p2p/QmSRerwCPUSreHhwMuTLHoVHqTfWuT8J57w3sXFZtU8ECo

WANDB_DISABLED=true CUDA_VISIBLE_DEVICES=0 python run_trainer.py \
--experiment_prefix albert_experiment  \
--initial_peers $IP  \
--logging_first_step  \
--output_dir ./outputs \
--overwrite_output_dir \
--logging_dir ./logs \
--dataset_path="/home/protago/Xiangpeng/hivemind/examples/albert/data/albert_tokenized_wikitext" \
--per_device_train_batch_size 4 \
[...]

PR #450, opened 2022-01-19
Fix offloaded optimizer with single peer

justheuristic:
The bug was originally found by @elricwan and @finger92 (seemingly independently) and reported in https://github.com/learning-at-home/hivemind/issues/447

Here's what was caused the issue:
> I investigated what went wrong in when training with only one trainer. Currently, hivemind.Optimizer is hard-wired to use the averaged gradients -- as in "averaged with peers".
> 
> If you are the only peer, gradients are not averaged, so optimizer runs with zero gradients all the time. This change should fix the problem in your specific case: [4ffd9ca](https://github.com/learning-at-home/hivemind/commit/4ffd9ca65c3e040c9599f4c64c0b960190c42250) I have seemingly introduced that bug myself in #440 . It only affects the github version of hivemind (i.e. not the pypi version)

The bug was introduced in #440 and affects the following setups (all 3 must be true):
- installed hivemind from github
- there is only one training peer in the swarm
- offload_optimizer is True

Here's the behavior after the fix is introduced:
image

### DIFF
--- a/hivemind/optim/optimizer.py
+++ b/hivemind/optim/optimizer.py
@@ -529,6 +529,7 @@ class Optimizer(torch.optim.Optimizer):
                 self._tag_along_with_zero_weight(self.scheduled_grads)
             else:
                 logger.log(self.status_loglevel, f"Skipping pre-scheduled averaging round: there are no other peers")
+                self.grad_averager.load_accumulators_into_averager_()
                 self.scheduled_grads.cancel()
             self.scheduled_grads = None
         return began_averaging_gradients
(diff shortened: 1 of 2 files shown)


hivemind/optim/optimizer.py
```diff
@@ -618,6 +618,12 @@ def _load_averaged_gradients_into_optimizer_(self):
 
         self.grad_averager.notify_used_averaged_gradients()
 
+    def _load_local_gradients_into_optimizer(self):
+        """Fallback to using local gradients in the optimizer (instead of averaged gradients)"""
+        logger.log(self.status_loglevel, f"Proceeding with local gradients")
```
borzunov, line 623: Please add a comment that this can be optimized in case of one peer (if we'd ever need to optimize this case).

borzunov (approved):

hivemind/optim/optimizer.py
justheuristic: just did it

Issue #446, opened 2022-01-14
Question about loading checkpoint 


finger92:
I was able to resume training by only 'load_state_dict' in monitor peer before using hivemind 1.0.0 version.
The code looks like this:
```
# monitor peer
if load_from_pretrained:
  self.model.load_state_dict(torch.load("pytorch_model.bin", map_location="cpu"), strict=False)
  ...
  self.collaborative_optimizer.load_state_dict(torch.load("optimizer.pt", map_location="cpu"))
```
The peers would load from monitor's state after start up.

However, in ver 1.0.0 or master code, 'load_state_dict' in monitor seems not work. 
My question is am I using the wrong method or should I load the checkpoint on the worker peer?

justheuristic:
Hi! Thanks for the report. I'll appreciate if you can run a few tests to isolate the problem.

__Note:__ there's quite a bunch of them. If you find any discrepancies, you don't need to run the rest of tests. Also: feel free to reach out if you need any assistance with running these checks.

__Q0:__ If possible, please specify hivemind versions you used: both the old one and the new one

__Q1:__ In the new version, when you launch one monitor and one training peer, does the __training peer__ print `Downloading parameter from <long PeerID string>`?

__Q2:__ Are you loading a checkpoint saved in an earlier hivemind version or in master? If new hivemind fails to load old checkpoint, does it load the checkpoint that was saved in the new version?

__Q3:__ Please print model and optimizer checksums on the monitor ``right after it has loaded`` -- both in the old and new hivemind version:

```python
print("Local epoch:", self.collaborative_optimizer.local_epoch)
print("Params checksum:", sum(p.sum().item() for p in self.model.parameters()))
print("Optimizer checksum:", sum(v.data.numpy().sum() for k,v in self.collaborative_optimizer.state_dict().items()))
```

And the print same values in a training peer right after it loads state for the first time.
- Does local_epoch match the epoch that was used when you last saved state?
- Do these values match? If they do not, did they match in earlier version?

__Q4:__ Can you please check if all keys match successfully? (print outputs of whatever.load_state_dict(...) - it contains a report of which keys matched and which did not)

finger92:
Q0: 0.10.0 vs 1.0.0
Q1:  I found a warn
[WARN] [hivemind.optim.state_averager.load_state_from_peers:669] Failed to load state from peer, received inconsistent number of optimizer statistics
Could this be the reason for fail to load checkpoint?

borzunov:
[...]

PR #452, opened 2022-01-24
Fix monitor state_averager in examples/albert (preliminary)

justheuristic:
This PR fixes several minor issues found in #446 : 

- fix `prefix=...` in training monitor
- create scheduler in training monitor
- rename experiment_prefix -> run_id
- enable checkpoints on aux peer by default
- decouple total steps from scheduler max steps

Tests:
- [x] train for a few hundred steps
- [x] download state from aux peer
image

In a subsequent PR:
- __create optimizer normally in training monitor (instead of TrainingStateAverager)__
- stream data from wikitext instead of preprocessing it
- extract common code from trainer and monitor to utils
- allow training monitor to assist in averaging
- refer to advanced example

### DIFF
--- a/examples/albert/README.md
+++ b/examples/albert/README.md
@@ -55,7 +55,7 @@ To join the collaboration with a GPU trainer,
   (see [default paths](./arguments.py#L117-L134) for reference)
 - Run:
   ```bash
-  ./run_trainer.py  --initial_peers ONE_OR_MORE_PEERS --per_device_train_batch_size BATCH_SIZE_FOR_YOUR_GPU_DEFAULT_4
+  ./run_trainer.py  --initial_peers ONE_OR_MORE_PEERS --per_device_train_batch_size BATCH_SIZE_FOR_YOUR_GPU
   ```
 
   Here, `ONE_OR_MORE_PEERS` stands for multiaddresses of one or multiple existing peers (training monitors or existing
--- a/examples/albert/README.md
+++ b/examples/albert/README.md
@@ -55,9 +55,7 @@ To join the collaboration with a GPU trainer,
   (see [default paths](./arguments.py#L117-L134) for reference)
 - Run:
   ```bash
-  ./run_trainer.py \
-      --initial_peers ONE_OR_MORE_PEERS \
-      --logging_first_step --output_dir ./outputs --overwrite_output_dir --logging_dir ./logs
+  ./run_trainer.py  --initial_peers ONE_OR_MORE_PEERS --per_device_train_batch_size BATCH_SIZE_FOR_YOUR_GPU_DEFAULT_4
   ```
 
   Here, `ONE_OR_MORE_PEERS` stands for multiaddresses of one or multiple existing peers (training monitors or existing
@@ -81,6 +79,9 @@ To join the collaboration with a GPU trainer,
 
   You may need to change the IP address to a publicly visible one if some of the initial peers are located behind NAT.
   If you have any trouble doing this, consider the ["Using IPFS"](#using-ipfs) section.
+
+  The `BATCH_SIZE_FOR_YOUR_GPU` should be tweaked so that the model fits into your GPU memory.
+  For 1080Ti or 2080Ti gpus, a good initial value is 4. For 8GB GPUs, try batch size 1-2.
 
 See the ["Tips and tricks"](#tips-and-tricks) section for more information on setting up collaborative training.
 
--- a/examples/albert/arguments.py
+++ b/examples/albert/arguments.py
@@ -148,4 +148,4 @@ class AlbertTrainingArguments(TrainingArguments):
 
     save_total_limit: int = 2
     save_steps: int = 500
-    max_steps: int = 10 ** 30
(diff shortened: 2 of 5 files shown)


examples/albert/README.md
```diff
@@ -55,9 +55,7 @@ To join the collaboration with a GPU trainer,
   (see [default paths](./arguments.py#L117-L134) for reference)
 - Run:
   ```bash
-  ./run_trainer.py \
-      --initial_peers ONE_OR_MORE_PEERS \
-      --logging_first_step --output_dir ./outputs --overwrite_output_dir --logging_dir ./logs
+  ./run_trainer.py  --initial_peers ONE_OR_MORE_PEERS --per_device_train_batch_size BATCH_SIZE_FOR_YOUR_GPU_DEFAULT_4
```
borzunov, line 58: nit:

```suggestion
  ./run_trainer.py  --initial_peers ONE_OR_MORE_PEERS --per_device_train_batch_size BATCH_SIZE_FOR_YOUR_GPU
```

Because (1) `BATCH_SIZE_FOR_YOUR_GPU_DEFAULT_4` is hard to parse and (2) we're talking about the default value below.

examples/albert/arguments.py
```diff
@@ -138,9 +138,14 @@ class AlbertTrainingArguments(TrainingArguments):
...
+    output_dir: str = "outputs"
     logging_steps: int = 100
+    logging_first_step: bool = True
+    overwrite_output_dir: bool = True
+
     save_total_limit: int = 2
     save_steps: int = 500
-
-    output_dir: str = "outputs"
+    max_steps: int = 10 ** 30
```
borzunov, line 151:
```suggestion
```

Let's do of the options below:

(1) remove it (it's inf by default AFAIU)
(2) set it to a reasonable value
(3) add a comment that you mean `inf` by writing `10 ** 30` (now, it's not 100% clear what this number means, someone may think that you really need that much microsteps)

borzunov (approved):
Awesome PR!

examples/albert/run_training_monitor.py
```diff
@@ -40,6 +40,7 @@ class TrainingMonitorArguments(BaseTrainingArguments):
     wandb_project: Optional[str] = field(
         default=None, metadata={"help": "Name of Weights & Biases project to report the training progress to"}
     )
+    store_checkpoints: bool = field(default=True, metadata={"help": "If False, disables CheckpointHandler altogether"})
```
mryab, line 43:
```suggestion
    store_checkpoints: bool = field(default=True, metadata={"help": "If False, disables periodic checkpoint saving"})
```

CheckpointHandler is an internal detail of this script which might not be known to the user

examples/albert/arguments.py
mryab: Is there a point in continuing training after `total_steps` have passed? It seems that with the current schedule, the learning rate for all subsequent updates will be 0

examples/albert/run_training_monitor.py
```diff
@@ -99,7 +99,8 @@ def __init__(
         self.state_averager = TrainingStateAverager(
             dht=dht,
             optimizer=opt,
-            prefix=experiment_prefix,
+            scheduler=get_linear_schedule_with_warmup(opt, num_warmup_steps=5000, num_training_steps=125_000),
```
mryab, line 102: It's better to replace these values with the defaults from the config to avoid possible divergence in the future and general "magic constant" readability issues. The best way might be to move the number of training and warmup steps to a common config used both in the monitor and the trainer code

examples/albert/arguments.py
justheuristic: went with a 3

examples/albert/run_training_monitor.py
borzunov: I think we shouldn't just move args to a common config, but move the whole code creating `hivemind.Optimizer` to a common file. Otherwise, it's likely that any model/optimizer/scheduler changes made by a user will lead to a divergence between monitors and trainers

examples/albert/arguments.py
justheuristic: it is, but I don't get what do you mean by this

examples/albert/run_training_monitor.py
justheuristic: This dataclass is currently not available to aux peer. For long-term optimal composition, please see CALM and/or petabert examples

examples/albert/arguments.py
borzunov: What's the point of decoupling these params?

justheuristic: Ah, I see the problem.

max_steps is the number of local steps, not global optimizer updates. The reason why previous version was incorrect is because peers would run for 125k local forward-backward cycles which would be much less than 125k scheduler epochs

borzunov: I think it's still unclear that (1) total_steps and max_steps are measured in different units and (2) the difference between them is still elusive for a user.

So I'm for removing max_steps completely (maybe in a future PR since this may require you to test the scripts again)

justheuristic: I've added explanations to each parameter to be safe

PR #454, opened 2022-02-08
Pin pytest version in requirements-dev, use file_descriptor in tests

justheuristic:
This fixes the "previous item was not torn down properly" error in CI.

https://github.com/pytest-dev/pytest/issues/9621

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,4 @@
-pytest==6.2.3
+pytest==6.2.3  # newer version has a different teardown interface
 pytest-forked
 pytest-asyncio==0.16.0
 pytest-cov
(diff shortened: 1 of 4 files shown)


requirements-dev.txt
```diff
@@ -1,4 +1,4 @@
-pytest
+pytest==6.2.3
```
justheuristic, line 1:
```suggestion
pytest==6.2.3  # newer version has different teardown interface
```

mryab (approved):

justheuristic:
(copied from discord)

- tests failing was not just random, there was a pattern to it
- pytest failed specifically with codecov enabled
- and it failed specifically on test_mode -> remote_call_many

Here's a proof: removing codecov fixes the issue https://github.com/learning-at-home/hivemind/runs/5124226685?check_suite_focus=true 
I'm now investigating if this is fixed by pinning codecov version to be compatible to that pytest

mryab:
To conclude, the CI failures for the pytest/codecov jobs are related to https://github.com/pytest-dev/pytest/issues/9621 and https://github.com/pytest-dev/pytest-cov/issues/520; adding these issues here to have backlinks to this PR in their discussion threads

mryab (approved):

PR #456, opened 2022-02-12
Pin isort version, bump black

mryab:
This PR bumps the version of black to the first [officially stable](https://black.readthedocs.io/en/stable/change_log.html#id1) version of 22.1.0, applying the required formatting changes. Also, it fixes the version of isort to prevent any unwanted inconsistencies in the future.

### DIFF
--- a/.github/workflows/check-style.yml
+++ b/.github/workflows/check-style.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: psf/black@stable
         with:
           options: "--check --diff"
-          version: "21.6b0"
+          version: "22.1.0"
   isort:
     runs-on: ubuntu-latest
     steps:
@@ -22,3 +22,6 @@ jobs:
         with:
           python-version: 3.8
       - uses: isort/isort-action@master
+        with:
+          isortVersion: "5.10.1"
+
--- a/.github/workflows/check-style.yml
+++ b/.github/workflows/check-style.yml
@@ -24,4 +24,3 @@ jobs:
       - uses: isort/isort-action@master
         with:
           isortVersion: "5.10.1"
-
--- a/hivemind/moe/client/switch_moe.py
+++ b/hivemind/moe/client/switch_moe.py
@@ -150,7 +150,7 @@ class RemoteSwitchMixtureOfExperts(RemoteMixtureOfExperts):
         # for each grid dimension, sum across all indices for a dimension. Optimizing this leads to uniform allocation
         balancing_loss = torch.stack(
             [
-                torch.mean(dim_softmax.mean(0) * dim_utilization) * (dim_size ** 2)
+                torch.mean(dim_softmax.mean(0) * dim_utilization) * (dim_size**2)
                 for dim_softmax, dim_utilization, dim_size in zip(
                     grid_softmax, self.grid_utilization, self.beam_search.grid_size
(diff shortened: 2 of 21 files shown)


yhn112 (approved):
LGTM

tests/test_p2p_daemon_bindings.py
```diff
@@ -125,7 +125,7 @@ async def test_read_write_unsigned_varint_max_bits_edge(max_bits):
     Test edge cases with different `max_bits`
     """
     for i in range(-3, 0):
-        integer = i + (2 ** max_bits)
+        integer = i + (2**max_bits)
```
borzunov, line 128:
```suggestion
        integer = i + 2**max_bits
```

tests/test_p2p_daemon_bindings.py
```diff
@@ -94,7 +94,7 @@ async def test_write_unsigned_varint_overflow(integer):
         await write_unsigned_varint(s, integer)
 
 
-@pytest.mark.parametrize("integer", (-1, -(2 ** 32), -(2 ** 64), -(2 ** 128)))
+@pytest.mark.parametrize("integer", (-1, -(2**32), -(2**64), -(2**128)))
```
borzunov, line 97:
```suggestion
@pytest.mark.parametrize("integer", (-1, -2**32, -2**64, -2**128))
```

hivemind/moe/client/switch_moe.py
```diff
@@ -150,7 +150,7 @@ def forward(self, input: torch.Tensor, *args: torch.Tensor, **kwargs: torch.Tens
         # for each grid dimension, sum across all indices for a dimension. Optimizing this leads to uniform allocation
         balancing_loss = torch.stack(
             [
-                torch.mean(dim_softmax.mean(0) * dim_utilization) * (dim_size ** 2)
+                torch.mean(dim_softmax.mean(0) * dim_utilization) * (dim_size**2)
```
borzunov, line 153:
```suggestion
                torch.mean(dim_softmax.mean(0) * dim_utilization) * dim_size**2
```

.github/workflows/check-style.yml
```diff
@@ -22,3 +22,6 @@ jobs:
         with:
           python-version: 3.8
       - uses: isort/isort-action@master
+        with:
+          isortVersion: "5.10.1"
+
```
borzunov, line 27: nit: Too many newlines

borzunov (approved):
Thanks for the PR!

I think it's worth to clarify that the new release is [the first non-beta release](https://github.com/psf/black/releases/tag/22.1.0) of black. Projects like Django committed to move to black as soon as this release comes out, and I agree we can also do that.

tests/test_p2p_daemon_bindings.py
yhn112: Disagree
Parenthesis save a few seconds of understanding in cases like this

.github/workflows/check-style.yml
mryab: Removed

tests/test_p2p_daemon_bindings.py
mryab: I'm also for keeping the brackets in this case

PR #460, opened 2022-03-04
Clean compression/__init__.py

borzunov:
This PR moves the functions defined in `compression/__init__.py` to `compression/serialization.py`, so the `__init__.py` contains only the imports.


justheuristic (approved):
Thanks!

PR #465, opened 2022-04-20
Implement a CLI for hivemind.DHT

mryab:
This PR creates a command-line script for creating a DHT instance which becomes available directly after the installation of hivemind. Importantly, this script does not create an auxiliary peer or allow to monitor/store task-specific metrics — the goal is simply to have a quick way to create a DHT instance to establish connectivity; later on, more specialized peers with user code may join the network.

Since the peer needs to be exposed to the network, I've moved the `log_visible_maddrs` function to `utils.networking`, further simplifying the ALBERT example folder. Also, I've fixed a couple of typos and reduced the high-level interaction surface of hivemind.utils(.networking)

In the future, we may add an interactive version of `hivemind-dht` similar to a database or key-value storage CLI client; e.g., with an ability to request or store specific keys and list the available nodes. This may greatly simplify debugging "lost" keys or DHT connectivity issues; for the time being, I've added a simpler version that monitors the DHT state without additional network load.

### DIFF
--- a/examples/albert/README.md
+++ b/examples/albert/README.md
@@ -27,7 +27,7 @@ Run the first DHT peer to welcome trainers and record training statistics (e.g.,
 
 ```
 $ ./run_training_monitor.py --wandb_project Demo-run
-Oct 14 16:26:36.083 [INFO] Running a DHT peer. To connect other peers to this one over the Internet,
+Oct 14 16:26:36.083 [INFO] Running a DHT instance. To connect other peers to this one, use
 use --initial_peers /ip4/1.2.3.4/tcp/1337/p2p/XXXX /ip4/1.2.3.4/udp/31337/quic/p2p/XXXX
 Oct 14 16:26:36.083 [INFO] Full list of visible multiaddresses: ...
 wandb: Currently logged in as: XXX (use `wandb login --relogin` to force relogin)
--- a/examples/albert/README.md
+++ b/examples/albert/README.md
@@ -28,7 +28,7 @@ Run the first DHT peer to welcome trainers and record training statistics (e.g.,
 ```
 $ ./run_training_monitor.py --wandb_project Demo-run
 Oct 14 16:26:36.083 [INFO] Running a DHT instance. To connect other peers to this one, use
-use --initial_peers /ip4/1.2.3.4/tcp/1337/p2p/XXXX /ip4/1.2.3.4/udp/31337/quic/p2p/XXXX
+ --initial_peers /ip4/1.2.3.4/tcp/1337/p2p/XXXX /ip4/1.2.3.4/udp/31337/quic/p2p/XXXX
 Oct 14 16:26:36.083 [INFO] Full list of visible multiaddresses: ...
 wandb: Currently logged in as: XXX (use `wandb login --relogin` to force relogin)
 wandb: Tracking run with wandb version 0.10.32
--- a/hivemind/hivemind_cli/run_dht.py
+++ b/hivemind/hivemind_cli/run_dht.py
@@ -48,8 +48,8 @@ def main():
     )
     parser.add_argument(
         "--identity_path",
-        help="Path to a pre-generated private key file. If defined, makes the peer ID deterministic. "
-        "May be generated using `./p2p-keygen` from `go-libp2p-daemon`.",
+        help="Path to a private key file. If defined, makes the peer ID deterministic. "
+             "If the file does not exist yet, writes a new private key to this file.",
     )
     parser.add_argument(
         "--refresh_period", type=int, default=30, help="Period (in seconds) for fetching the keys from DHT"
--- a/hivemind/hivemind_cli/run_dht.py
+++ b/hivemind/hivemind_cli/run_dht.py
@@ -49,7 +49,7 @@ def main():
     parser.add_argument(
         "--identity_path",
         help="Path to a private key file. If defined, makes the peer ID deterministic. "
-             "If the file does not exist yet, writes a new private key to this file.",
+        "If the file does not exist yet, writes a new private key to this file.",
     )
     parser.add_argument(
         "--refresh_period", type=int, default=30, help="Period (in seconds) for fetching the keys from DHT"
--- a/hivemind/hivemind_cli/run_dht.py
+++ b/hivemind/hivemind_cli/run_dht.py
@@ -49,7 +49,7 @@ def main():
     parser.add_argument(
         "--identity_path",
         help="Path to a private key file. If defined, makes the peer ID deterministic. "
-        "If the file does not exist yet, writes a new private key to this file.",
+        "If the file does not exist, writes a new private key to this file.",
     )
     parser.add_argument(
         "--refresh_period", type=int, default=30, help="Period (in seconds) for fetching the keys from DHT"
--- a/hivemind/hivemind_cli/run_dht.py
+++ b/hivemind/hivemind_cli/run_dht.py
@@ -1,0 +1,76 @@
+import time
+from argparse import ArgumentParser
+
+from hivemind.dht import DHT, DHTNode
+from hivemind.utils.logging import get_logger, use_hivemind_log_handler
+from hivemind.utils.networking import log_visible_maddrs
+
+use_hivemind_log_handler("in_root_logger")
+logger = get_logger(__name__)
+
+
+async def report_status(dht: DHT, node: DHTNode):
+    logger.info(
+        f"{len(node.protocol.routing_table.uid_to_peer_id) + 1} DHT nodes (including this one) "
+        f"are in the local routing table "
+    )
+    logger.debug(f"Routing table contents: {node.protocol.routing_table}")
+    logger.info(f"Local storage contains {len(node.protocol.storage)} keys")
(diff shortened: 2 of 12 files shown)


hivemind/utils/__init__.py
```diff
@@ -4,7 +4,14 @@
...
 from hivemind.utils.nested import *
-from hivemind.utils.networking import *
+from hivemind.utils.networking import (
+    Endpoint,
+    choose_ip_address,
+    get_free_port,
+    get_port,
+    log_visible_maddrs,
+    replace_port,
+)
```
mryab, line 14: Just as a reminder, the goal here is to incrementally move from wildcard imports to explicit enumeration of all exposed functions and classes. All public objects remain available for direct imports, but the higher we go in the import hierarchy, the less we want to expose (mostly the key functionality at the top level of hivemind, which also helps on autocompletion)

hivemind/utils/networking.py
```diff
@@ -74,3 +78,23 @@ def choose_ip_address(
...
+        available_ips = [Multiaddr(addr) for addr in visible_maddrs if "ip4" in addr or "ip6" in addr]
+        if available_ips:
+            preferred_ip = choose_ip_address(available_ips)
+            selected_maddrs = [addr for addr in visible_maddrs if preferred_ip in str(addr)]
+        else:
+            selected_maddrs = visible_maddrs
+        initial_peers_str = " ".join(str(addr) for addr in selected_maddrs)
+
+    _logger.info(
+        f"Running a DHT instance. To connect other peers to this one, use "
```
mryab, line 97: There was an "over the Internet" clause, but as we know, this works perfectly fine for local networks as well

mryab:
image
Example output of three `hivemind-dht` instances (two have a refresh period of 5 seconds, one has a default period of 30 seconds, but is launched with `GOLOG_LOGLEVEL=INFO HIVEMIND_LOGLEVEL=DEBUG`) and one DHT launched from the Python REPL for the purpose of adding custom keys

mryab:
The launch command and the first output lines for the first node are
```
~/hivemind$ hivemind-dht --refresh_period 5
Apr 21 02:09:37.060 [INFO] Running a DHT instance. To connect other peers to this one, use --initial_peers /ip4/172.27.77.70/tcp/43331/p2p/QmNMLCmjYUFgQ1TTkEv8P6kxQ3HYJHVhhWL6BH3nMLD3kW
Apr 21 02:09:37.060 [INFO] Full list of visible multiaddresses: /ip4/172.27.77.70/tcp/43331/p2p/QmNMLCmjYUFgQ1TTkEv8P6kxQ3HYJHVhhWL6BH3nMLD3kW /ip4/127.0.0.1/tcp/43331/p2p/QmNMLCmjYUFgQ1TTkEv8P6kxQ3HYJHVhhWL6BH3nMLD3kW
Apr 21 02:09:37.061 [INFO] 1 DHT nodes (including this one) are in the local routing table
Apr 21 02:09:37.061 [INFO] Local storage contains 0 keys
```

examples/albert/README.md
```diff
@@ -27,7 +27,7 @@ Run the first DHT peer to welcome trainers and record training statistics (e.g.,
 
 ```
 $ ./run_training_monitor.py --wandb_project Demo-run
-Oct 14 16:26:36.083 [INFO] Running a DHT peer. To connect other peers to this one over the Internet,
+Oct 14 16:26:36.083 [INFO] Running a DHT instance. To connect other peers to this one, use
 use --initial_peers /ip4/1.2.3.4/tcp/1337/p2p/XXXX /ip4/1.2.3.4/udp/31337/quic/p2p/XXXX
```
yhn112, line 31:
```suggestion
 --initial_peers /ip4/1.2.3.4/tcp/1337/p2p/XXXX /ip4/1.2.3.4/udp/31337/quic/p2p/XXXX
```

justheuristic (approved):

hivemind/utils/networking.py
```diff
@@ -1,14 +1,18 @@
...
 
 from multiaddr import Multiaddr
 
+from hivemind.utils.logging import TextStyle, get_logger
+
 Hostname, Port = str, int  # flavour types
 Endpoint = str  # e.g. 1.2.3.4:1337 or [2a21:6с8:b192:2105]:8888, https://networkengineering.stackexchange.com/a/9435
 LOCALHOST = "127.0.0.1"
 
+_logger = get_logger(__name__)
```
borzunov, line 14: fyi: It's named without the underscore in other parts of hivemind. You can leave it as is if you think that we should gradually move to `_logger` everywhere.

hivemind/hivemind_cli/run_dht.py
```diff
@@ -0,0 +1,76 @@
...
+        f"are in the local routing table "
+    )
+    logger.debug(f"Routing table contents: {node.protocol.routing_table}")
+    logger.info(f"Local storage contains {len(node.protocol.storage)} keys")
+    logger.debug(f"Local storage contents: {node.protocol.storage}")
+
+
+def main():
+    parser = ArgumentParser()
+    parser.add_argument(
```
borzunov, line 24: [Lots](https://github.com/learning-at-home/hivemind/blob/master/hivemind/p2p/p2p_daemon.py#L92) of other `P2P`/`DHT` params are not covered here, and we use some of them while testing libp2p connectivity features, etc.

What's your take on this? Should we add them gradually (when deemed necessary) or write code that automatically fetches their names, help lines, and defaults using reflexion?

In case of the former, I'm worried that the help lines & defaults would quickly become not synchronized.

borzunov (changes requested):
Thanks for the PR! I've left some discussion points below.

borzunov:
> In the future, we may add an interactive version of hivemind-dht similar to a database or key-value storage CLI client

One more nice-to-have thing for the future: make this script show a _really simple_ web interface showing the DHT contents. We don't need to use any frameworks for that (we could write it in the same way as `python -m http.server` works).

Such an interface would be great for educational purposes: if a student works with hivemind, they would be able to explore DHT/make sense out of it _before_ they learn how all the asyncio & libp2p stuff necessary to work with it in the code.

hivemind/utils/__init__.py
mryab: Kept two imports: the one that is used externally (`log_visible_maddrs`) and the one that is used in a couple of tests (`get_free_port`), since I want to restrict the scope of this PR. In follow-up PRs, we can actually move `get_free_port` to `test_utils`, since it is not used in any of hivemind-internal code

hivemind/utils/networking.py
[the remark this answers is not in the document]
mryab: Renamed (though at some point I guess I'll start renaming them for the sake of internal consistency — we totally shouldn't import loggers from internal modules)

hivemind/hivemind_cli/run_dht.py
```diff
@@ -0,0 +1,76 @@
...
+    logger.info(
+        f"{len(node.protocol.routing_table.uid_to_peer_id) + 1} DHT nodes (including this one) "
+        f"are in the local routing table "
+    )
+    logger.debug(f"Routing table contents: {node.protocol.routing_table}")
+    logger.info(f"Local storage contains {len(node.protocol.storage)} keys")
+    logger.debug(f"Local storage contents: {node.protocol.storage}")
+
+
+def main():
```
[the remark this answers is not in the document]
mryab: Added `tests/test_cli_scripts.py` with a basic test for `hivemind-dht`, we can also test `hivemind-server` similarly in the future

hivemind/hivemind_cli/run_dht.py
mryab: I guess that for the first PR that introduces `hivemind-dht`, we don't need to support all possible scenarios in which the DHT can be used. I deliberately kept it to a minimum to use the core functionality and took my inspiration from the ALBERT example; if you believe that some of the essential arguments are missing, I'd be happy to add them.

For the desynchronization of DHT docstrings, I'd not worry too much about it at first: I'm not aware of any imminent PRs that change the purpose of the given DHT arguments, and if somebody does change its API, I believe that at least one of us (me, you or @justheuristic) will have a look at the corresponding PR. Though it doesn't guarantee that we will notice a desync, the issue will be easily fixable, and adding extra abstraction at this point seems to me to be an overkill

hivemind/hivemind_cli/run_dht.py
```diff
@@ -0,0 +1,76 @@
...
+        "--use_ipfs",
+        action="store_true",
+        help='Use IPFS to find initial_peers. If enabled, you only need to provide the "/p2p/XXXX" '
+        "part of the multiaddrs for the initial_peers "
+        "(no need to specify a particular IPv4/IPv6 host and port)",
+    )
+    parser.add_argument(
+        "--identity_path",
+        help="Path to a pre-generated private key file. If defined, makes the peer ID deterministic. "
+        "May be generated using `./p2p-keygen` from `go-libp2p-daemon`.",
```
borzunov, line 52:
```suggestion
        help="Path to a private key file. If defined, makes the peer ID deterministic. "
             "If the file does not exist yet, writes a new private key to this file.",
```

This became out of sync :)

borzunov (approved):
Great! I have only one issue left.

PR #467, opened 2022-05-02
Pin the version of Golang for libp2p build

mryab:
Currently, the tests for the libp2p daemon build appear to fail/hang on master; this is most likely related to the version of Go not being pinned in the workflow environment. This PR fixes the issue by pinning the version of Golang to the latest supported by the current libp2p version and mentioning it in the README.


borzunov:
This is merged as a part of #466.

PR #468, opened 2022-05-17
use_local_updates in optimizer

justheuristic:
[this PR fixes some edge case bugs introduced since 1.1.0, those bugs do not affect pip version]
- fix use_local_updates in Optimizer, add test cases to ensure that it works
- ensure that grad_compression is always used

### DIFF
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -297,12 +297,49 @@ def test_progress_tracker():
 
 
 @pytest.mark.forked
+@pytest.mark.parametrize(
+    "use_local_updates, delay_state_averaging, delay_optimizer_step, delay_grad_averaging, reuse_grad_buffers",
+    [
+        (False, False, False, False, False),
+        (False, True, False, False, False),
+        (False, True, True, True, False),
+        (False, False, False, False, True),
+        (False, True, True, True, True),
+        (False, True, True, False, True),
+        (True, False, False, False, False),
+        (
+            True,
+            True,
+            False,
+            False,
+            False,
+        ),
+    ],
+)
 def test_optimizer(
+    use_local_updates: bool,
+    delay_state_averaging: bool,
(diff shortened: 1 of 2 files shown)


artek0chumak (approved):

tests/test_optimizer.py
```diff
@@ -391,6 +430,6 @@ def run_trainer(batch_size: int, batch_time: float, client_mode: bool):
     assert 4 / 0.3 * 0.8 <= optimizer.tracker.performance_ema.samples_per_second <= 4 / 0.3 * 1.2
 
     assert not optimizer.state_averager.is_alive()
-    assert not optimizer.grad_averager.is_alive()
     assert not optimizer.tracker.is_alive()
+    assert optimizer.grad_averager is None if use_local_updates else (not optimizer.grad_averager.is_alive())
```
mryab: Better to split the conditional in multiple lines:

mryab:
```suggestion
    if use_local_updates:
        assert optimizer.grad_averager is None
    else:
        assert not optimizer.grad_averager.is_alive()
```

tests/test_optimizer.py
```diff
@@ -297,12 +297,49 @@ def run_worker(index: int, batch_size: int, period: float, **kwargs):
...
+        (False, True, True, True, True),
+        (False, True, True, False, True),
+        (True, False, False, False, False),
+        (
+            True,
+            True,
+            False,
+            False,
+            False,
+        ),
```
mryab, line 316: To make formatting more consistent, you can surround the entire block of cases above with `# fmt: off` and `# fmt: on`

mryab (approved):

tests/test_optimizer.py
justheuristic: good point, thx

PR #469, opened 2022-05-23
Update p2pd to v0.3.8 (and libp2p to v0.17.0)

borzunov:
This PR includes refactoring of `setup.py` making it easier to include more binaries into the distribution. The current version of `p2p-keygen` can be easily reimplemented in Python, so we don't actually include it now (but keep the opportunity to do that in case if this util becomes more sophisticated).

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -12,8 +12,7 @@ jobs:
     strategy:
       matrix:
         python-version: [ 3.7, 3.8, 3.9 ]
-        iteration: [1, 2, 3, 4, 5]
-    timeout-minutes: 10
+    timeout-minutes: 12
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python
@@ -72,7 +71,7 @@ jobs:
   codecov_in_develop_mode:
 
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 12
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python


setup.py
```diff
@@ -14,10 +13,13 @@
 from setuptools.command.build_py import build_py
 from setuptools.command.develop import develop
 
-P2PD_VERSION = "v0.3.6"
-P2PD_CHECKSUM = "627d0c3b475a29331fdfd1667e828f6d"
```
borzunov, line 18: Keeping the checksums for both `p2pd` and `p2p-keygen` would be tiresome. On the other hand, it doesn't hurt if we download these binaries one more time in case if a user reinstalls hivemind.

setup.py
```diff
@@ -14,10 +13,12 @@
...
-LIBP2P_TAR_URL = f"https://github.com/learning-at-home/go-libp2p-daemon/archive/refs/tags/{P2PD_VERSION}.tar.gz"
-P2PD_BINARY_URL = f"https://github.com/learning-at-home/go-libp2p-daemon/releases/download/{P2PD_VERSION}/p2pd"
+P2PD_VERSION = "v0.3.0-hivemind.8"
```
mryab, line 16: Is there a reason for the `.8` suffix?

setup.py
```diff
@@ -14,10 +13,12 @@
...
+
+P2PD_SOURCE_URL = f"https://github.com/learning-at-home/go-libp2p-daemon/archive/refs/tags/{P2PD_VERSION}.tar.gz"
+P2PD_BINARY_URL = f"https://github.com/learning-at-home/go-libp2p-daemon/releases/download/{P2PD_VERSION}/"
+EXECUTABLES = ["p2pd", "p2p-keygen"]
```
mryab, line 20: To me, it's not entirely clear why we need `p2p-keygen` right now, especially since it can't be used in an installed version of hivemind by itself (it is not exposed via PATH)

setup.py
mryab: They weigh ~7-8MB each, which is quite a lot for a Python package installation (especially because it happens under the hood). Just curious, how do other Python libraries with significant binary deps (e.g. PyTorch or NumPy) handle this?

justheuristic (approved):

setup.py
justheuristic: It does hurt if user reinstalls hivemind in an environment that lacks internet access AKA exactly what we do in the internal testing infrastructure.

setup.py
mryab: In that case, perhaps we should support this in the current PR to avoid downloading 8MB binaries that cannot be used?

setup.py
borzunov: Initially, we've forked `v0.3.0` of the upstream and incremented the last number for each batch of hivemind-specific modifications. I'm not sure it's okay, since if the upstream ever updates to `v0.3.1` and further, it will collude with the versions denoting our modifications. Thus, I suggest to switch to the new versioning scheme with such "post-releases" (e.g., `v0.3.0-hivemind.8` instead of `v0.3.8`).

setup.py
borzunov: Exposing via `PATH` is indeed the next step, however, just having the binary is already *much* better than nothing.

Note that without this change, one needs to install Go, compile `p2p-keygen` and send it to the machine with the training monitor - all that just for launching a monitor that doesn't change its address when restarted.

setup.py
```diff
@@ -13,21 +13,28 @@
...
 from setuptools.command.develop import develop
+from setuptools.command.install import install
+
+P2PD_VERSION = "v0.3.0-hivemind.8"
+
+P2PD_SOURCE_URL = f"https://github.com/learning-at-home/go-libp2p-daemon/archive/refs/tags/{P2PD_VERSION}.tar.gz"
+P2PD_BINARY_URL = f"https://github.com/learning-at-home/go-libp2p-daemon/releases/download/{P2PD_VERSION}/"
+
+# The value is sha256 of the binary from the release page
+EXECUTABLES = {
```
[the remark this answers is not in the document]
borzunov: I thought about it but chose `EXECUTABLES` since the code uses only keys from this dict most of the time

setup.py
borzunov: [reverted after discussion with @mryab]

mryab (approved):

borzunov:
I'll increase the test time limit from 10 min to 12 min since some jobs don't finish in time (it seems that they are just ~20 sec late).

I think this is due to internal GitHub changes, since I see no regression in the number of successful tests if we compare [master](https://github.com/learning-at-home/hivemind/runs/6572226299) and [this PR](https://github.com/learning-at-home/hivemind/runs/6572442669) (in both cases, 13 of 16 runs are successful, 3 of 16 don't finish in time).

Issue #242, opened 2021-04-22
Convert hivemind.Server/RemoteModuleCall/RemoteCallMany to libp2p backend


justheuristic:
[depends on #238 to be merged ]
After we've implemented P2P transport with nat traversal, we should switch the main components to libp2p backend to take advantage of this new transport.

One of three main components is `hivemind.server.Server` and its counterpart `hivemind.client.RemoteExpert`

On a client side, hivemind creates a RemoteExpert pytorch module that calls experts via [`_RemoteModuleCall`](https://github.com/learning-at-home/hivemind/blob/master/hivemind/client/expert.py#L70-L104) (and [`_RemoteCallMany`](https://github.com/learning-at-home/hivemind/blob/master/hivemind/client/moe.py#L153-L162) for DMoE)

A server receives incoming connections with several [`ConnectionHandler`](https://github.com/learning-at-home/hivemind/blob/master/hivemind/server/connection_handler.py) processes running in parallel. These processes run gRPC servers and hence should be switched to libp2p.

* Checklist
   * [x] find some way to attach several processes to one RPC (as in server/connection_handler.py)
   * [ ] make sure it passes `tests/test_moe.py`
   * [ ] make sure it passes `tests/test_training.py`
   * [x] tune performance in `tests/benchmark_througphput.py`

GreenFatGuy:
working on it

justheuristic:
Current status:
- we're broken DHT get_experts (or declare experts) capability, gotta fix it
- there's some error when dealing with large messages (@deniskamazur PLZ add an example that fails, e.g. for benchmark_throughput)
- gotta merge changes from master

GreenFatGuy:
Current progress:

* merge of `master` is done
* dht bug is fixed, the problem was that dht had not enough data to initialize RemoteExpert properly + it should be constructed on client side now
* working on fixes for big benchmarks: the plan is to split data into partitions.

GreenFatGuy:
Current status:

Implemented streaming for backward and forward for moe/expert/RemoteModuleCall. Now we split all inputs and outputs for forward/backward into smaller parts before sending, then stream this parts and assemble again on the other end. There is no problem with big batches now.

In plans:
 * fix tests for new moe expert/server on p2p
 * implement variance in forward/backward call: if we can - do non-streaming version (old one), if don't - new one with streaming. This could possibly increase throughput (do experiment).
 * try different architectures. I need a server for it, because my machine can not do this

[...]

PR #470, opened 2022-05-23
Convert hivemind.server to libp2p backend

GreenFatGuy:
#242 
In this PR we are trying to get rid of GRPC in MoE module of hivemind. I do this PR draft now to be able to start review process and do it piece by piece, because some core thing are done and some results are achieved (I believe). There is still work to be done (it will be mentioned at the end of this message).

## What is currently done:
  - `RemoteExperts` are able to communicate through libp2p
  - Some throughput performance optimizations were done: we achieved ~2GiB/sec at `ffn_forward` benchmark using GTX3060. It is done by enabling balancing multiple handlers of the same protocol in `p2pd`. Thus we can do packing/unpacking of messages on forward/backward in parallel. Results are at the end of the message
  - Examples from hivemind tutorial are working
  - `tests/test_dht_experts.py` are passing
  
  
## What are topics to discuss:
 
- [x] Current implementation of `RemoteExpert` is quite heavy compared to GRPC version. GRPC version was in fact a few string fields containing endpoints, nothing more. Current version contains heavy object: connection to p2p-daemon, which is not serializable. Probably this is not the best decision.
- [x] Current `moe/server/Server` will not probably work if it has no `DHT` value inside, however old api documentation says that `DHT` is not mandatory for `Server` instance. There are two ways I see: make `DHT` mandatory or create a `P2P` instance inside a `Server`. Doing second option `DHT` inside `Server` probably might not be useful anymore

## What is yet to be done:
- [x] Separate `RemoteExpertInfo` with endpoints from `RemoteExpert` and make function to make expert from its info 
- [x] Separate thread and queue for `async` actions from `_RemoteModuleCall` 
- [x] `tests/test_moe.py` and `tests/test_training.py`. And probably some other tests.
- [x] Rewrite `RemoteMixtureOfExpert` to the `P2P` (as `RemoteExpert`)
- [x] Refactoring of `Server` instance after discussion mentioned above
- [x] Wildly benchmark every possible scenario
- [ ] FIx documentation for tutorials 
- [ ] Add test on scenarios, which might be not covered
~~After this PR is done we have to discuss `dht.replicate_p2p()` because it is not fork-safe and not in an obvious way~~

## Current benchmarks results:
Current benchmarks were performed with GTX3060 and `--preset ffn_forward`. Each experiment was performed at least 5 times and results bellow are average. On demand I can provide more detailed data. Also worth mentioning that this results can change during review process.

| Branch         | Batch size | Number of handlers | Throughput      |
| ---------------- | -------------- | -------------------------- | -------------------- |
| `server-p2p` | 1024         | 1                               | ~838 MiB/sec   |
| `server-p2p` | 1024         | 5                               | ~2085 MiB/sec |
| `server-p2p` | 1024         | 10                             | ~2055 MiB/sec |
| `master`       | 1024         | *`default`                   | ~1526 MiB/sec |
| `master`       | 2048         | *`default`                   | ~2248 MiB/sec |

*`default` in column  `Number of handlers`: TLDR it is 64. 
  It means formula `max(1, num_handlers or num_clients // 2)` where `num_handlers = None` and num_clients = 128`. 
  
  
  ## What can be done after merging this

Thing discovered during review. They are not blocking this PR, but it is better to do them.

- [ ] Get rid of multiaddrs everywhere. P2P daemon should be able to communicate using `PeerID` only
- [ ] In some places there are CPU-bound thing happening inside async task. It is better to move them into thread executors. For example  `hivemind/moe/client/expert.py` forward/Backward
- [ ] currently hivemind.Server does not check that inputs are correct. If user sends malformed inputs, it may OOM the server. We should check for that in some future PR. See #3 
- [ ] if clients sends tensor of shape [0, 123], it will be split into zero messages and uid will not be passed. Server will receive uid=None and fail with cryptic KeyError(None). We should either forbid this on client side or ensure that zero-element tensors are serialized into a stream with first emty message.
- [ ] Test load balancing for unary handlers on python side


borzunov:
[Renamed it in spirit of #296 and #323.]

justheuristic:
Gentle reminder: @borzunov , please review, we badly need this code in master!

mryab:
Perhaps it's best to wait for CI to pass and for the PR to be rebased to master — right now, these are the obvious problems that would prevent us from merging it

hivemind/dht/dht.py
```diff
@@ -55,6 +55,7 @@ def __init__(
         **kwargs,
     ):
         self._parent_pid = os.getpid()
+        self._my_pid = os.getpid()
```
borzunov, line 58: nit:

```suggestion
        self._origin_pid = os.getpid()
```

(for consistency with [MPFuture](https://github.com/learning-at-home/hivemind/blob/master/hivemind/utils/mpfuture.py#L98))

hivemind/hivemind_cli/run_server.py
```diff
@@ -18,8 +18,7 @@ def main():
     # fmt:off
     parser = configargparse.ArgParser(default_config_files=["config.yml"])
     parser.add('-c', '--config', required=False, is_config_file=True, help='config file path')
-    parser.add_argument('--listen_on', type=str, default='0.0.0.0:*', required=False,
-                        help="'localhost' for local connections only, '0.0.0.0' for ipv4 '[::]' for ipv6")
+
```
borzunov, line 21: Let's add `--host_maddrs` and `--announce_maddrs` arguments forwarded to `hivemind.DHT`/`hivemind.P2P`, like we have in [examples/albert](https://github.com/learning-at-home/hivemind/blob/master/examples/albert/arguments.py#L26).

hivemind/moe/__init__.py
```diff
@@ -1,5 +1,6 @@
 from hivemind.moe.client import RemoteExpert, RemoteMixtureOfExperts, RemoteSwitchMixtureOfExperts
 from hivemind.moe.server import (
+    ConnectionHandler,
```
borzunov, line 3:
```suggestion
```

We usually only export "the most public" classes to the root of the `hivemind` namespace (i.e., the classes most useful for a developer that uses hivemind).

Is there a reason to export `ConnectionHandler`? I thought this was an internal class.

hivemind/moe/server/__init__.py
```diff
@@ -1,3 +1,4 @@
+from hivemind.moe.server.connection_handler import ConnectionHandler
```
borzunov, line 1:
```suggestion
```

See the explanation in `moe/__init__.py`.

benchmarks/benchmark_throughput.py
```diff
@@ -229,7 +247,11 @@ def benchmark_throughput(
         )
     elif args.preset == "minimalistic":
         benchmark_throughput(
-            num_experts=1, num_clients=1, num_handlers=1, num_batches_per_client=args.num_batches_per_client
+            num_experts=1,
+            num_clients=1,
+            num_handlers=1,
+            num_batches_per_client=args.num_batches_per_client,
+            batch_size=1024,
```
borzunov, line 254: Is changing `batch_size` here intended?

benchmarks/benchmark_throughput.py
```diff
@@ -6,13 +6,13 @@
 
 import torch
 
-from hivemind.moe.client import RemoteExpert
-from hivemind.moe.server import ExpertBackend, Server
-from hivemind.moe.server.layers import name_to_block
+import hivemind
+from hivemind import P2P
+from hivemind.dht import DHT
```
borzunov, line 11: nit:

```suggestion
```

I'd suggest to:

- Either keep `import hivemind`, use `hivemind.DHT` and `hivemind.P2P` (i.e., apply this suggestion), and keep the "from imports" only for the long module names. This is good because it minimizes the number of imports.
- Or remove `import hivemind` and use the "from imports" only, like it was before this PR. This is good because it minimizes this file's diff.

Mixing both of the mentioned import styles is rather confusing.

hivemind/moe/server/server.py
```diff
@@ -344,15 +331,15 @@ def _server_runner(pipe, *args, **kwargs):
...
+        dht_maddrs = server.dht.get_visible_maddrs()
+        pipe.send((True, (server.dht.peer_id, dht_maddrs)))
         pipe.recv()  # wait for shutdown signal
 
     finally:
         logger.info("Shutting down server...")
         server.shutdown()
         server.join()
-        logger.info("Server shut down")
+        logger.info("Server shut down.")
```
borzunov, line 342: nit: We usually don't use trailing dots in the log messages (e.g., see #419). Please remove them.

hivemind/moe/server/server.py
```diff
@@ -332,7 +319,7 @@ def background_server(*args, shutdown_timeout=5, **kwargs) -> Tuple[Endpoint, Li
         if runner.is_alive():
             logger.info("Server failed to shutdown gracefully, terminating it the hard way...")
             runner.kill()
-            logger.info("Server terminated")
+            logger.info("Server terminated.")
```
borzunov, line 322: nit: We usually don't use trailing dots in the log messages (e.g., see #419). Please remove them.

tests/test_dht_experts.py
```diff
@@ -59,11 +60,11 @@ def test_beam_search(
         {"expert." + ".".join([str(random.randint(0, dim - 1)) for dim in grid_dims]) for _ in range(total_experts)}
     )
     for batch_start in range(0, len(real_experts), batch_size):
+        dht_ = random.choice(dht)
```
borzunov, line 63: nit: Having both `dht` and `dht_` is confusing :) Let's rename them to `dht_instances` (a list) and `dht` (one instance) for clarity.

borzunov (changes requested):
Great work!

I've left several minor comments for now, but I also have a couple of more significant issues to discuss.

benchmarks/benchmark_throughput.py
GreenFatGuy: This change was needed for checking inputs/outputs streaming in forward/backward. We can change it back to 512 if you want. I don't think there is any effect.

hivemind/moe/__init__.py
GreenFatGuy: I think there is no reason. I'll remove it

.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
+
+This directory caches those eggs to prevent repeated downloads.
+
+However, it is safe to delete this directory.
+
```
borzunov, line 6: This file seems to be added by accident.

benchmarks/benchmark_throughput.py
GreenFatGuy: Just looked at `master`: there is default `batch_size` for `minimalistic` benchmark. I made it the same here

benchmarks/benchmark_throughput.py
```diff
@@ -31,14 +31,26 @@ def print_device_info(device=None):
         logger.info(f"Cached:   {round(torch.cuda.memory_cached(0) / 1024 ** 3, 1)} GB")
 
 
-def client_process(can_start, benchmarking_failed, port, num_experts, batch_size, hid_dim, num_batches, backprop=True):
+def client_process(
+    can_start,
+    benchmarking_failed,
+    server_peer_info,
```
borzunov, line 37:
```suggestion
    server_maddrs,
```

benchmarks/benchmark_throughput.py
```diff
@@ -31,14 +31,26 @@ def print_device_info(device=None):
...
+    hid_dim,
+    num_batches,
+    backprop=True,
+) -> None:
     torch.set_num_threads(1)
     can_start.wait()
-    experts = [RemoteExpert(f"expert{i}", endpoint=f"{LOCALHOST}:{port}") for i in range(num_experts)]
+
+    p2p = RemoteExpertWorker.run_coroutine(P2P.create())
+    RemoteExpertWorker.run_coroutine(p2p._client.connect(server_peer_info.peer_id, server_peer_info.addrs))
```
borzunov, line 48:
```suggestion
    p2p = RemoteExpertWorker.run_coroutine(P2P.create(initial_peers=server_maddrs))
```

benchmarks/benchmark_throughput.py
```diff
@@ -59,77 +71,81 @@ def benchmark_throughput(
...
     timestamps = dict(started=time.perf_counter())
 
     try:
-        # start clients and await server
-        # Note: client processes must be launched BEFORE touching gpu, even torch.cuda.is_available can cause trouble
+        server_dht = DHT(start=True)
+        server_dht_peer_info = PeerInfo(
+            peer_id=server_dht.peer_id,
+            addrs=[addr.decapsulate("/p2p/" + addr.get("p2p")) for addr in server_dht.get_visible_maddrs()],
+        )
```
borzunov, line 92:
```suggestion
```

benchmarks/benchmark_throughput.py
```diff
@@ -59,77 +71,81 @@ def benchmark_throughput(
...
+
         clients = [
             mp.Process(
                 target=client_process,
                 name=f"client_process-{i}",
                 args=(
                     can_start,
                     benchmarking_failed,
-                    port,
+                    server_dht_peer_info,
```
borzunov, line 101:
```suggestion
                    server_dht.get_visible_maddrs(),
```

benchmarks/benchmark_throughput.py
```diff
@@ -229,7 +245,10 @@ def benchmark_throughput(
         )
     elif args.preset == "minimalistic":
         benchmark_throughput(
-            num_experts=1, num_clients=1, num_handlers=1, num_batches_per_client=args.num_batches_per_client
+            num_experts=1,
```
borzunov, line 248: nit: I'd suggest to revert such meaningless changes to minimize the diff

borzunov (changes requested):
Due to the manual parsing of maddrs, the benchmark code is overly complex. In fact, there's no need to parse the addresses and use the private fields (like in `p2p._client.connect()`) - one can just pass `server_dht.get_visible_maddrs()` like `initial_peers`.

hivemind/utils/grpc.py
```diff
@@ -208,3 +223,28 @@ def combine_from_streaming(stream: Iterable[runtime_pb2.Tensor]) -> runtime_pb2.
         buffer_chunks.append(tensor_part.buffer)
     serialized_tensor.buffer = b"".join(buffer_chunks)
     return serialized_tensor
+
+
+RpcMessage = TypeVar("RpcMessage")
+
+
+async def gather_from_grpc(
```
borzunov, line 231: When this PR is ready, is it true that we won't use grpc anywhere and this function gonna disappear?

hivemind/utils/networking.py
```diff
@@ -1,12 +1,16 @@
...
 
 from multiaddr import Multiaddr
 
 Hostname, Port = str, int  # flavour types
-Endpoint = str  # e.g. 1.2.3.4:1337 or [2a21:6с8:b192:2105]:8888, https://networkengineering.stackexchange.com/a/9435
+Endpoint = (
+    Tuple[  # e.g. 1.2.3.4:1337 or [2a21:6с8:b192:2105]:8888, https://networkengineering.stackexchange.com/a/9435
+        str, Tuple[str, ...]
+    ]
+)
```
borzunov, line 13: Ideally, we should always use `PeerID` instead of `Endpoint`, so libp2p will choose a route to the target peer by itself (as I explained during our call).

If we postpone this change to the next PR, we still don't such an `Endpoint` tuple since it duplicates the existing `hivemind.PeerInfo` structure (but worse since it doesn't have the named fields).

Also, the comment here became irrelevant.

borzunov, line 13:
```suggestion
```

Ideally, we should always use `PeerID` instead of `Endpoint`, so libp2p will choose a route to the target peer by itself (as I explained during our call).

If we postpone this change to the next PR, we still don't such an `Endpoint` tuple since it duplicates the existing `hivemind.PeerInfo` structure (but worse since it doesn't have the named fields).

Also, the comment here became irrelevant.

hivemind/hivemind_cli/run_server.py
```diff
@@ -32,6 +31,11 @@ def main():
                         help="expert type from test_utils.layers, e.g. 'ffn', 'transformer', 'det_dropout' or 'nop'")
     parser.add_argument('--hidden_dim', type=int, default=1024, required=False, help='main dimension for expert_cls')
 
+    parser.add_argument('--host_maddrs', type=list, nargs='+', default=['/ip4/0.0.0.0/tcp/0'], required=False,
+                        help='Multiaddrs to listen for external connections from other p2p instances; default: all IPv4 and TCP: /ip4/0.0.0.0/tcp/0')
+    parser.add_argument('--announce_maddrs', type=list, nargs='+', default=None, required=False,
+                        help='Visible multiaddrs the host announces for external connections from other p2p instances')
+
```
borzunov, line 38: Now let's ensure that these args are passed into `hivemind.DHT` inside `hivemind.Server`. Probably, we need to pass them along with `initial_peers` [here](https://github.com/learning-at-home/hivemind/blob/master/hivemind/moe/server/server.py#L158).

hivemind/moe/client/expert.py
```diff
@@ -89,15 +200,51 @@ def forward(
...
+                    lambda t: runtime_pb2.ExpertRequest(
+                        uid=ctx.uid,
+                        tensors=[
+                            t,
+                        ],
+                    ),
+                    as_aiter(*split),
+                ),
+            )
+        )
```
borzunov, line 230: This looks like black got mad and reformatted the code to be written vertically :) Let's write it in a more readable way: 

```suggestion
        outputs = RemoteExpertWorker.run_coroutine(
            stub.rpc_forward_partial(
                amap_in_executor(
                    lambda t: runtime_pb2.ExpertRequest(uid=ctx.uid, tensors=[t]),
                    as_aiter(*split),
                ),
            )
        )
```

Please also fix the same fragment in `backward_partial()`.

hivemind/moe/server/connection_handler.py
```diff
@@ -24,59 +28,118 @@ class ConnectionHandler(mp.context.ForkProcess):
...
+            for p in split_for_streaming(t, DEFAULT_MAX_MSG_SIZE // 2)
         ]
 
-        return runtime_pb2.ExpertResponse(tensors=serialized_response)
+        async for part in as_aiter(*output_split):
+            yield runtime_pb2.ExpertResponse(
+                tensors=[
+                    part,
+                ],
+            )
```
borzunov, line 118: nit:

```suggestion
            yield runtime_pb2.ExpertResponse(tensors=[part])
```

Please do the same in `rpc_backward_partial()` below :)

justheuristic:
Minor request: it would be great to rename forward_partial and backward_partial into forward_stream / backward_stream.

Rationale: _partial can be associated with incremental computation, e.g. one token at a time, as in https://fast-transformers.github.io/recurrent_transformers/

It's okay not to implement that if you have any objections, but please explain your motivation by replying to this thread.

benchmarks/benchmark_throughput.py
GreenFatGuy: This is the result of `black` work :)
Should we still left it as one line?

hivemind/utils/grpc.py
GreenFatGuy: yes. You want to wipe out  `grpc` from the all namings?

hivemind/utils/networking.py
GreenFatGuy: What is your suggestion on `Endpoint`? I think we can delete it, because now we have `PeerInfo`, in future we want to have only `PeerID`. Both are already exiting structures

hivemind/hivemind_cli/run_server.py
GreenFatGuy: We are already passing them. `Server` starts as `Server.create(**args, ........)`. I added `**kwargs` to the `Server.create` which are further passed to the `DHT`, which passes it into `P2P`.

Maybe it is not the best way, but I saw how this arguments are passed from `DHT` into `P2P` (it is done through `**kwargs`) and done the same.

hivemind/utils/grpc.py
mryab: At least from this one, since it's introduced in this PR :)

Also, is it true that we can generalize this and [accumulate_parts_streaming](https://github.com/learning-at-home/hivemind/blob/master/hivemind/averaging/allreduce.py#L335) in some way, since they both deserialize a series of tensors received over the network asynchronously?

hivemind/p2p/p2p_daemon_bindings/p2pclient.py
```diff
@@ -61,8 +61,8 @@ async def listen(self) -> AsyncIterator["Client"]:
         async with self.control.listen():
             yield self
 
-    async def add_unary_handler(self, proto: str, handler: TUnaryHandler):
-        await self.control.add_unary_handler(proto, handler)
+    async def add_unary_handler(self, proto: str, handler: TUnaryHandler, balanced: bool = False):
+        await self.control.add_unary_handler(proto, handler, balanced=balanced)
```
mryab, line 65: Perhaps I couldn't find it, but can you please show how the `balanced` argument is eventually used? Right now I can only see that it is passed around the code with no obvious reason

mryab (changes requested):
Thanks for the PR! I've left a couple of initial comments, some of which should help us reduce the diff without any significant functional changes. Others are a bit more difficult to change (e.g. the RemoteExpertWorker class): if you feel that getting rid of them will be too long, we can proceed with them and refactor the code in subsequent PRs

Also, please update the [MoE guide in docs](https://github.com/learning-at-home/hivemind/blob/master/docs/user/moe.md), because it is very likely to become outdated with this PR

benchmarks/benchmark_throughput.py
borzunov: Yeah, this line shouldn't be changed if you use `black` of the same version as specified in `requirements.txt` :)

hivemind/utils/grpc.py
borzunov: While we've agreed with @justheuristic that wiping out all the grpc stuff can be done in a future PR, let's avoid introducing new grpc stuff here :)

hivemind/utils/networking.py
borzunov: Yeah, I suggest to delete it and use `PeerInfo` instead :) Replacing `PeerInfo` to `PeerID` can be done in a future PR (I can also do it myself, so you don't get overwhelmed) :)

hivemind/dht/dht.py
```diff
@@ -55,6 +55,7 @@ def __init__(
         **kwargs,
     ):
         self._parent_pid = os.getpid()
+        self._origin_pid = os.getpid()
```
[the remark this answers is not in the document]
borzunov: Yes, it's necessary below for the correct work of `DHT.replicate_p2p()` after forks.

borzunov:
As agreed with @justheuristic and @mryab, we're okay with the following important issues postponed to future PRs:

- Using `PeerID`s only for addressing peers anywhere besides `initial_peers` (no multiaddrs and `PeerInfo`/`Endpoint` structures)
- Removing all the grpc stuff, including deps in `requirements.txt` and service definitions in `*.proto` files

GreenFatGuy:
JFYI, I am currently finishing with fixing test (only a couple of them left broken in `test_training.py`). Also I will fix all issues mentioned above and push all my local commits. Hopefully pipeline will become green and I will be able to mark PR as ready.

benchmarks/benchmark_throughput.py
GreenFatGuy:
```
>>> black --version
<<< black, 22.3.0 (compiled: yes)
```

The only command for formatting I do is `black ./`.
Tried it just right now - `black` changes it to the current variant

hivemind/utils/grpc.py
GreenFatGuy: Honestly I don't know right now, because I didn't touch `averaging` in this PR, so I don't know much about it.
On the first glance `_accumulate_parts_streaming` does something more complex than just gathering parts together

hivemind/hivemind_cli/run_server.py
GreenFatGuy: Is it ok to pass this arguments through `**kwargs`?

hivemind/moe/client/expert.py
GreenFatGuy: To reuse code which decides how to comuncate with server (via streaming or at once) in `RemoteMixtureOfExperts` I made this methods as free functions. Please, review this code once more if it looks OK now or do I have to make some changes

hivemind/moe/client/expert.py
```diff
@@ -89,15 +200,51 @@ def forward(
...
+                deserialized_outputs = cls.forward_partial(serialized_tensors, ctx, stub)
+                break
+        else:
+            deserialized_outputs = cls.forward_oneshot(serialized_tensors, ctx, stub)
 
         return tuple(deserialized_outputs)
 
-    @staticmethod
+    @classmethod
+    def forward_partial(cls, serialized_tensors: List[runtime_pb2.Tensor], ctx, stub) -> List[torch.Tensor]:
```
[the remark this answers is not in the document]
GreenFatGuy: they are free functions know. Also I made them private inside module. Please, look at this code again and tell me if there is something you do not agree with

hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -128,6 +129,12 @@ def from_protobuf(cls, peer_info_pb: p2pd_pb2.PeerInfo) -> "PeerInfo":
         addrs = [Multiaddr(addr) for addr in peer_info_pb.addrs]
         return PeerInfo(peer_id, addrs)
 
+    @classmethod
+    def from_endpoint(cls, endpoint: Endpoint) -> "PeerInfo":
+        peer_id = PeerID.from_base58(endpoint[0])
+        addrs = [Multiaddr(a) for a in endpoint[1]]
+        return PeerInfo(peer_id, addrs)
```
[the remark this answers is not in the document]
GreenFatGuy: Eventually yes. I renamed it into `from_tuple` since `Endpoint` is removed. This method needed in call `moe/server/dht_handler/get_experts` to convert dht values, which are sotred as tuples of strings into `PeerInfo`.

hivemind/utils/networking.py
GreenFatGuy: UPD: returned `utils/networking.py` back. Turned out it should be done more carefully, if should be at all.

GreenFatGuy:
@borzunov @mryab @justheuristic Just pushed some major changes. CI should be green now (if not - I will fix it tomorrow).

I left some issues unresolved where I was not sure that I made as you asked. Please look at them again and mark as resolved if you think it is.

Sorry for that you have to review some parts again. I am ready to discuss this PR as whole if you think we need this.

There is still some debt on me: it is documentation about [MoE guide](https://github.com/learning-at-home/hivemind/blob/master/docs/user/moe.md), however this is supplementary to the code changes. Hopefully I will

benchmarks/benchmark_throughput.py
borzunov: What if you run it from the root directory of the repo? My guess is that it fails to read the max line length setting from [pyproject.toml](https://github.com/learning-at-home/hivemind/blob/master/pyproject.toml#L7).

hivemind/utils/networking.py
```diff
@@ -33,7 +33,6 @@ def strip_port(endpoint: Endpoint) -> Hostname:
 def get_free_port(params=(socket.AF_INET, socket.SOCK_STREAM), opt=(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)):
     """
     Finds a tcp port that can be occupied with a socket with *params and use *opt options.
-
```
borzunov, line 36: Please keep the newlines in this files for the sake of readability.

benchmarks/benchmark_throughput.py
borzunov: I see, that's fine :)

hivemind/utils/networking.py
borzunov: Good catch! Let's remove it along with all the grpc stuff, we can do this in a separate PR :)

hivemind/p2p/p2p_daemon_bindings/datastructures.py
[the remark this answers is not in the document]
borzunov: I think we would be able to remove this method in the next PR that will get rid of sending multiaddrs. No need to worry about it in this PR.

hivemind/utils/grpc.py
```diff
@@ -208,3 +223,28 @@ def combine_from_streaming(stream: Iterable[runtime_pb2.Tensor]) -> runtime_pb2.
         buffer_chunks.append(tensor_part.buffer)
     serialized_tensor.buffer = b"".join(buffer_chunks)
     return serialized_tensor
+
+
+StreamMessage = TypeVar("StreamMessage")
+
+
+async def gather_from_streaming(
```
borzunov, line 231: In a future PR, we should remove `grpc.py` and move this method to a new file (e.g., `hivemind/utils/streaming.py`). But we can keep it as is in this PR :)

hivemind/moe/client/switch_moe.py
```diff
@@ -80,7 +80,6 @@ def forward(self, input: torch.Tensor, *args: torch.Tensor, **kwargs: torch.Tens
 
         # Compute scores, find most appropriate experts with beam search
         grid_scores = self.proj(input_for_gating).split_with_sizes(self.beam_search.grid_size, dim=-1)
-
```
borzunov, line 83: nit: Let's keept the newlines in this file to minimize the diff.

hivemind/moe/server/server.py
```diff
@@ -313,14 +302,14 @@ def shutdown(self):
 
 
 @contextmanager
-def background_server(*args, shutdown_timeout=5, **kwargs) -> Tuple[Endpoint, List[Multiaddr]]:
-    """A context manager that creates server in a background process, awaits .ready on entry and shuts down on exit"""
+def background_server(*args, shutdown_timeout=5, **kwargs) -> PeerInfo:
+    """A context manager that creates server in a background thread, awaits .ready on entry and shutdowns on exit"""
```
borzunov, line 306:
```suggestion
    """A context manager that creates server in a background thread, awaits .ready on entry and shuts down on exit"""
```

nit: `shutdown` is a noun, `shut down` is a verb

borzunov:
@GreenFatGuy According to codecov, some things in this PR are not covered with tests (e.g., streaming-related code and `connection_handler.py`). I am not talking about the isolated codecov warnings (those are okay), but about files with lots of such warnings.

If it's not a codecov error, can we add some tests covering this functionality? This way, we'd be sure that changes in this PR and future PRs are not breaking the code you've written.

GreenFatGuy:
Sure. I have an opened task for this PR in the description above.

> Add test on scenarios, which might be not covered

hivemind/moe/client/moe.py
```diff
@@ -95,7 +101,7 @@ def forward(self, input: torch.Tensor, *args: torch.Tensor, **kwargs: torch.Tens
         chosen_experts: List[List[RemoteExpert]] = self.beam_search.batch_find_best_experts(
             [scores.detach().cpu().numpy() for scores in grid_scores], self.k_best
         )
-
+        print(chosen_experts)
```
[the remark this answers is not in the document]
GreenFatGuy: This is just an artifact of the nightly debug and a rush to have a green CI =)

hivemind/utils/grpc.py
GreenFatGuy: We can do this particular change right now. It seems very easy to do

hivemind/hivemind_cli/run_server.py
borzunov: Oops, sorry, I looked at the wrong branch back then :) It's all good, I'm resolving this!

benchmarks/benchmark_throughput.py
GreenFatGuy: It just got fixed now. Don't know what was the case

borzunov (changes requested):
Thanks a lot for addressing all our comments :)

Here's another batch of minor issues. Overall, I'm close to pressing "Approve" :)

tests/test_p2p_daemon_bindings.py
```diff
@@ -566,7 +566,13 @@ async def handle_another_proto(stream_info, reader, writer):
     async def handler_third(stream_info, reader, writer):
         event_third.set()
 
-    await p2pcs[1].stream_handler(another_proto, handler_third)
+    # p2p raises now for doubled stream handlers
+    with pytest.raises(ControlFailure):
+        await p2pcs[1].stream_handler(another_proto, handler_third)
+
+    # add in balanced mode, know handler should be placed in round robin queue
+    # also it should be next to be called
```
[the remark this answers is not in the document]
GreenFatGuy: sorry, don't quite understand what else do you mean. Changes around the lines mentioned in this comment are testing balanced mode: 
- no flag adding same protocol handler causes an exception
- adding same protocol with flag is fine
- both handlers are called (calls are balanced)

hivemind/moe/client/expert.py
```diff
@@ -1,43 +1,72 @@
...
+from hivemind.utils.streaming import gather_from_streaming, split_for_streaming
 
 DUMMY = torch.empty(0, requires_grad=True)  # dummy tensor that triggers autograd in RemoteExpert
 
 
-def _get_expert_stub(endpoint: Endpoint, *extra_options: Tuple[str, Any]):
-    """Create a gRPC stub to access remote expert or use previously created stub from a process-wide cache"""
-    channel_options = (("grpc.max_send_message_length", -1), ("grpc.max_receive_message_length", -1)) + extra_options
-    return ChannelCache.get_stub(endpoint, runtime_grpc.ConnectionHandlerStub, aio=False, options=channel_options)
+def _get_expert_stub(p2p: P2P, server_peer_info: PeerInfo):  # -> ConnectionHandlerStub:
```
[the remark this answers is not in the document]
GreenFatGuy: my IDE (VSCode with plugins) do not really like this changes, but it seems to work

hivemind/moe/client/expert.py
```diff
@@ -1,43 +1,72 @@
...
+from dataclasses import dataclass
+from queue import Queue
+from threading import Thread
+from typing import Any, Awaitable, Dict, Iterable, List, Optional, Sequence, Tuple, Union
 
 import torch
 import torch.nn as nn
 from torch.autograd.function import once_differentiable
 
+import hivemind
```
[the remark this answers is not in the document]
borzunov: Oops, you're right, my suggestion is actually making the circular import issues worse. Let's keep it as is then :)

As for `dht_handler.py`, I think `hivemind/moe/server` is a good place for it since it's an implementation detail of moe.server (not used by anything else in hivemind).

tests/test_p2p_daemon_bindings.py
```diff
@@ -560,13 +560,19 @@ async def handle_another_proto(stream_info, reader, writer):
 
     writer.close()
 
-    # test case: registering twice can override the previous registration
+    # test case: registering twice can not override the previous registration without balanced flag
```
borzunov, line 563:
```suggestion
    # test case: registering twice can't override the previous registration without balanced flag
```

nit: In English, it's either `cannot` or `can't` :)

borzunov (approved):
Please finish covering the new code with tests, then it's good to go :)

hivemind/moe/client/expert.py
```diff
@@ -52,18 +81,171 @@ def forward(self, *args, **kwargs):
...
+    )
+
+    return await gather_from_streaming(grad_inputs, lambda r: r.tensors, deserialize_torch_tensor)
+
+
+async def _backward_unary(uid: str, serialized_tensors: Iterable[runtime_pb2.Tensor], stub) -> List[torch.Tensor]:
+    grad_inputs: runtime_pb2.ExpertResponse = await stub.rpc_backward(
+        runtime_pb2.ExpertRequest(uid=uid, tensors=list(serialized_tensors))
+    )
+    return [deserialize_torch_tensor(t) for t in grad_inputs.tensors]
```
mryab, line 199: Side note: it's generally best to avoid CPU-heavy work inside asyncio, since it stalls the event loop. Thus, we need to be extremely careful with such constructions and place them in separate processes: otherwise, we lose nearly all the profit from asynchronous execution

hivemind/moe/client/expert.py
```diff
@@ -52,18 +81,171 @@ def forward(self, *args, **kwargs):
...
+    outputs: runtime_pb2.ExpertResponse = await stub.rpc_forward(
+        runtime_pb2.ExpertRequest(uid=uid, tensors=list(serialized_tensors))
+    )
+    return [deserialize_torch_tensor(t) for t in outputs.tensors]
+
+
+async def expert_forward(uid: str, inputs: Sequence[torch.Tensor], compressions: Iterable, stub) -> List[torch.Tensor]:
+    serialized_tensors = (
+        serialize_torch_tensor(tensor, compression) for tensor, compression in zip(inputs, compressions)
+    )
```
mryab, line 241: By the way, why did you decide to serialize tensors inside this coroutine instead of doing it before the function call? The execution flow seems to be the same

hivemind/moe/client/moe.py
```diff
@@ -408,17 +413,15 @@ def _collect_responses(
...
+def _process_dispatched_task(task: Future, detect_anomalies: bool) -> Optional[Tuple[torch.Tensor]]:
     if task.exception() or task.cancelled():
         logger.warning(f"Task {task} failed: {type(task.exception())}")
         return None
 
-    deserialized_outputs = []
-    for tensor in task.result().tensors:
-        deserialized_tensor = deserialize_torch_tensor(tensor)
-        if detect_anomalies and not deserialized_tensor.isfinite().all():
+    outputs = tuple(task.result())
```
mryab, line 421: Maybe you can avoid the conversion to a tuple if you specify the output after serialization to be a tuple?

hivemind/moe/server/connection_handler.py
```diff
@@ -1,21 +1,25 @@
+import asyncio
 import multiprocessing as mp
-import os
-from typing import Dict
+from typing import AsyncIterator, Dict, Iterable, List, Optional, Tuple, Union
```
mryab, line 3: Ironically, the `Optional` import is not used

hivemind/moe/server/connection_handler.py
```diff
@@ -24,59 +28,106 @@ class ConnectionHandler(mp.context.ForkProcess):
...
-
-            await server.start()
-            self.ready.set()
-            await server.wait_for_termination()
-            logger.debug(f"ConnectionHandler terminated: (pid={os.getpid()})")
+            try:
+                self._p2p = await self.dht.replicate_p2p()
+                await self.add_p2p_handlers(self._p2p, balanced=True)
+
+                await asyncio.Future()
```
mryab, line 46: Is this just a shorthand for "wait infinitely"?

hivemind/moe/server/connection_handler.py
```diff
@@ -24,59 +28,106 @@ class ConnectionHandler(mp.context.ForkProcess):
...
+    ) -> Tuple[str, List[torch.Tensor]]:
+        expert_uid = None
+
+        def _unpack(req: runtime_pb2.ExpertRequest) -> Iterable[runtime_pb2.Tensor]:
+            nonlocal expert_uid
+
+            if expert_uid is None:
+                expert_uid = req.uid
+            elif expert_uid != req.uid:
+                raise ValueError("Expert uids differ in one reques")
```
mryab, line 73:
```suggestion
                raise ValueError("Expert uids differ in one request")
```

hivemind/moe/server/server.py
```diff
@@ -66,22 +64,19 @@ def __init__(
...
             self.checkpoint_saver = None
         self.runtime = Runtime(self.experts, **kwargs)
 
-        if self.dht and self.experts:
+        if self.experts:
             self.dht_handler_thread = DHTHandlerThread(
                 experts=self.experts,
                 dht=self.dht,
-                endpoint=self.listen_on,
+                peer_id=self.dht.peer_id,
```
mryab, line 79: Since you are already passing `dht` and it is not None at this point, you can avoid passing `peer_id` and obtain it from `dht` inside `DHTHandlerThread` when you need it

hivemind/moe/server/server.py
```diff
@@ -313,14 +301,14 @@ def shutdown(self):
 
 
 @contextmanager
-def background_server(*args, shutdown_timeout=5, **kwargs) -> Tuple[Endpoint, List[Multiaddr]]:
-    """A context manager that creates server in a background process, awaits .ready on entry and shuts down on exit"""
+def background_server(*args, shutdown_timeout=5, **kwargs) -> PeerInfo:
+    """A context manager that creates server in a background thread, awaits .ready on entry and shuts down on exit"""
```
mryab, line 305:
```suggestion
    """A context manager that creates server in a background process, awaits .ready on entry and shuts down on exit"""
```

mryab: L307 states that it's a process

hivemind/moe/server/server.py
```diff
@@ -147,17 +140,15 @@ def create(
 
         :param start: if True, starts server right away and returns when server is ready for requests
         :param stats_report_interval: interval between two reports of batch processing performance statistics
+        :param kwargs: any other params will be forwarded to DHT upon creation
```
mryab, line 143:
```suggestion
        :param kwargs: any other params that will be forwarded to DHT upon creation
```

hivemind/moe/server/dht_handler.py
```diff
@@ -73,27 +75,29 @@ async def _declare_experts(
...
     if expiration_time is None:
         expiration_time = get_dht_time()
     num_workers = len(uids) if dht.num_workers is None else min(len(uids), dht.num_workers)
     found: Dict[ExpertUID, DHTValue] = await node.get_many(uids, expiration_time, num_workers=num_workers)
 
     experts: List[Optional[RemoteExpert]] = [None] * len(uids)
     for i, uid in enumerate(uids):
-        if found[uid] is not None and isinstance(found[uid].value, Endpoint):
-            experts[i] = RemoteExpert(uid, found[uid].value)
+        elem = found[uid]
```
mryab, line 100:
```suggestion
        expert_for_uid = found[uid]
```

hivemind/moe/client/expert.py
```diff
@@ -52,18 +81,171 @@ def forward(self, *args, **kwargs):
...
+            outputs = RemoteExpertWorker.run_coroutine(self.stub.rpc_info(runtime_pb2.ExpertUID(uid=self.uid)))
+            self._rpc_info = MSGPackSerializer.loads(outputs.serialized_info)
+        return self._rpc_info
 
     def extra_repr(self):
-        return f"uid={self.uid}, endpoint={self.endpoint}"
+        return f"uid={self.uid}, server_peer_info={self.server_peer_info}"
+
+
+class RemoteExpertWorker:
```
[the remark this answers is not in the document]
GreenFatGuy: Good idea, I made it also a private class. Now functions which create `RemoteExpert` are free functions. 
I think it is possible to make them also private, but I don't know right now if we need this. Please let me know if you think they should be private

hivemind/moe/client/expert.py
```diff
@@ -1,43 +1,72 @@
...
+from dataclasses import dataclass
+from queue import Queue
+from threading import Thread
+from typing import Any, Awaitable, Dict, Iterable, List, Optional, Sequence, Tuple, Union
 
 import torch
 import torch.nn as nn
 from torch.autograd.function import once_differentiable
 
+from hivemind import moe
```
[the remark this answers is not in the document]
GreenFatGuy: I agree that generic imports are bad, but this done to avoid circular import. `ConnectionHandler` is the part of `moe/server`. There is also `dht_handlers.py` which imports `moe.client.Remoteexpert`. If we try to import something from `moe/server directly` we will get circular import.


### TLDR

 - `moe.server.dht_handlers` depends on `moe.clien.expert` (already). Thus `moe.server` depends on `moe.client`
 - importing `moe.server.connection_handler` into  `moe.client.expert` creates circular deependency.

hivemind/utils/streaming.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+StreamMessage = TypeVar("StreamMessage")
+
+
+async def gather_from_streaming(
+    stream: AsyncIterator[StreamMessage],
+    key: Callable[[StreamMessage], Iterable[runtime_pb2.Tensor]],
+    deserializer: Callable[[runtime_pb2.Tensor], torch.Tensor],
+) -> List[torch.Tensor]:
```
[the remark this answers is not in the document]
GreenFatGuy: Good ideas!

However there is a problem with circular imports if we want to inline `deserialize_torch_tensor`: it is from `compression` module, which has dependency on `utils`

hivemind/moe/client/expert.py
```diff
@@ -52,18 +81,171 @@ def forward(self, *args, **kwargs):
...
+        experts: List[Optional[RemoteExpert]] = []
+        for i in infos:
+            if i is not None:
+                experts.append(RemoteExpert(i, p2p))
+            else:
+                experts.append(None)
+        return experts
+
+    @classmethod
+    def spawn_experts(
```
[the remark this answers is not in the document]
GreenFatGuy: Renamed it as `create_remote_experts` because `get_experts` is already taken by `moe.server.dht_handler.get_expert`. For it is a public function, be if you think so we can make it private

hivemind/moe/client/expert.py
GreenFatGuy: Agreed, but I would do this changes later, because it seems to be a thing which do not change functionality. It is rather performance tune, and performance right now is acceptable for merging.

Added that into list of thing to be done afterwards.

hivemind/moe/server/connection_handler.py
GreenFatGuy: yes, it is. Added a comment

justheuristic:
@mryab another round please :)

hivemind/utils/streaming.py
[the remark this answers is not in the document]
GreenFatGuy: Renamed into `combine_and_deserialize_from_streaming` and removed `key`. However, `deserializer` is still passed outside, because of circular imports problem

borzunov (approved):
Please address these comments. The rest looks good :)

tests/test_connection_handler.py
```diff
@@ -0,0 +1,170 @@
...
+        p for t in [serialize_torch_tensor(inputs_long)] for p in split_for_streaming(t, chunk_size_bytes=2**16)
+    )
+    output_generator = await client_stub.rpc_forward_stream(
+        amap_in_executor(
+            lambda tensor_part: runtime_pb2.ExpertRequest(uid="expert2", tensors=[tensor_part]),
+            iter_as_aiter(split),
+        ),
+    )
+    outputs_list = [part async for part in output_generator]
+    del output_generator
```
mryab, line 63: Is explicitly deleting the generator necessary? Usually, such statements mean that something fishy is going on with the variable lifetime

tests/test_connection_handler.py
```diff
@@ -0,0 +1,170 @@
...
+    response = await client_stub.rpc_forward(
+        runtime_pb2.ExpertRequest(uid="expert2", tensors=[serialize_torch_tensor(inputs)])
+    )
+    outputs = deserialize_torch_tensor(response.tensors[0])
+    assert len(response.tensors) == 1
+    assert torch.allclose(outputs, inputs * 2)
+
+    # forward streaming
+    split = (
+        p for t in [serialize_torch_tensor(inputs_long)] for p in split_for_streaming(t, chunk_size_bytes=2**16)
```
mryab, line 54: If 2**16 is the default chunk size, we can use it from hivemind or at least define it at the top of this file

tests/test_connection_handler.py
```diff
@@ -0,0 +1,170 @@
...
+    )
+    output_generator = await client_stub.rpc_forward_stream(
+        amap_in_executor(
+            lambda tensor_part: runtime_pb2.ExpertRequest(uid="expert2", tensors=[tensor_part]),
+            iter_as_aiter(split),
+        ),
+    )
+    outputs_list = [part async for part in output_generator]
+    del output_generator
+    assert len(outputs_list) == 8  # message size divided by DEFAULT_MAX_MSG_SIZE
```
mryab, line 64: Instead of introducing a constant, it's better to perform the division right here

tests/test_connection_handler.py
```diff
@@ -0,0 +1,170 @@
...
+
+    results = await combine_and_deserialize_from_streaming(
+        amap_in_executor(lambda r: r.tensors, iter_as_aiter(outputs_list)), deserialize_torch_tensor
+    )
+    assert len(results) == 1
+    assert torch.allclose(results[0], inputs_long * 2)
+
+    # forward errors
```
mryab, line 72: Perhaps it's possible to separate this 200-line test into several smaller ones? This way, the error spot becomes more precise and running the tests in parallel also gets easier

tests/test_moe.py
```diff
@@ -261,9 +272,10 @@ def test_client_anomaly_detection():
     server.start()
     try:
         server.ready.wait()
+        dht_experts = DHT(initial_peers=dht.get_visible_maddrs(), start=True)
```
mryab, line 275:
```suggestion
        client_side_dht = DHT(initial_peers=dht.get_visible_maddrs(), start=True)
```

tests/test_p2p_daemon_bindings.py
```diff
@@ -560,13 +560,19 @@ async def handle_another_proto(stream_info, reader, writer):
...
         event_third.set()
 
-    await p2pcs[1].stream_handler(another_proto, handler_third)
+    # p2p raises now for doubled stream handlers
+    with pytest.raises(ControlFailure):
+        await p2pcs[1].stream_handler(another_proto, handler_third)
+
+    # add in balanced mode: handler should be placed in round robin queue
+    # and become the next to be called
+    await p2pcs[1].stream_handler(another_proto, handler_third, True)
```
mryab, line 575:
```suggestion
    await p2pcs[1].stream_handler(another_proto, handler_third, balanced=True)
```

mryab (commented):
I think we're almost ready :) Left a couple of (hopefully, last) comments, some of which are less serious and will be resolved as soon as I see them added to a TODO. Also, please check out https://github.com/learning-at-home/hivemind/pull/477

hivemind/moe/client/expert.py
GreenFatGuy: The thing here is that after the remote call of forward/backward we get `AsyncIterator` back, because tensors are being streamed. 

To make things the way you ask we have to materialize both inputs and outputs. We have to make a list of serialized tensors, pass them into coroutine, then after getting the results from net, we have to make a list of serialized tensors inside coroutine, and only then  deserialize them outside of async and create a list `torch.Tensors`.

The problem I see is materializing lists on every tiny step. On the other hand I can't say right now if it is more costly then serialization/deserialization itself.

You think this way is better than now? (we have to add +2 materializations of python list/tuple, instead of processing "on the fly")

By the way, I think that I did not quite what you asked: I just moved generator out of coroutine. But this does not mean that serialization happens outside. Generators are lazy, so serialization is still performed inside coroutine, only when we begin to iterate over parts of genrator. If i understand it the right way, serialization happens inside the executor which is hidden inside `amap_in_executor`

benchmarks/benchmark_throughput.py
```diff
@@ -7,7 +7,8 @@
 import torch
 
 from hivemind.dht import DHT
-from hivemind.moe.client.expert import RemoteExpert, RemoteExpertInfo, RemoteExpertWorker
+from hivemind.moe.client.expert import RemoteExpert, RemoteExpertInfo
+from hivemind.moe.client.remote_expert_worker import _RemoteExpertWorker
```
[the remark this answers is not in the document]
GreenFatGuy: Ok. Just wanted to add that the way we create `RemoteExpert` in benchmark doesn't look like the common way, it is more like an exceptional case

tests/test_connection_handler.py
GreenFatGuy: `DEFAULT_MAX_MSG_SIZE` in hivemind is `2**22` (`4 * 1024**2`). This number was chosen as just "big enough".

tests/test_connection_handler.py
justheuristic: done

docs/user/moe.md
```diff
@@ -104,10 +101,10 @@ hivemind-server --expert_cls ffn --hidden_dim 512 --num_experts 10 --expert_patt
...
-By default, the server will only accept connections from your local machine. To access it globally, you should replace
-`127.0.0.1` part from initial peers with server's IP address. Hivemind supports both ipv4 and ipv6 protocols and uses the same notation
-as [libp2p](https://docs.libp2p.io/concepts/addressing/). You can find more details on multiaddresses in the 
-[DHT tutorial](https://learning-at-home.readthedocs.io/en/latest/user/dht.html).
+By default, the server will only accept connections fr om your local network. 
+To enable training over the internet (or some other network), you should set `--host_maddrs` and `--announce_maddrs`.
+These option also allow you to select ipv4 / ipv6 network protocols and tcp / quic transport protocols.
+You can find more details in the [DHT tutorial](https://learning-at-home.readthedocs.io/en/latest/user/dht.html).
 
 ## Train the experts
```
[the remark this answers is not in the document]
justheuristic: No we do not - and this was one of the  design goals. worker is an internal structure that should not complicate user interaction.

hivemind/moe/client/expert.py
```diff
@@ -1,43 +1,73 @@
...
+from hivemind.p2p.p2p_daemon import DEFAULT_MAX_MSG_SIZE
+from hivemind.proto import runtime_pb2
+from hivemind.utils import (
+    MSGPackSerializer,
+    amap_in_executor,
+    iter_as_aiter,
+    nested_compare,
+    nested_flatten,
+    nested_pack,
+)
```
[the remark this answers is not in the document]
justheuristic: done

tests/test_connection_handler.py
justheuristic: done

hivemind/moe/server/dht_handler.py
```diff
@@ -14,28 +14,27 @@
...
-        assert get_port(endpoint) is not None
-        self.endpoint = endpoint
         self.experts = experts
         self.dht = dht
         self.update_period = update_period
         self.stop = threading.Event()
 
     def run(self) -> None:
-        declare_experts(self.dht, self.experts.keys(), self.endpoint)
+        declare_experts(self.dht, self.experts.keys(), self.dht.peer_id)
```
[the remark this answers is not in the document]
justheuristic: fixed (here and in tests)

hivemind/utils/networking.py
```diff
@@ -10,26 +10,6 @@
 LOCALHOST = "127.0.0.1"
 
 
-def get_port(endpoint: Endpoint) -> Optional[Port]:
```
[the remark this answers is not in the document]
justheuristic: removed

hivemind/moe/client/moe.py
```diff
@@ -1,20 +1,21 @@
...
 import torch
 import torch.nn as nn
 from torch.autograd.function import once_differentiable
 
-from hivemind.compression import deserialize_torch_tensor, serialize_torch_tensor
+from hivemind.compression import serialize_torch_tensor
 from hivemind.dht import DHT
 from hivemind.moe.client.beam_search import MoEBeamSearcher
-from hivemind.moe.client.expert import DUMMY, RemoteExpert, _get_expert_stub
+from hivemind.moe.client.expert import DUMMY, RemoteExpert, _get_expert_stub, expert_backward, expert_forward
```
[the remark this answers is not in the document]
justheuristic: renamed to get_expert_stub

tests/test_connection_handler.py
```diff
@@ -0,0 +1,170 @@
+from __future__ import annotations
+
+import asyncio
+from typing import Any, Dict
+
+import pytest
+import torch
+
+import hivemind
```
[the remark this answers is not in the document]
justheuristic: done

tests/test_connection_handler.py
justheuristic: removed

justheuristic:
Note: i've bumped timeout from 12 to 15 minutes. Rationale: otherwise you have a ~30% chance of not converging in time if you are randomly assigned to a node with weaker CPU.

hivemind/moe/client/__init__.py
```diff
@@ -1,3 +1,3 @@
-from hivemind.moe.client.expert import RemoteExpert
+from hivemind.moe.client.expert import RemoteExpert, batch_create_remote_experts, create_remote_experts
```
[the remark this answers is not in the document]
mryab: This is why I asked :) Maybe it's better to just avoid importing them and to have an explicit import where it's really needed?

mryab (approved):

PR #473, opened 2022-05-28
Generate new private key if identity file doesn't exist

borzunov:


examples/albert/arguments.py
```diff
@@ -38,7 +38,7 @@ class BaseTrainingArguments:
         default=None,
         metadata={
             "help": "Path to a pre-generated private key file. If defined, makes the peer ID deterministic. "
-            "May be generated using ``./p2p-keygen`` from ``go-libp2p-daemon``."
+            "If the file does not exist, generates a new private key."
```
justheuristic, line 41: [nit] ... generates and saves?

hivemind/p2p/p2p_daemon.py
```diff
@@ -113,8 +115,8 @@ async def create(
                          Details: https://pkg.go.dev/github.com/libp2p/go-libp2p-kad-dht#ModeOpt
         :param force_reachability: Force reachability mode (public/private)
         :param host_maddrs: Multiaddrs to listen for external connections from other p2p instances
-        :param identity_path: Path to a pre-generated private key file. If defined, makes the peer ID deterministic.
-                              May be generated using ``./p2p-keygen`` from ``go-libp2p-daemon``.
+        :param identity_path: Path to a private key file. If defined, makes the peer ID deterministic.
+                              If the file does not exist, generates a new private key.
```
justheuristic, line 119: [nit] ... and saves it to that file

hivemind/p2p/p2p_daemon.py
```diff
@@ -211,6 +217,15 @@ async def create(
...
         return self
 
+    @staticmethod
+    def generate_identity(identity_path: str) -> None:
+        private_key = RSAPrivateKey()
+        protobuf = crypto_pb2.PrivateKey(key_type=crypto_pb2.KeyType.RSA, data=private_key.to_bytes())
+
+        with open(identity_path, "wb") as f:
+            f.write(protobuf.SerializeToString())
+        os.chmod(identity_path, 0o400)
```
justheuristic, line 227: awesome!

justheuristic (approved):

hivemind/p2p/p2p_daemon.py
```diff
@@ -165,7 +167,11 @@ async def create(
         ]:
             if value:
                 process_kwargs[param] = self._maddrs_to_str(value)
+
         if identity_path is not None:
+            if not os.path.isfile(identity_path):
```
justheuristic, line 172: [nit] there may be a situation where not only file is missing, but the preceding directory is also nonexistent. Here's what can happen:
- create folders (os.mkdirs(..., exist_ok=True))
- raise an explicit error message ("e.g. parent directory does not exist, not "file not found"")

borzunov: I've added the explicit error message to `P2P.generate_identity_file()`.

PR #475, opened 2022-05-31
Add DDP support to hivemind.optim

borzunov:
**Status:** This PR is an early draft intended to validate the design of `hivemind.DDPOptimizer`. I didn't run the code even once yet.

**Co-authored-by:** @justheuristic

### DIFF
--- a/hivemind/optim/ddp.py
+++ b/hivemind/optim/ddp.py
@@ -91,10 +91,6 @@ class DDPOptimizer(Optimizer):
 
         return super().__getattribute__(name)
 
-    def is_alive(self) -> bool:
-        # On followers, this always returns False since there's nothing to shut down in __del__()
-        return self.is_ddp_leader() and super().is_alive()
-
     def _compute_state_version(self) -> int:
         """Return a non-decreasing integer that goes up whenever model params and/or buffers were updated"""
 
@@ -162,18 +158,6 @@ class DDPOptimizer(Optimizer):
 
         return loss if self.is_ddp_leader() else None
 
-    def load_state_from_peers(self, **kwargs) -> None:
-        if self.is_ddp_leader():
-            super().load_state_from_peers(**kwargs)
-
-        self._sync_among_ddp_ranks()
-
-    def load_state_dict(self, state_dict: dict) -> None:
-        if self.is_ddp_leader():
-            super().load_state_dict(state_dict)
-
-        self._sync_among_ddp_ranks()
-
     @property
     def param_groups(self) -> ParamGroups:
         if self.is_ddp_leader():
@@ -188,3 +172,7 @@ class DDPOptimizer(Optimizer):
     def shutdown(self):
         if self.is_ddp_leader():
             super().shutdown()
+
+    def __del__(self):
+        if self.is_ddp_leader():
+            super().__del__()
--- a/hivemind/optim/ddp.py
+++ b/hivemind/optim/ddp.py
@@ -1,0 +1,190 @@
+import time
+from typing import Callable, Optional, Union
+
+import torch
+from torch.distributed.distributed_c10d import _get_default_group, _get_default_store
+
+from hivemind.dht import DHT
+from hivemind.optim.grad_scaler import GradScaler
+from hivemind.optim.optimizer import Optimizer
+from hivemind.optim.state_averager import OptimizerFactory, Parameters, ParamGroups, TorchOptimizer, TrainingStateAverager
+from hivemind.utils import get_logger
+
+logger = get_logger(__name__)
+
+
+class DDPOptimizer(Optimizer):
+    _DDP_LEADER_RANK = 0
+    _BROADCAST_BUFFER_SIZE = 250 * 1024 ** 2
+
+    @staticmethod
+    def is_ddp_enabled():
+        return torch.distributed.is_initialized()
+
+    @staticmethod
+    def is_ddp_leader():
+        return not torch.distributed.is_initialized() or torch.distributed.get_rank() == DDPOptimizer._DDP_LEADER_RANK
+
+    def __init__(
+        self,
+        *,
+        dht: Optional[DHT] = None,
+        optimizer: Union[TorchOptimizer, OptimizerFactory],
+        params: Optional[Union[Parameters, ParamGroups]] = None,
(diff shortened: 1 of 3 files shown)


hivemind/optim/ddp.py
```diff
@@ -0,0 +1,190 @@
...
+from hivemind.optim.optimizer import Optimizer
+from hivemind.optim.state_averager import OptimizerFactory, Parameters, ParamGroups, TorchOptimizer, TrainingStateAverager
+from hivemind.utils import get_logger
+
+logger = get_logger(__name__)
+
+
+class DDPOptimizer(Optimizer):
+    _DDP_LEADER_RANK = 0
+    _BROADCAST_BUFFER_SIZE = 250 * 1024 ** 2
```
justheuristic, line 18: New pytorch seems to have finally implemented broadcast_coalesced in distributed,
we can directly address this https://pytorch.org/docs/stable/_modules/torch/nn/parallel/comm.html#broadcast_coalesced as long as we bump minimal pytorch version. Wadayathink?

hivemind/optim/ddp.py
```diff
@@ -0,0 +1,190 @@
...
+
+    @staticmethod
+    def is_ddp_enabled():
+        return torch.distributed.is_initialized()
+
+    @staticmethod
+    def is_ddp_leader():
```
justheuristic, line 25: nit: would recommend reusing the same terminology as *somewhere*, such as inside [DistributedDataParallel](https://github.com/pytorch/pytorch/blob/master/torch/nn/parallel/distributed.py)

For instance, the above DDP uses
- leader rank -> authoritative rank
- is_ddp_enabled -> _initialized

hivemind/optim/ddp.py
```diff
@@ -0,0 +1,190 @@
...
+            )
+
+        return super().__getattribute__(name)
+
+    def is_alive(self) -> bool:
+        # On followers, this always returns False since there's nothing to shut down in __del__()
+        return self.is_ddp_leader() and super().is_alive()
+
+    def _compute_state_version(self) -> int:
+        """Return a non-decreasing integer that goes up whenever model params and/or buffers were updated"""
```
justheuristic, line 99: This function is meant as a workaround to catch the moment when optimizer has updated parameters (load from peers, apply optimizer step average params)

All changes to state are currently handled in StateAverager.
Maybe we can implement StateAverager.local_version that gets incremented every time StateAverager loads, averages or updates state by optimizer

hivemind/optim/ddp.py
```diff
@@ -0,0 +1,190 @@
...
+                checksum = sum(p.sum().item() for p in self._main_parameters)
+                logger.debug(f"Parameter checksum (ddp_rank={rank}): {float(checksum)}")
+
+        return loss if self.is_ddp_leader() else None
+
+    def load_state_from_peers(self, **kwargs) -> None:
+        if self.is_ddp_leader():
+            super().load_state_from_peers(**kwargs)
+
+        self._sync_among_ddp_ranks()
```
justheuristic, line 169: We should not synchronize here: non-master ranks cannot call this and we will deadlock.

__We should only sync in step__ -- and after step check IF master updated/loaded/averaged step and then broadcast.

hivemind/optim/ddp.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def load_state_dict(self, state_dict: dict) -> None:
+        if self.is_ddp_leader():
+            super().load_state_dict(state_dict)
+
+        self._sync_among_ddp_ranks()
```
justheuristic, line 175: We should not synchronize here: non-master ranks cannot call this and we will deadlock: see load_state_from_peers

hivemind/optim/ddp.py
```diff
@@ -0,0 +1,190 @@
...
+        else:
+            return self._param_groups
+
+    def zero_grad(self, set_to_none: bool = False):
+        # We explicitly define this method to mark that it should be available on the DDP followers
+        super().zero_grad(set_to_none)
+
+    def shutdown(self):
+        if self.is_ddp_leader():
+            super().shutdown()
```
justheuristic, line 190: else raise NotImplemented

hivemind/optim/ddp.py
```diff
@@ -0,0 +1,190 @@
...
+            raise RuntimeError(
+                f"{self.__class__.__name__}.{name} is only available on the DDP leader. "
+                f"Please access it only if DDPOptimizer.is_ddp_leader() == True"
+            )
+
+        return super().__getattribute__(name)
+
+    def is_alive(self) -> bool:
+        # On followers, this always returns False since there's nothing to shut down in __del__()
+        return self.is_ddp_leader() and super().is_alive()
```
justheuristic, line 96: if leader:
   return is_alive
else:
   raise NotImplementedError?

hivemind/optim/state_averager.py
```diff
@@ -131,10 +131,10 @@ def __init__(
         )
 
     @staticmethod
-    def _check_params(
+    def check_params(
```
borzunov, line 134:
```suggestion
    def prepare_params(
```

hivemind/optim/ddp.py
```diff
@@ -0,0 +1,178 @@
...
+from hivemind.dht import DHT
+from hivemind.optim.grad_scaler import GradScaler
+from hivemind.optim.optimizer import Optimizer
+from hivemind.optim.state_averager import OptimizerFactory, Parameters, ParamGroups, TorchOptimizer, TrainingStateAverager
+from hivemind.utils import get_logger
+
+logger = get_logger(__name__)
+
+
+class DDPOptimizer(Optimizer):
```
borzunov, line 16: **Note to self:** A better way to do it is:

- Don't inherit `hivemind.Optimizer`
- Make `_create_optimizer()` method and forward `__init__`'s kwargs there
- Make `opt` property
- _Maybe_ create `__getattr__` that can forward attrs to `opt`

borzunov, line 16: **Note to self:** A better way to do it is:

- Don't inherit `hivemind.Optimizer`
- Make `_create_optimizer()` method and forward `__init__`'s kwargs there
- Make `opt` property
- _Maybe_ create `__getattr__` that can forward attrs to `opt`

PR #479, opened 2022-06-06
Use PeerID exclusively to address MoE experts

justheuristic:
This PR changes declare_experts / RemoteExpert to use only p2p peer ID, not the whole multiaddress.
This slightly reduces the code complexity and gives you an easier time sharing experts with dynamic IP.

It also fixes one DHT edge case i've discovered when working on it.

Minor changes:
- fixed an edge case: previously, DHT would **freeze** if accessing DHT.peer_id or otherwise calling .run_coroutine from inside another run_coroutine
- merged RemoteExpertInfo and UidEndpoint into one structure (ExpertInfo), now in expert_uid.py
- extracted expert_uid.py from hivemind.moe.server to hivemind.moe in order to avoid circular imports
- renamed get_expert_stub into get_server_stub since it is not expert-specific

### DIFF
--- a/hivemind/moe/client/beam_search.py
+++ b/hivemind/moe/client/beam_search.py
@@ -199,7 +199,7 @@ class MoEBeamSearcher:
             coord: ExpertInfo(uid=match.value[0], peer_id=PeerID.from_base58(match.value[1]))
             for coord, match in entry.value.items()
             if isinstance(coord, Coordinate)
-            and (0 <= coord < grid_size or grid_size is None)
+            and (grid_size is None or 0 <= coord < grid_size)
             and isinstance(match, ValueWithExpiration)
             and isinstance(match.value, tuple)
             and len(match.value) == 2
--- a/hivemind/moe/client/beam_search.py
+++ b/hivemind/moe/client/beam_search.py
@@ -5,25 +5,21 @@ from functools import partial
 from typing import Deque, Dict, Iterator, List, Optional, Sequence, Set, Tuple, Union
 
 from hivemind.dht import DHT, DHTExpiration, DHTNode
-from hivemind.moe.client.expert import (
-    RemoteExpert,
-    RemoteExpertInfo,
-    batch_create_remote_experts,
-    create_remote_experts,
-)
-from hivemind.moe.server.expert_uid import (
+from hivemind.moe.client.expert import RemoteExpert, batch_create_remote_experts, create_remote_experts
+from hivemind.moe.expert_uid import (
     FLAT_EXPERT,
     PREFIX_PATTERN,
     UID_DELIMITER,
     Coordinate,
+    ExpertInfo,
     ExpertPrefix,
     ExpertUID,
     Score,
-    UidEndpoint,
     is_valid_prefix,
+    is_valid_uid,
 )
-from hivemind.p2p import PeerInfo
-from hivemind.utils import MPFuture, get_dht_time, get_logger
+from hivemind.p2p import PeerID
+from hivemind.utils import MPFuture, ValueWithExpiration, get_dht_time, get_logger
 
 logger = get_logger(__name__)
 
@@ -100,7 +96,7 @@ class MoEBeamSearcher:
 
     def get_initial_beam(
         self, scores: Sequence[float], beam_size: int, return_future: bool = False
-    ) -> List[Tuple[Score, ExpertPrefix, Dict[Coordinate, UidEndpoint]]]:
+    ) -> List[Tuple[Score, ExpertPrefix, Dict[Coordinate, ExpertInfo]]]:
         """
         :param scores: prefer suffix coordinates that have highest scores
         :param beam_size: select this many active suffixes with highest scores
@@ -130,9 +126,9 @@ class MoEBeamSearcher:
         negative_caching: bool,
         cache_expiration: DHTExpiration,
         num_workers: Optional[int] = None,
-    ) -> List[Tuple[Score, ExpertPrefix, Dict[Coordinate, UidEndpoint]]]:
+    ) -> List[Tuple[Score, ExpertPrefix, Dict[Coordinate, ExpertInfo]]]:
         num_workers = num_workers or dht.num_workers or beam_size
-        beam: List[Tuple[Score, ExpertPrefix, Dict[Coordinate, UidEndpoint]]] = []
+        beam: List[Tuple[Score, ExpertPrefix, Dict[Coordinate, ExpertInfo]]] = []
         unattempted_indices: List[Coordinate] = sorted(
             range(len(scores)), key=scores.__getitem__
         )  # from worst to best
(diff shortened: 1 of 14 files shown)


hivemind/moe/server/dht_handler.py
```diff
@@ -93,9 +93,9 @@ async def _get_experts(
...
 
-    experts: List[Optional[RemoteExpert]] = [None] * len(uids)
+    experts: List[Optional[RemoteExpertInfo]] = [None] * len(uids)
     for i, uid in enumerate(uids):
-        expert_info_for_uid = found[uid]
-        if expert_info_for_uid is not None and isinstance(expert_info_for_uid.value, tuple):
-            experts[i] = RemoteExpertInfo(uid, PeerInfo.from_tuple(expert_info_for_uid.value))
+        expert_server_peer_id = found[uid]
+        if expert_server_peer_id is not None and isinstance(expert_server_peer_id.value, str):
+            experts[i] = RemoteExpertInfo(uid, PeerID.from_base58(expert_server_peer_id.value))
```
borzunov, line 100: nit: Is "expert server peer" == "expert peer"? If it is, let's keep it simple.

```suggestion
        expert_peer_id = found[uid]
        if expert_peer_id is not None and isinstance(expert_peer_id.value, str):
            experts[i] = RemoteExpertInfo(uid, PeerID.from_base58(expert_peer_id.value))
```

borzunov (changes requested):

tests/test_dht_experts.py
```diff
@@ -35,7 +35,7 @@ def test_store_get_experts(n_peers=10):
     declare_experts(other_peer, [other_expert])
     first_notfound, first_found = get_experts(first_peer, ["foobar", other_expert])
     assert isinstance(first_found, hivemind.RemoteExpert)
-    assert first_found.server_peer_info.peer_id == other_peer.peer_id
+    assert first_found.server_peer_id.peer_id == other_peer.peer_id
```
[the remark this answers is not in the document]
justheuristic: fixed

hivemind/moe/server/dht_handler.py
justheuristic: fixed

hivemind/moe/client/expert.py
```diff
@@ -57,12 +58,12 @@ def uid(self):
         return self._info.uid
 
     @property
-    def server_peer_info(self):
-        return self._info.peer_info
+    def server_peer_id(self):
```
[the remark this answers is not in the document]
justheuristic: done

hivemind/moe/client/expert.py
```diff
@@ -57,12 +50,12 @@ def uid(self):
         return self._info.uid
 
     @property
-    def server_peer_info(self):
-        return self._info.peer_info
+    def server_peer_id(self) -> PeerID:
```
borzunov, line 53:
```suggestion
    def peer_id(self) -> PeerID:
```

I'd suggest to rename it like this because (a) it's clear that `RemoteExpert.peer_id` is the server's peer ID (not our own) and (b) many `hivemind` classes have the `peer_id` property and it would be cool to be in line with them.

@GreenFatGuy Are you okay with this change? (@justheuristic asked for your confirmation)

hivemind/moe/client/beam_search.py
```diff
@@ -306,6 +312,9 @@ async def _find_best_experts(
                     if isinstance(next_coord, int) and 0 <= next_coord < len(dim_scores)
                 ),
             )
+
+            if not best_active_pairs:
+                break
```
GreenFatGuy, line 317: If I understand everything right, then without this `if` we would also break the cycle, because eventually would get empty `beam` on L330 and reach `break` at L333. However on L332 we have `log.warning`, which telling us that this situation should get some attention.
TLDR : shouldn't we do `log.warning` here?

hivemind/moe/server/dht_handler.py
```diff
@@ -87,15 +88,15 @@ def get_experts(
...
 
-    experts: List[Optional[RemoteExpert]] = [None] * len(uids)
+    experts: List[Optional[ExpertInfo]] = [None] * len(uids)
     for i, uid in enumerate(uids):
-        expert_info_for_uid = found[uid]
-        if expert_info_for_uid is not None and isinstance(expert_info_for_uid.value, tuple):
-            experts[i] = RemoteExpertInfo(uid, PeerInfo.from_tuple(expert_info_for_uid.value))
+        server_peer_id = found[uid]
+        if server_peer_id is not None and isinstance(server_peer_id.value, str):
+            experts[i] = ExpertInfo(uid, PeerID.from_base58(server_peer_id.value))
```
mryab, line 101: This construction (involving `PeerID.from_base58`) occurs at least 3 times in the codebase, perhaps it's worth changing `ExpertInfo` into a dataclass with an additional classmethod `from_binary` for simplicity?

hivemind/moe/client/beam_search.py
```diff
@@ -151,11 +147,14 @@ async def _get_initial_beam(
...
-                        coord: UidEndpoint(uid=match.value[0], peer_info=PeerInfo.from_tuple(match.value[1]))
+                        coord: ExpertInfo(uid=match.value[0], peer_id=PeerID.from_base58(match.value[1]))
                         for coord, match in maybe_prefix_data.value.items()
                         if isinstance(coord, Coordinate)
-                        and isinstance(getattr(match, "value", None), list)
+                        and isinstance(match, ValueWithExpiration)
+                        and isinstance(match.value, tuple)
                         and len(match.value) == 2
+                        and is_valid_uid(match.value[0])
+                        and isinstance(match.value[1], str)
```
mryab, line 157: Maybe extract this 5-line validation into a function?

hivemind/moe/client/beam_search.py
```diff
@@ -235,7 +237,11 @@ async def _get_active_successors(
         return successors
 
     def find_best_experts(
-        self, grid_scores: Sequence[Sequence[float]], beam_size: int, return_future: bool = False
+        self,
+        grid_scores: Sequence[Sequence[float]],
+        beam_size: int,
+        num_workers: Optional[int] = None,
```
mryab, line 243: In which case do we want to call this function with a different num_workers than what was created? 

Moreover, this function (without an underscore) is only used in tests, so not sure if it's worth making it more complicated

hivemind/moe/client/beam_search.py
mryab: Yes, I think that it should at least require an explanation via a comment/log entry

justheuristic: good catch, this if is not necessary, removed

hivemind/moe/client/beam_search.py
justheuristic: The reason i've added this is that num_workers was a __documented__ parameter in this function.

justheuristic: alternatively, i could rm it from docstring

justheuristic: [rm-ed]

hivemind/moe/client/beam_search.py
GreenFatGuy: Well, in fact this `if` gives some profit. We wont do a lot of calls just to find out that there is nothing to search for. As I understand, there will be 0 net communications in either way, be there will be some small time save.

Anyway, any change is acceptable: to add logging or to remove this `if` at all.

hivemind/moe/client/beam_search.py
justheuristic: done

borzunov (approved):

PR #481, opened 2022-06-08
Fix shape validation in GradientAverager

mryab:
There is a couple of typos in the shape-validating code of `__init__` in GradientAverager for the case where we pass both the parameters and their gradients that need to be averaged. Fixed the typos, the check itself and added a test to make sure we raise an error in this case


justheuristic (approved):
good catch!

PR #482, opened 2022-06-12
Change expiration time in declare_experts, fix update_period discrepancy

justheuristic:
* hivemind-server will now set expiration dynamically based on update_period
* default update_period is now 30s everywhere. Was 5s in DHTHandlerThread but overriden by 30s from Server.__init__

**Why?:** as @ial32 suggested on discord, the current hard-coded 300s expiration makes it hard to debug and may cause problems if update_period >= 300.

### DIFF
--- a/hivemind/moe/server/dht_handler.py
+++ b/hivemind/moe/server/dht_handler.py
@@ -20,17 +20,18 @@ from hivemind.utils import MPFuture, get_dht_time
 
 
 class DHTHandlerThread(threading.Thread):
-    def __init__(self, experts, dht: DHT, update_period: int = 5, **kwargs):
+    def __init__(self, experts, dht: DHT, update_period: int = 30, expiration: Optional[int] = None, **kwargs):
         super().__init__(**kwargs)
         self.experts = experts
         self.dht = dht
         self.update_period = update_period
+        self.expiration = expiration if expiration is not None else 2 * update_period
         self.stop = threading.Event()
 
     def run(self) -> None:
         declare_experts(self.dht, self.experts.keys())
         while not self.stop.wait(self.update_period):
-            declare_experts(self.dht, self.experts.keys())
+            declare_experts(self.dht, self.experts.keys(), expiration=self.expiration)
 
 
 def declare_experts(
--- a/hivemind/moe/server/dht_handler.py
+++ b/hivemind/moe/server/dht_handler.py
@@ -20,7 +20,7 @@ from hivemind.utils import MAX_DHT_TIME_DISCREPANCY_SECONDS, MPFuture, get_dht_t
 
 
 class DHTHandlerThread(threading.Thread):
-    def __init__(self, experts, dht: DHT, update_period: int = 30, expiration: Optional[int] = None, **kwargs):
+    def __init__(self, experts, dht: DHT, update_period: float = 30, expiration: Optional[int] = None, **kwargs):
         super().__init__(**kwargs)
         if expiration is None:
             expiration = max(2 * update_period, MAX_DHT_TIME_DISCREPANCY_SECONDS)
--- a/hivemind/moe/server/dht_handler.py
+++ b/hivemind/moe/server/dht_handler.py
@@ -31,20 +31,20 @@ class DHTHandlerThread(threading.Thread):
         self.stop = threading.Event()
 
     def run(self) -> None:
-        declare_experts(self.dht, self.experts.keys())
+        declare_experts(self.dht, self.experts.keys(), expiration_time=get_dht_time() + self.expiration)
         while not self.stop.wait(self.update_period):
-            declare_experts(self.dht, self.experts.keys(), expiration=self.expiration)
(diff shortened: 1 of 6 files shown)


borzunov (approved):

borzunov (approved):

hivemind/moe/server/dht_handler.py
```diff
@@ -16,21 +16,24 @@
     split_uid,
 )
 from hivemind.p2p import PeerID
-from hivemind.utils import MPFuture, get_dht_time
+from hivemind.utils import MAX_DHT_TIME_DISCREPANCY_SECONDS, MPFuture, get_dht_time
 
 
 class DHTHandlerThread(threading.Thread):
-    def __init__(self, experts, dht: DHT, update_period: int = 5, **kwargs):
+    def __init__(self, experts, dht: DHT, update_period: int = 30, expiration: Optional[int] = None, **kwargs):
```
mryab, line 23: If I'm not mistaken, it is not possible to pass `expiration` into this class for a default usage, since the corresponding constructor call in server.py [does not pass](https://github.com/learning-at-home/hivemind/blob/default_expiration/hivemind/moe/server/server.py#L74-L78) the expiration parameter. Is this intentional? If we have reports about making this hard to configure, maybe we should provide a way to configure this expiration explicitly

hivemind/moe/server/dht_handler.py
```diff
@@ -16,21 +16,24 @@
...
         self.stop = threading.Event()
 
     def run(self) -> None:
         declare_experts(self.dht, self.experts.keys())
         while not self.stop.wait(self.update_period):
-            declare_experts(self.dht, self.experts.keys())
+            declare_experts(self.dht, self.experts.keys(), expiration=self.expiration)
 
 
 def declare_experts(
```
mryab, line 39: If expiration is always passed, I think we can remove the default value from this function

hivemind/moe/server/dht_handler.py
```diff
@@ -16,21 +16,24 @@
...
-from hivemind.utils import MPFuture, get_dht_time
+from hivemind.utils import MAX_DHT_TIME_DISCREPANCY_SECONDS, MPFuture, get_dht_time
 
 
 class DHTHandlerThread(threading.Thread):
-    def __init__(self, experts, dht: DHT, update_period: int = 5, **kwargs):
+    def __init__(self, experts, dht: DHT, update_period: int = 30, expiration: Optional[int] = None, **kwargs):
         super().__init__(**kwargs)
+        if expiration is None:
+            expiration = max(2 * update_period, MAX_DHT_TIME_DISCREPANCY_SECONDS)
```
mryab, line 26: Even though I can understand why "the maximum time discepancy between DHT instances" kinda makes sense for a lower bound on the record expiration time, it's still not quite obvious why we use it here and not in all other cases for the expiration time limits. Maybe it's best to designate another module-local constant for this, since its purpose differs from measuring the discrepancy between two timestamps?

hivemind/moe/server/dht_handler.py
justheuristic: i'm afraid we can't, it will break compatibility with some examples

mryab: Couldn't find any: https://github.com/learning-at-home/hivemind/search?q=declare_experts

hivemind/moe/server/dht_handler.py
justheuristic: I respectfully disagree: the max discrepancy is introduced precisely in order to account for the allowed discrepancy between nodes.

tests/test_dht_experts.py
```diff
@@ -43,7 +44,7 @@ def test_store_get_experts(n_peers=10):
     time.sleep(1.0)
     remaining_peer1 = random.choice([peer for peer in peers if peer.is_alive()])
     remaining_peer2 = random.choice([peer for peer in peers if peer.is_alive()])
-    assert all(declare_experts(remaining_peer1, ["new_expert.1"]))
+    assert all(declare_experts(remaining_peer1, ["new_expert.1"], expiration_time=hivemind.get_dht_time() + 30))
```
mryab, line 47:
```suggestion
    assert all(declare_experts(remaining_peer1, ["new_expert.1"], expiration_time=get_dht_time() + 30))
```

tests/test_moe.py
```diff
@@ -2,6 +2,7 @@
 import pytest
 import torch
 
+import hivemind
```
mryab, line 5: Better to use a more specific import:
```suggestion
from hivemind.utils import get_dht_time
```

mryab (approved):
(assuming the imports above are fixed)

tests/test_moe.py
justheuristic: done

tests/test_dht_experts.py
justheuristic: done

PR #483, opened 2022-06-14
Simplify ExpertBackend interface

justheuristic:
The core idea is that we should not make hivemind internals conditional on a specific training technique, such as linear warmup or gradient clipping by norm. Instead, we let user define their own scheduler and/or optimizer as necessary.

- remove gradient clipping from ExpertBackend: this behavior can be achieved with a user-defined Optimizer
- remove stats from ExpertBackend: this behavior can be achieved with a user-defined Scheduler
- rename full_state -> state_dict, rationale: there is no "non-full" state in this context
- rename ExpertBackend.expert -> ExpertBackend.module to avoid confusion


hivemind/hivemind_cli/run_server.py
```diff
@@ -54,7 +54,8 @@ def main():
                         help='Server will report experts to DHT once in this many seconds')
     parser.add_argument('--expiration', type=float, required=False, default=None,
                         help='DHT entries will expire after this many seconds')
-    parser.add_argument('--num_total_steps', type=int, required=False, help='The total number of steps for LR schedule')
+    parser.add_argument('--num_training_steps', type=int, required=False, help='The total number of steps for LR schedule')
```
justheuristic, line 57: Source updated: https://github.com/huggingface/transformers/blob/v4.19.4/src/transformers/optimization.py#L75

hivemind/moe/server/expert_backend.py
```diff
@@ -163,65 +141,47 @@ def backward(self, *inputs: torch.Tensor) -> Tuple[torch.Tensor, ...]:
             torch.autograd.backward(
                 outputs_flat, grad_tensors=grad_outputs_flat, create_graph=False, retain_graph=False
             )
-            self.apply_gradients(batch_size)
+            self.on_backward(batch_size)
```
justheuristic, line 144: rationale: this does not necessarily apply gradients, e.g.
- virtual batching applies gradients once every k steps
- pretrained models do not apply the returned gradients

borzunov (approved):

hivemind/moe/server/layers/optim.py
```diff
@@ -0,0 +1,63 @@
...
+    @property
+    def param_groups(self):
+        return self.optim.param_groups
+
+    def add_param_group(self, param_group: dict) -> None:
+        return self.optim.add_param_group(param_group)
+
+
+class ClippingWrapper(OptimizerWrapper):
+    """A wrapper to pytorch.optimizer that clips gradients by global norm before each step"""
```
mryab, line 49:
```suggestion
    """A wrapper of torch.Optimizer that clips gradients by global norm before each step"""
```

hivemind/moe/server/expert_backend.py
```diff
@@ -1,9 +1,10 @@
-from typing import Any, Callable, Dict, Sequence, Tuple, Union
+from typing import Any, Dict, Optional, Sequence, Tuple, Union
 
 import torch
 from torch import nn
 
 from hivemind.moe.server.task_pool import TaskPool
+from hivemind.optim.state_averager import LRSchedulerBase
```
mryab, line 7: Since this import is not actually related to hivemind.optim, I’d suggest so simply inline the statement that is being imported

hivemind/moe/server/expert_backend.py
```diff
@@ -87,30 +75,22 @@ def __init__(
...
         """
         Apply forward pass to an aggregated batch of requests. Used by Runtime, do not call this manually;
         To submit a request for asynchronous processing, please use ``ExpertBackend.forward_pool.submit_task``.
 
         Subclassing:
            This method receives a sequence of torch tensors following ``nested_flatten(self.forward_schema)``;
-
            It should return gradients w.r.t. inputs that follow ``nested_flatten(self.outputs_schema)``;
-
-           .. todo we handle layer states (e.g. batchnorm stats) incorrectly, updating them twice.
```
mryab, line 103: Is this not correct anymore? :)

hivemind/moe/server/expert_backend.py
```diff
@@ -34,49 +35,36 @@ class ExpertBackend:
...
-        clip_grad_norm: float = None,
         **kwargs,
     ):
         super().__init__()
-        self.expert, self.optimizer, self.name = expert, optimizer, name
-
-        if scheduler is None:
-            self.scheduler = None
-        else:
-            assert optimizer is not None and num_warmup_steps is not None and num_total_steps is not None
```
mryab, line 64: Should we validate that the scheduler requires an optimizer? Unless it’s something that changes state other than optimization hyperparameters, but then it’s not an LRScheduler

hivemind/moe/server/expert_backend.py
```diff
@@ -163,65 +141,47 @@ def backward(self, *inputs: torch.Tensor) -> Tuple[torch.Tensor, ...]:
...
-
-    def get_stats(self) -> Dict:
-        """
-        Return current expert training statistics (number of updates, number of processed examples after
-        last optimizer step)
-        """
-        return {"updates": self.update_count, "examples_processed": self.examples_processed}
-
-    def get_full_state(self) -> Dict:
+    def state_dict(self) -> Dict:
```
mryab, line 161: The docstring is no longer correct, because batch processing statistics were removed

hivemind/moe/server/layers/optim.py
```diff
@@ -0,0 +1,63 @@
+import torch
+
+
+class OptimizerWrapper(torch.optim.Optimizer):
+    """A wrapper for pytorch.optim.Optimizer that forwards all methods to the wrapped optimizer"""
+
+    def __init__(self, optim: torch.optim.Optimizer):
+        object.__init__(self)
```
mryab, line 8: object? In that case, it’s not a true Optimizer

hivemind/moe/server/layers/optim.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @property
+    def param_groups(self):
+        return self.optim.param_groups
+
+    def add_param_group(self, param_group: dict) -> None:
+        return self.optim.add_param_group(param_group)
+
+
+class ClippingWrapper(OptimizerWrapper):
```
mryab, line 48: In case we need OptimizerWrapper just for this application, I’d suggest not to overcomplicate the code and just write one class for a specific use case

hivemind/moe/server/server.py
```diff
@@ -33,15 +34,15 @@
 
 class Server(threading.Thread):
     """
-    Server allows you to host "experts" - pytorch subnetworks used by Decentralized Mixture of Experts.
+    Server allows you to host "experts" - pytorch subnetworks that can be accessed by your peers in the swarm.
```
mryab, line 37:
```suggestion
    Server allows you to host "experts" - pytorch subnetworks that can be accessed remotely by peers.
```

hivemind/moe/server/server.py
```diff
@@ -189,21 +189,26 @@ def create(
         else:
             args_schema = (BatchTensorDescriptor.from_tensor(sample_input, compression),)
 
-        scheduler = schedule_name_to_scheduler[scheduler]
+        scheduler_cls = schedule_name_to_scheduler[scheduler]
+        if scheduler_cls is not None:
```
mryab, line 193: Perhaps it’s better to check `scheduler` for None even before accessing the dictionary

mryab (changes requested):

hivemind/moe/server/expert_backend.py
justheuristic: fixed

hivemind/moe/server/expert_backend.py
justheuristic: fixed

hivemind/moe/server/expert_backend.py
justheuristic: thanks)

hivemind/moe/server/layers/optim.py
justheuristic: on the contrary it defines all optimizer fields and methods as properties

hivemind/moe/server/server.py
justheuristic: scheduler is a string, it cannot be None. It can, though, be a sting "none", but that is handled by the dictionary

hivemind/moe/server/expert_backend.py
mryab: A warning would be sufficient, I suppose

justheuristic: restored the warning

mryab (approved):

PR #489, opened 2022-06-23
Handle errors in Runtime

justheuristic:
- fix edge case where expert requests with 3.99-4MB payload would fail due to max message size (due to serialization overhead)
- recover from errors in the Runtime, propagate them to the corresponding tasks
   - previously, a failing function would terminate the entire server - which was a major pain for me personally :)
   - failure to process a request will now trigger P2PHandlerError instead of P2PDaemonError (cuz it does not kill the daemon)
- allow optional metadata in ExpertRequest / ExpertResponse for extendability [todo: validate it vs. @mryab ]

### DIFF
--- a/hivemind/moe/server/runtime.py
+++ b/hivemind/moe/server/runtime.py
@@ -85,16 +85,23 @@ class Runtime(threading.Thread):
                     logger.debug(f"Processing batch {batch_index} from pool {pool.name}")
 
                     start = time()
-                    outputs = pool.process_func(*batch)
-                    batch_processing_time = time() - start
+                    try:
+                        outputs = pool.process_func(*batch)
+                        batch_processing_time = time() - start
 
-                    batch_size = outputs[0].size(0)
-                    logger.debug(f"Pool {pool.name}: batch {batch_index} processed, size {batch_size}")
+                        batch_size = outputs[0].size(0)
+                        logger.debug(f"Pool {pool.name}: batch {batch_index} processed, size {batch_size}")
 
-                    if self.stats_report_interval is not None:
-                        self.stats_reporter.report_stats(pool.name, batch_size, batch_processing_time)
+                        if self.stats_report_interval is not None:
+                            self.stats_reporter.report_stats(pool.name, batch_size, batch_processing_time)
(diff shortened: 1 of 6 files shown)


hivemind/moe/server/runtime.py
```diff
@@ -85,16 +85,23 @@ def run(self):
...
 
-                    if self.stats_report_interval is not None:
-                        self.stats_reporter.report_stats(pool.name, batch_size, batch_processing_time)
+                        if self.stats_report_interval is not None:
+                            self.stats_reporter.report_stats(pool.name, batch_size, batch_processing_time)
+
+                        output_sender_pool.apply_async(pool.send_outputs_from_runtime, args=[batch_index, outputs])
+                    except KeyboardInterrupt:
+                        raise
+                    except BaseException as exception:
```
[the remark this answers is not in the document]
justheuristic: It is, and it's kind of intentional: the expert runs user-defined computation that can fail in peculiar ways: python-side errors, pytorch memory problems, and even an occasional StopIteration (which is still BaseException)

hivemind/moe/server/runtime.py
```diff
@@ -85,16 +85,23 @@ def run(self):
...
+                        logger.exception(f"Caught {exception}, attempting to recover", exc_info=True)
```
mryab, line 102: `exc_info` is most likely not necessary

hivemind/p2p/p2p_daemon_bindings/control.py
```diff
@@ -27,6 +27,9 @@
 logger = get_logger(__name__)
 
 DEFAULT_MAX_MSG_SIZE = 4 * 1024**2
+MAX_UNARY_PAYLOAD_SIZE = DEFAULT_MAX_MSG_SIZE // 2
+# note: we check v.s. 2x max message size to account for serialization overhead. The actual overhead is
```
mryab, line 31:
```suggestion
# note: we check vs. 2x max message size to account for serialization overhead. The actual overhead is
```

mryab (approved):

PR #491, opened 2022-06-28
metadata type changed to bytes

GreenFatGuy:
Change type of ExpertRequest/Response metadata to bytes because bytes are native output/input for serializers (for example `MSGPackSerializer` used in hivemind)


justheuristic:
@GreenFatGuy please check the compatibility fix. If it looks alright, feel free to merge (with optionally updated description)

PR #495, opened 2022-07-14
Expose the number of gathered peers in the optimizer

IAL32:
This PR exposes the number of peers participating in the last averaging round to `hivemind.Optimizer`.

### DIFF
--- a/hivemind/optim/optimizer.py
+++ b/hivemind/optim/optimizer.py
@@ -345,6 +345,13 @@ class Optimizer(torch.optim.Optimizer):
         return self.state_averager.is_alive()
 
     @property
+    def gathered_peers(self) -> int:
+        """
+        The number of peers that participated in the last averaging round with this worker.
+        """
+        return len(self.state_averager.gathered_peers)
+
+    @property
     def local_epoch(self) -> int:
         """
         This worker's current epoch, kept synchronized with peers. If peer's local_epoch lags behind others, it will
--- a/hivemind/optim/state_averager.py
+++ b/hivemind/optim/state_averager.py
@@ -117,6 +117,7 @@ class TrainingStateAverager(DecentralizedAverager):
         self.opt_keys_for_averaging, self.extra_tensors = average_opt_statistics, extra_tensors
         self.sync_epoch_when_averaging = sync_epoch_when_averaging
         self.local_epoch = 0
+        self.gathered_peers = {}
 
         self.delay_before_averaging = PerformanceEMA(alpha=performance_ema_alpha)
         self.step_executor = ThreadPoolExecutor(max_workers=2 if self.delta_rule_averaging else 1)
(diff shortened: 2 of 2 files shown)


hivemind/optim/optimizer.py
```diff
@@ -344,6 +344,13 @@ def _compute_schema_hash(self) -> int:
     def is_alive(self) -> bool:
         return self.state_averager.is_alive()
 
+    @property
+    def gathered_peers(self) -> int:
```
borzunov, line 348: @justheuristic Is it a good idea to expose this or this info is already exposed somewhere?

Also, I'd expect a property named `gathered_peers` to return a collection of peers (e.g., peer IDs), not just the count of them. An alternative is to rename this property to something like `num_gathered_peers`.

borzunov (changes requested):
Thanks for the PR! I'll double-check with my colleagues that we indeed don't expose this info exposed somewhere else. Meanwhile, I've left a minor comment regarding the naming/semantics.

hivemind/optim/optimizer.py
justheuristic: Hi, @IAL32 !

I recognize that it is convenient to keep track of group size, but merging this specific code into hivemind may introduce confusion in some modes. If use_local_updates=False, optimizer will have two different averaging rounds (for state and for gradients), and it is not clear which of them we should report. Furthermore, if [moshpit](https://arxiv.org/abs/2103.03239) is enabled, optimizer can run multiple concurrent averaging rounds.

If you keep track of it in your experiments, you can subclass hivemind.Optimizer as such:

```python

class MyOptimizer(hivemind.Optimizer)
    @property
    def gathered_peers(self) -> int:
        return len(self.state_averager.gathered_peers)

    def _make_state_averager(self, **kwargs) -> TrainingStateAverager:
        # https://github.com/learning-at-home/hivemind/blob/master/hivemind/optim/optimizer.py#L282-L297
        return MyModifiedStateAverager(see_base_funciton_code(**kwargs))

class MyModifiedStateAverager(TrainingStateAverager):
    gathered_peers = ...

```

borzunov:
@IAL32 By the way, if you are just looking for a way to fetch the current number of active collaborators, you can use the `optimizer.tracker.global_progress.num_peers` property. It is periodically updated basing on the DHT data.

For the reference, see the definition of the [GlobalTrainingProgress](https://github.com/learning-at-home/hivemind/blob/master/hivemind/optim/progress_tracker.py#L25) structure.

borzunov:
Unfortunately, we're not ready to merge this PR in its current form. We propose you to use one of the workarounds discussed above: (a) fetch the number of active peers from the `GlobalTrainingProgress` class or (b) inherit `hivemind.Optimizer` and `TrainingStateAverager` to expose the `gathered_peers` property without changing the code of the library.

Feel free to reach us if you think we've missed something or if you need any help.

PR #503, opened 2022-08-29
Fix a potential deadlock in await_asynchronously with nested locks

justheuristic:
Discovered by @borzunov  when reviewing https://github.com/bigscience-workshop/petals/pull/53

Please see the test description in files_changed for more details

### DIFF
--- a/hivemind/utils/asyncio.py
+++ b/hivemind/utils/asyncio.py
@@ -1,5 +1,7 @@
 import asyncio
 import concurrent.futures
+import multiprocessing as mp
+import os
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import AbstractAsyncContextManager, AbstractContextManager, asynccontextmanager
 from typing import AsyncIterable, AsyncIterator, Awaitable, Callable, Iterable, Optional, Tuple, TypeVar, Union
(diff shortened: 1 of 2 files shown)


hivemind/utils/asyncio.py
```diff
@@ -167,12 +169,24 @@ async def attach_event_on_finished(iterable: AsyncIterable[T], event: asyncio.Ev
...
+    CONTEXT_EXECUTOR = None
+    EXECUTOR_LOCK = mp.Lock()
+
     def __init__(self, context: AbstractContextManager):
         self._context = context
 
+    @classmethod
+    def get_process_wide_executor(cls):
+        if os.getpid() != cls.EXECUTOR_PID:
+            with cls.EXECUTOR_LOCK:
```
borzunov, line 182:
```suggestion
        with cls.EXECUTOR_LOCK:
            if os.getpid() != cls.EXECUTOR_PID:
```

borzunov: Yeah, I agree that the double if is better in this case :)

justheuristic: implemented the latter option with double checked locking

borzunov (approved):

PR #513, opened 2022-10-11
Follow up #501 and #511 with minor fixes

borzunov:
- In `hivemind.Server`, use the graceful shutdown for `ConnectionHandler`
    - Introduced in #501 
- In `hivemind.P2P`, if we are the first peer, skip checking if the provided identity is free
    - Introduced in #511

### DIFF
--- a/hivemind/moe/server/server.py
+++ b/hivemind/moe/server/server.py
@@ -247,10 +247,8 @@ class Server(threading.Thread):
         if self.checkpoint_saver is not None:
             self.checkpoint_saver.start()
 
-        for process in self.conn_handlers:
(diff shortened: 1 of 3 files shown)


hivemind/moe/server/server.py
```diff
@@ -301,11 +298,10 @@ def shutdown(self):
             self.checkpoint_saver.join()
 
         self.dht.shutdown()
-        self.dht.join()
```
borzunov, line 304: `hivemind.DHT` already calls `.join()` in `shutdown()`.

justheuristic (approved):

PR #517, opened 2022-11-01
Fix "unable to open shared memory" while using MPFuture

borzunov:
photo_2022-11-01 06 03 12

Currently, one may sometimes get the "unable to open shared memory" error (see the screenshot) while using `hivemind.MPFuture`. Interestingly, the smaller `HIVEMIND_SHM_BUFFER_SIZE` is, the more often the error occurs (e.g., in Petals, it occurs right after starting the server if `HIVEMIND_SHM_BUFFER_SIZE=2`).

Turns out, it happens when the origin process garbage-collects all instances of MPFuture using the same shmem buffer, then the underlying buffer is freed, and target processes can't reconnect to it anymore when unpickling its instances of MPFuture.

This PR fixes this important issue.

### DIFF
--- a/hivemind/utils/mpfuture.py
+++ b/hivemind/utils/mpfuture.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 import asyncio
 import concurrent.futures._base as base
 import multiprocessing as mp
-import multiprocessing.connection
 import os
+import pickle
 import threading
 import uuid
 from contextlib import nullcontext
@@ -303,7 +303,7 @@ class MPFuture(base.Future, Generic[ResultType]):
     def __getstate__(self):
         return dict(
             _sender_pipe=self._sender_pipe,
-            _shared_state_code=self._shared_state_code,
+            _shared_state_code=pickle.dumps(self._shared_state_code),
             _origin_pid=self._origin_pid,
             _uid=self._uid,
             _use_lock=self._use_lock,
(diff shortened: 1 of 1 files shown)


hivemind/utils/mpfuture.py
```diff
@@ -303,7 +303,7 @@ def __del__(self):
     def __getstate__(self):
         return dict(
             _sender_pipe=self._sender_pipe,
-            _shared_state_code=self._shared_state_code,
+            _shared_state_code=ForkingPickler.dumps(self._shared_state_code).tobytes(),
```
borzunov, line 306: Despite the name, it does work well with all multiprocessing start methods (`fork`, `spawn`, and `forkserver`). This may be tested using this short script:

```python
import multiprocessing as mp
import torch
import time

torch.multiprocessing.set_sharing_strategy("file_system")

def foo(q):
    t = torch.zeros(3).share_memory_()
    q.put(mp.reduction.ForkingPickler.dumps(t).tobytes())
    t[1] = 2

if __name__ == '__main__':
    mp.set_start_method('spawn')
    q = mp.Queue()
    p = mp.Process(target=foo, args=(q,))
    p.start()
    time.sleep(1)
    print(mp.reduction.ForkingPickler.loads(q.get()))
    p.join()
```

borzunov, line 306: Despite the name, it does work well with all multiprocessing start methods (`fork`, `spawn`, and `forkserver`). This may be tested using this short script:

```python
import multiprocessing as mp
import torch
import time
from multiprocessing.reduction import ForkingPickler

torch.multiprocessing.set_sharing_strategy("file_system")

def foo(q):
    t = torch.zeros(3).share_memory_()
    q.put(ForkingPickler.dumps(t).tobytes())
    t[1] = 777

if __name__ == '__main__':
    mp.set_start_method('spawn')
    q = mp.Queue()
    p = mp.Process(target=foo, args=(q,))
    p.start()
    time.sleep(1)
    print(ForkingPickler.loads(q.get()))
    p.join()
```

justheuristic (approved):
Great job! Merge when ready.

PR #518, opened 2022-11-01
Simplify running loop in DHT and Averager

borzunov:
I think there's no need to double-check that we didn't miss anything and the pipe is available to read when asyncio wakes us up. `loop.add_reader` guarantees that we will be awaken iff the pipe has new info.


borzunov:
We need to do a detailed investigation to ensure that this doesn't decrease stability. Closing it since this PR is not needed so much to do this investigation.

PR #521, opened 2022-11-26
Fix MPFuture failing outside inference mode

borzunov:
Fixes this error discovered while working on https://github.com/bigscience-workshop/petals/pull/91:

Screenshot 2022-11-26 at 07 40 22

### DIFF
--- a/hivemind/utils/mpfuture.py
+++ b/hivemind/utils/mpfuture.py
@@ -127,7 +127,8 @@ class MPFuture(base.Future, Generic[ResultType]):
 
     @_state.setter
     def _state(self, new_state: State):
-        self._shared_state_code[...] = ALL_STATES.index(new_state)
+        with torch.inference_mode():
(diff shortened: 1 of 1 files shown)


justheuristic (approved):

borzunov:
Technically, this happens when MPFuture and the underlying shared state tensor are created inside `inference_mode()`, then the state is changed outside of it:

```python
In [1]: import torch

In [2]: with torch.inference_mode():
   ...:     t = torch.zeros(3)
   ...: 

In [3]: t[...] = torch.ones(3)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Input In [3], in <cell line: 1>()
----> 1 t[...] = torch.ones(3)

RuntimeError: Inplace update to inference tensor outside InferenceMode is not allowed.You can make a clone to get a normal tensor before doing inplace update.See https://github.com/pytorch/rfcs/pull/17 for more details.
```

PR #525, opened 2022-11-29
Remove stale PeerIDs in hivemind-dht's routing table

borzunov:
Before this PR, hivemind-dht-based initial peers collected lots of stale PeerIDs and other peers could not actually make DHT queries anymore.

### DIFF
--- a/hivemind/hivemind_cli/run_dht.py
+++ b/hivemind/hivemind_cli/run_dht.py
@@ -1,5 +1,6 @@
 import time
 from argparse import ArgumentParser
+from secrets import token_hex
 
 from hivemind.dht import DHT, DHTNode
 from hivemind.utils.logging import get_logger, use_hivemind_log_handler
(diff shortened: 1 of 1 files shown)


hivemind/hivemind_cli/run_dht.py
```diff
@@ -18,6 +19,9 @@ async def report_status(dht: DHT, node: DHTNode):
     logger.info(f"Local storage contains {len(node.protocol.storage)} keys")
     logger.debug(f"Local storage contents: {node.protocol.storage}")
 
+    # Send a heartbeat to keep routing table healthy (this way, stale PeerIDs will be removed)
```
justheuristic, line 22:
```suggestion
    # Contact peers and keep the routing table healthy (remove stale PeerIDs)
```

justheuristic (approved):

hivemind/hivemind_cli/run_dht.py
justheuristic: Why: it is unclear whether this meets the definition of a heartbeat

Issue #382, opened 2021-09-13
[Feature Request] make DHT resistant to KeyboardInterrupt


justheuristic:
Currently, if you're running hivemind.DHT in jupyter (or an interactive console), you will sometimes need to hit "interrupt" when prototyping. Unfortunately, KeyboardInterrupt will also be caught by the DHT process and will shutdown that DHT instance.

To reproduce:

```python
import time
import hivemind


dht = DHT(start=True)

while True:
    time.sleep(1)
    dht.store('key', 'value', expiration_time=hivemind.get_dht_time() + 999)
    print(end='.', flush=True)
    # [interrupt here via kernel -> interrupt]
    
# and then check:
assert dht.is_alive()
```

Notes:
- Whatever fix we implement must still be able to shutdown DHT if it gets deallocated or when calling .shutdown()
- This problem also applies to DecentralizedAverager (fix or create an issue)

ikmckenz:
I believe this was solved by https://github.com/learning-at-home/hivemind/pull/493, and so the issue can be closed?

borzunov:
#493 fixes it for the DHT but not for `hivemind.DecentralizedAverager`. I'll rename the issue to reflect this, thanks for pointing it out!

borzunov:
I think the fix for `DecentralizedAverager` should be very similar though.

PR #530, opened 2022-12-08
Make DecentralizedAverager resistant to KeyboardInterrupt

ikmckenz:
Very simple one line change, using the same approach as taken by the DHT. Resolves https://github.com/learning-at-home/hivemind/issues/382.

Tested working: 

Screenshot from 2022-12-07 19-04-07


justheuristic (approved):
Thanks for the PR! That should make averager more convenient to use interactively.

PR #532, opened 2022-12-09
Fix typo in beam_search.py

eltociear:
ther -> there

### DIFF
--- a/hivemind/moe/client/beam_search.py
+++ b/hivemind/moe/client/beam_search.py
@@ -171,7 +171,7 @@ class MoEBeamSearcher:
     ) -> Dict[ExpertPrefix, Dict[Coordinate, ExpertInfo]]:
         """
         :param prefixes: a list of prefix for which to find active successor uids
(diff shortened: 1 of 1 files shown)


hivemind/moe/client/beam_search.py
```diff
@@ -171,7 +171,7 @@ def get_active_successors(
     ) -> Dict[ExpertPrefix, Dict[Coordinate, ExpertInfo]]:
         """
         :param prefixes: a list of prefix for which to find active successor uids
-        :param grid_size: if specified, only return successors if ther are in range [0, grid_size)
+        :param grid_size: if specified, only return successors if there are in range [0, grid_size)
```
borzunov, line 174:
```suggestion
        :param grid_size: if specified, only return successors if they are in range [0, grid_size)
```

borzunov:
Hi @eltociear!

Thanks for pointing out the typo. I believe it's better to write `they` here.

borzunov (approved):

PR #541, opened 2023-01-03
Add comment warning for non_blocking + share_memory

justheuristic:
This pull-request does not change code, but explains a certain bug that we (mostly I) keep doing when refactoring task pool.
Most recently, it took me 4 hours to find it in Petals. The problem with this bug is that it can only be observed on GPU, so our CIs will not catch it. Hence, I believe it would be better to state explicitly that why loading batches to runtime is non-blocking, while loading outputs must be blocking.


mryab (approved):
Nice catch!

borzunov (approved):

PR #546, opened 2023-01-06
Import bitsandbytes only if it is used

borzunov:
When bitsandbytes is installed, hivemind always tries to import it even if it doesn't work. This leads to a number of undesirable behaviors:

- `import hivemind` fails if bitsandbytes is installed but fails to resolve CUDA dynamic libraries or load its own dynamic libraries. This happens all the time:
    - on CPU-only hosts with CUDA installed (e.g., Colabs)
    - on hosts with misconfigured CUDA (e.g., most nora envs)
    - on macOS (even when the daemon is compiled correctly and it should work), etc.
- Users see the irrelevant bitsandbytes welcome message even if they don't use it. This message is currently impossible to suppress. Multiple users has reported that this is not expected when they import hivemind/petals.

This PR fixes it, so now hivemind's behavior matches with, e.g., HF transformers (import bitsandbytes only if it's necessary).

### DIFF
--- a/hivemind/compression/quantization.py
+++ b/hivemind/compression/quantization.py
@@ -1,4 +1,3 @@
-import importlib.util
 import math
 import os
 import warnings
@@ -130,13 +129,13 @@ class BlockwiseQuantization(Quantization):
     def quantize(
         self, tensor: torch.Tensor, allow_inplace: bool = False
     ) -> Tuple[np.ndarray, Tuple[np.ndarray, np.ndarray]]:
-        # This runs actual import only on the 1st call, copies references after that
-        from bitsandbytes.functional import quantize_blockwise
-
         try:
-            quantized, (absmax, codebook) = quantize_blockwise(tensor)
-        except NameError:
+            # This runs actual import only on the 1st call, copies references after that
+            from bitsandbytes.functional import quantize_blockwise
+        except ImportError:
             raise ImportError(BNB_MISSING_MESSAGE)
+
+        quantized, (absmax, codebook) = quantize_blockwise(tensor)
         return quantized.numpy(), (absmax.numpy(), codebook.numpy())
 
     def compress(self, tensor: torch.Tensor, info: CompressionInfo, allow_inplace: bool = False) -> runtime_pb2.Tensor:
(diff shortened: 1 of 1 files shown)


hivemind/compression/quantization.py
```diff
@@ -129,11 +127,17 @@ class BlockwiseQuantization(Quantization):
     compression_type = runtime_pb2.BLOCKWISE_8BIT
     codebook_dtype, indices_dtype = np.float32, np.uint8
 
+    def __init__(self):
+        from bitsandbytes.functional import dequantize_blockwise, quantize_blockwise
```
borzunov, line 131: If run for the 2nd time, it won't actually import things again, just copy the references.

hivemind/compression/quantization.py
```diff
@@ -129,11 +127,19 @@ class BlockwiseQuantization(Quantization):
     compression_type = runtime_pb2.BLOCKWISE_8BIT
     codebook_dtype, indices_dtype = np.float32, np.uint8
 
+    def __init__(self):
+        super().__init__()
+
+        from bitsandbytes.functional import dequantize_blockwise, quantize_blockwise
```
borzunov, line 133: If it runs for the 2nd time, it won't actually import things again, just copy the references.

hivemind/compression/quantization.py
```diff
@@ -9,13 +9,11 @@
 import numpy as np
 import torch
 
-if importlib.util.find_spec("bitsandbytes") is not None:
```
mryab, line 12: importlib.util is no longer used, so its import can be removed

hivemind/compression/quantization.py
```diff
@@ -132,6 +130,9 @@ class BlockwiseQuantization(Quantization):
     def quantize(
         self, tensor: torch.Tensor, allow_inplace: bool = False
     ) -> Tuple[np.ndarray, Tuple[np.ndarray, np.ndarray]]:
+        # This runs actual import only on the 1st call, copies references after that
+        from bitsandbytes.functional import quantize_blockwise
```
mryab, line 134: Since this is no longer under a conditional import (and will thus cause ImportError instead of NameError), can we wrap this statement instead of the `quantize_blockwise` call below with try-except? The same applies to `extract`

mryab (approved):

PR #553, opened 2023-02-07
Improve bfloat16 serialization (backward compatible)

justheuristic:
- added bfloat16 serialization that sends 2 bytes per value (previously, we sent 4);
- changed de-serialization code so it supports both modes of serialization.
- the new mode can be enabled via `export LEGACY_BFLOAT16=0`
- tested in pytorch 1.12 and 1.13

Next minor release: switch to new serialization mode
Next major release: remove legacy mode

### DIFF
--- a/.github/workflows/run-benchmarks.yml
+++ b/.github/workflows/run-benchmarks.yml
@@ -28,7 +28,7 @@ jobs:
           pip install -r requirements-dev.txt
       - name: Build bitsandbytes
         run: |
-          pip install bitsandbytes==0.34.0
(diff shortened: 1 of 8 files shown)


borzunov:
I would suggest to:

- Rename `LEGACY_BFLOAT16` to `USE_LEGACY_BFLOAT16` (both the env var and the constant), so that it's clear that it is a bool flag.
- Remove the `LEGACY_PYTORCH` constant and use `hasattr()` instead, so we do not depend on knowing the exact version where these fields were renamed.

borzunov (approved):

PR #557, opened 2023-03-11
Remove direct coroutine call

srogatch:
Python 3.11 forbids direct calls to coroutines.


mryab (approved):
LGTM, @justheuristic do you see any downsides to this? The docs for `asyncio.wait_for` mention automatic conversion of coroutines to tasks — perhaps it means that this change is safe

borzunov (approved):
Thanks for the PR!

I'll try adding 3.11 to the CI to see if there're other incompatibilities, see also #558.

PR #558, opened 2023-03-11
Require torch<2.0 until 2.0 is supported, add Python 3.10 to CI

borzunov:
PyTorch 2.0 is not supported yet, it's support will be added in #559 (there are multiple issues to resolve). Until then, we need to require `torch<2.0.0` (otherwise 2.0.0 is installed, so CI doesn't work right now).

This PR also adds Python 3.10 to the CI.

### DIFF
--- a/.github/workflows/run-benchmarks.yml
+++ b/.github/workflows/run-benchmarks.yml
@@ -11,9 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.9
       - name: Cache dependencies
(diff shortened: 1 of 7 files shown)


.github/workflows/run-benchmarks.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         with:
           python-version: 3.9
       - name: Cache dependencies
-        uses: actions/cache@v2
+        uses: actions/cache@v3
```
borzunov, line 20: v2 is deprecated.

requirements-dev.txt
```diff
@@ -5,7 +5,6 @@ pytest-cov
 coverage==6.0.2  # see https://github.com/pytest-dev/pytest-cov/issues/520
 tqdm
 scikit-learn
-torchvision
```
borzunov, line 8: Seems like an excess dependency.

If present, it forces to upgrade torch to 2.0.0 during `pip install -r requirements-dev.txt`, which breaks installation from source (`build_and_test_p2pd` in CI).

mryab (approved):

PR #559, opened 2023-03-16
Support PyTorch 2.0.0

justheuristic:


.github/workflows/run-tests.yml
```diff
@@ -10,8 +10,9 @@ jobs:
 
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        python-version: [ 3.7, 3.8, 3.9 ]
+        python-version: [ 3.8, 3.9, 3.10 ]
```
borzunov, line 15:
```suggestion
        python-version: [ '3.8', '3.9', '3.10' ]
```

Now Python 3.1 is being tested

borzunov:
I've rebased this PR to the master branch after #558 was merged.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 PyYAML
-torch>=1.9.0,<2.0.0
+torch>=1.9.0
```
borzunov, line 2: We'll have torch 1.13.0 on Python 3.7, torch 2.0.0 on Python 3.8-3.10. This allows us to test both torch 1.13.0 and 2.0.0.

hivemind/optim/optimizer.py
```diff
@@ -634,7 +634,7 @@ def _load_local_gradients_into_optimizer(self):
         # - if not offload_optimizer, we must un-scale gradients (divide them by the number of accumulation steps)
         self._load_averaged_gradients_into_optimizer_()
 
-    def zero_grad(self, set_to_none: bool = False):
+    def zero_grad(self, set_to_none: bool = True):
```
borzunov, line 637: `set_to_none=True` by default in PyTorch 2.0, I suggest to propagate it here.

hivemind/optim/optimizer.py
```diff
@@ -634,7 +638,9 @@ def _load_local_gradients_into_optimizer(self):
         # - if not offload_optimizer, we must un-scale gradients (divide them by the number of accumulation steps)
         self._load_averaged_gradients_into_optimizer_()
 
-    def zero_grad(self, set_to_none: bool = False):
+    _SET_TO_NONE_DEFAULT = Version(torch.__version__).major >= 2
+
+    def zero_grad(self, set_to_none: bool = _SET_TO_NONE_DEFAULT):
```
borzunov, line 643: So it's consistent with a torch-wide default.

justheuristic (commented):
LGTM

tests/test_optimizer.py
```diff
@@ -151,7 +151,9 @@ def test_state_averager(offload_optimizer: bool, reuse_tensors: bool, sync_epoch
         F.mse_loss(model2(x), -torch.ones(3)).backward()
         avgr2.step(optimizer_step=True, zero_grad=True, averaging_round=(step == 10), delay_averaging=False)
 
-    assert torch.all(model1.weight.grad == 0) and torch.all(model2.weight.grad == 0), "zero grad did not trigger"
+    assert (model1.weight.grad is None or torch.all(model1.weight.grad == 0)) and (
+        model2.weight.grad is None or torch.all(model2.weight.grad == 0)
+    ), "zero grad did not trigger"
```
mryab, line 156: Can we also use `Version(torch.__version__).major >= 2` here? (or at least import _SET_TO_NONE_DEFAULT) 

This way, the test won't have a bug where we always set the grad to None regardless of the PyTorch version

mryab (approved):

PR #586, opened 2023-08-24
Use proper p2pd binary on macOS

borzunov:
This is enough for Petals to work (and pass its tests on macOS).

Making hivemind tests work on macOS turns out to be far more difficult due to lots of multiprocessing magic there. I suggest to work on that in a separate PR (see #585).

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,8 +81,8 @@ of [Go toolchain](https://golang.org/doc/install) (1.15 or 1.16 are supported).
 
 - __Linux__ is the default OS for which hivemind is developed and tested. We recommend Ubuntu 18.04+ (64-bit), but
   other 64-bit distros should work as well. Legacy 32-bit is not recommended.
-- __macOS__ is also supported, but some features may not work properly.
-  If you have issues, you can run hivemind using [Docker](https://docs.docker.com/desktop/mac/install/).
+- __macOS__ is partially supported.
+  If you have issues, you can run hivemind using [Docker](https://docs.docker.com/desktop/mac/install/) instead.
   We recommend using [our Docker image](https://hub.docker.com/r/learningathome/hivemind).
 - __Windows 10+ (experimental)__ can run hivemind
   using [WSL](https://docs.microsoft.com/ru-ru/windows/wsl/install-win10). You can configure WSL to use GPU by
--- a/README.md
+++ b/README.md
@@ -81,8 +81,8 @@ of [Go toolchain](https://golang.org/doc/install) (1.15 or 1.16 are supported).
 
 - __Linux__ is the default OS for which hivemind is developed and tested. We recommend Ubuntu 18.04+ (64-bit), but
   other 64-bit distros should work as well. Legacy 32-bit is not recommended.
-- __macOS 10.x__ can run hivemind using [Docker](https://docs.docker.com/desktop/mac/install/).
-  We recommend using [our Docker image](https://hub.docker.com/r/learningathome/hivemind).
+- __macOS__ is also supported, but some features may not work properly.
+  If you have issues, you can run hivemind using [our Docker image](https://hub.docker.com/r/learningathome/hivemind).
 - __Windows 10+ (experimental)__ can run hivemind
   using [WSL](https://docs.microsoft.com/ru-ru/windows/wsl/install-win10). You can configure WSL to use GPU by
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -13,14 +14,15 @@
...
 from setuptools.command.develop import develop
 
-P2PD_VERSION = "v0.3.16"
+P2PD_VERSION = "v0.3.17"
 
 P2PD_SOURCE_URL = f"https://github.com/learning-at-home/go-libp2p-daemon/archive/refs/tags/{P2PD_VERSION}.tar.gz"
 P2PD_BINARY_URL = f"https://github.com/learning-at-home/go-libp2p-daemon/releases/download/{P2PD_VERSION}/"
 
 # The value is sha256 of the binary from the release page
-EXECUTABLES = {
```
borzunov, line 22: I made `EXECUTABLES` when we also wanted to ship `p2p-keygen` binary. We now moved away from it, since `p2p-keygen` functionality can be implemented in 3 lines of Python.

README.md
```diff
@@ -81,8 +81,8 @@ of [Go toolchain](https://golang.org/doc/install) (1.15 or 1.16 are supported).
 
 - __Linux__ is the default OS for which hivemind is developed and tested. We recommend Ubuntu 18.04+ (64-bit), but
   other 64-bit distros should work as well. Legacy 32-bit is not recommended.
-- __macOS 10.x__ can run hivemind using [Docker](https://docs.docker.com/desktop/mac/install/).
-  We recommend using [our Docker image](https://hub.docker.com/r/learningathome/hivemind).
+- __macOS__ is also supported, but some features may not work properly.
+  If you have issues, you can run hivemind using [our Docker image](https://hub.docker.com/r/learningathome/hivemind).
```
mryab, line 85: Maybe it's best to keep the part about Docker in general, as some people would like to build their own image

mryab, line 85:
```suggestion
  If you have issues, you can run hivemind using [Docker](https://docs.docker.com/desktop/mac/install/).
  We recommend using [our Docker image](https://hub.docker.com/r/learningathome/hivemind).
```

setup.py
```diff
@@ -71,31 +73,31 @@ def build_p2p_daemon():
...
-    for executable, expected_hash in EXECUTABLES.items():
-        binary_path = os.path.join(here, "hivemind", "hivemind_cli", executable)
+    binary_path = os.path.join(here, "hivemind", "hivemind_cli", "p2pd")
+    os_name = platform.system().lower()
+    expected_hash = P2P_BINARY_HASH[os_name]
 
-        if sha256(binary_path) != expected_hash:
-            binary_url = os.path.join(P2PD_BINARY_URL, executable)
-            print(f"Downloading {binary_url}")
+    if sha256(binary_path) != expected_hash:
```
mryab, line 89: Just to clarify: do we verify it this way to automatically download the correct binary for a non-Linux system? Would it be more clear if we simply checked that the binary exists?

mryab (approved):

README.md
```diff
@@ -81,8 +81,8 @@ of [Go toolchain](https://golang.org/doc/install) (1.15 or 1.16 are supported).
 
 - __Linux__ is the default OS for which hivemind is developed and tested. We recommend Ubuntu 18.04+ (64-bit), but
   other 64-bit distros should work as well. Legacy 32-bit is not recommended.
-- __macOS 10.x__ can run hivemind using [Docker](https://docs.docker.com/desktop/mac/install/).
-  We recommend using [our Docker image](https://hub.docker.com/r/learningathome/hivemind).
+- __macOS__ is also supported, but some features may not work properly.
```
justheuristic, line 84: Nit: the phrase "supported, but some features may not work properly" may rise some eyebrows. Consider "macOS is partially supported" or something in the spirit of "most (but not all) of the main features work"

Not willing to fight for it. feel free to ignore if you disagree

justheuristic (approved):

setup.py
borzunov: Sometimes devs replace the binary for testing purposes (e.g., to test an update in libp2p-daemon). As far as I understand, this code allows to replace the binary to the standard one when you reinstall hivemind (since the expected behavior in this case is to get a clean installation).

README.md
borzunov: +

PR #587, opened 2023-08-25
[minor] allow overriding args/kwargs behavior in Runtime

justheuristic:
This PR extracts a part of runtime that is responsible for calling pool's process func.
This is to allow developer to extend this behavior in ~~petals~~ downstream applications without having to rewrite the whole runtime.run method.

### DIFF
--- a/hivemind/moe/server/runtime.py
+++ b/hivemind/moe/server/runtime.py
@@ -7,7 +7,7 @@ from queue import SimpleQueue
 from selectors import EVENT_READ, DefaultSelector
 from statistics import mean
 from time import perf_counter
-from typing import Dict, NamedTuple, Optional
+from typing import Any, Dict, NamedTuple, Optional, Tuple
 
 import torch
 from prefetch_generator import BackgroundGenerator
@@ -105,7 +105,7 @@ class Runtime(threading.Thread):
                     self.shutdown()
 
     def process_batch(self, pool: TaskPoolBase, batch_index: int, *batch: torch.Tensor) -> Tuple[Any, int]:
-        """process one batch of tasks from a given pool, return a batch of results and task size (complexity)"""
+        """process one batch of tasks from a given pool, return a batch of results and total batch size"""
         outputs = pool.process_func(*batch)
         return outputs, outputs[0].size(0)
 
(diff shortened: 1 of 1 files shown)


hivemind/moe/server/runtime.py
```diff
@@ -108,6 +99,17 @@ def run(self):
...
 
+    def process_batch(self, pool: TaskPoolBase, batch_index: int, *batch: torch.Tensor):
+        """process one batch of tasks from a given pool, return a batch of results"""
+        start = time()
+        outputs = pool.process_func(*batch)
+        batch_processing_time = time() - start
+        batch_size = outputs[0].size(0)
+        logger.debug(f"Pool {pool.name}: batch {batch_index} processed, size {batch_size}")
+        if self.stats_report_interval is not None:
+            self.stats_reporter.report_stats(pool.name, batch_size, batch_processing_time)
```
mryab, line 110: I think it might lead to unintended "0 batches processed" log entries if the user overrides this without carefully considering the original function. Best to leave only the batch size computation inside the function and keep all the logging/time measuring logic outside the function

justheuristic: good catch, fixed it now, please take another look

hivemind/moe/server/runtime.py
```diff
@@ -108,6 +104,11 @@ def run(self):
                 if not self.shutdown_trigger.is_set():
                     self.shutdown()
 
+    def process_batch(self, pool: TaskPoolBase, batch_index: int, *batch: torch.Tensor) -> Tuple[Any, int]:
```
mryab, line 107: I think `outputs` is `List[torch.Tensor]` due to https://github.com/learning-at-home/hivemind/blob/master/hivemind/moe/server/task_pool.py#L241

mryab (approved):

borzunov (approved):

hivemind/moe/server/runtime.py
```diff
@@ -108,6 +105,11 @@ def run(self):
...
```
borzunov, line 108: `batch_index` is not used, is it okay?

justheuristic: yep, it is meant to be exposed for subclassing

PR #591, opened 2023-09-05
Fix edge cases in (de)serialize_torch_tensor

justheuristic:
During an earlier patch, we lost the requires_grad property during serialize_torch_tensor. This PR adds it back.

### DIFF
--- a/hivemind/compression/floating.py
+++ b/hivemind/compression/floating.py
@@ -12,10 +12,11 @@ class Float16Compression(CompressionBase):
     FP16_MIN, FP16_MAX = torch.finfo(torch.float16).min, torch.finfo(torch.float16).max
 
     def compress(self, tensor: torch.Tensor, info: CompressionInfo, allow_inplace: bool = False) -> runtime_pb2.Tensor:
+        assert torch.is_floating_point(tensor) and tensor.dtype != torch.bfloat16
         requires_grad = tensor.requires_grad
+        tensor = tensor.detach().cpu()
         dtype_name = tensor.numpy().dtype.name
-        tensor = tensor.detach().cpu().float()
-        tensor = tensor if allow_inplace else tensor.clone()
+        tensor = tensor.to(torch.float32, copy=not allow_inplace)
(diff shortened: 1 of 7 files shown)


mryab (approved):

hivemind/compression/floating.py
```diff
@@ -12,22 +12,28 @@ class Float16Compression(CompressionBase):
     FP16_MIN, FP16_MAX = torch.finfo(torch.float16).min, torch.finfo(torch.float16).max
 
     def compress(self, tensor: torch.Tensor, info: CompressionInfo, allow_inplace: bool = False) -> runtime_pb2.Tensor:
+        assert torch.is_floating_point(tensor) and tensor.dtype != torch.bfloat16
```
mryab, line 15: Is there a reason why we should fail with an error in case of bf16 inputs? It is indeed not sensible, but  if the user wants to do so, it's probably better to issue a warning instead of flat out refusing to pass that through quantization

hivemind/compression/quantization.py
```diff
@@ -135,14 +138,15 @@ def quantize(
         except ImportError:
             raise ImportError(BNB_MISSING_MESSAGE)
 
-        quantized, (absmax, codebook) = quantize_blockwise(tensor)
+        quantized, (absmax, codebook, *extra_params) = quantize_blockwise(tensor, blocksize=4096, nested=False)
+        assert tuple(extra_params) == (4096, False, tensor.dtype, None, None)  # blocksize, nested, dtype, offset, s2
```
mryab, line 142: Maybe we can make that tuple on the right a module-level constant? It's used twice in the code, better to make it clear we're using some predefined values

mryab (approved):

hivemind/compression/floating.py
justheuristic: added ValueError with a more user-legible reason

hivemind/compression/quantization.py
justheuristic: done

Issue #592, opened 2023-10-03
[BUG] Unable to start hivemind server when using gradient clipping 


SamAg19:
**Describe the bug**
Trying to start up a hivemind server enabling gradient clipping `--clip_grad_norm` results in an error in https://github.com/learning-at-home/hivemind/blob/master/hivemind/moe/server/layers/optim.py due a setter property not being present for `defaults` in the file

**To Reproduce**
After starting up a dht peer, ran the following command:
Command:
`hivemind-server --expert_cls ffn --hidden_dim 512 --num_experts 5 --expert_pattern "expert.[0:5]" --initial_peers /ip4/172.29.152.117/tcp/45239/p2p/12D3KooWCAArT2ncW2H9BRoneEri4b2xh99SdMgXPZEuzkobx3bY --clip_grad_norm 1.0`

Output:
```
Sep 28 17:55:49.317 [INFO] Running DHT node on ['/ip4/172.29.152.117/tcp/38207/p2p/12D3KooWPEZEe796QZDZn8qCXXnC4RDxqkz6ADbAQ4fXLCeiMj5s', '/ip4/127.0.0.1/tcp/38207/p2p/12D3KooWPEZEe796QZDZn8qCXXnC4RDxqkz6ADbAQ4fXLCeiMj5s'], initial peers = ['/ip4/172.29.152.117/tcp/45239/p2p/12D3KooWCAArT2ncW2H9BRoneEri4b2xh99SdMgXPZEuzkobx3bY']
Sep 28 17:55:49.317 [INFO] Generating 5 expert uids from pattern expert.[0:5]
Traceback (most recent call last):
  File "/home/samag19/miniconda3/envs/py311/bin/hivemind-server", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/samag19/miniconda3/envs/py311/lib/python3.11/site-packages/hivemind/hivemind_cli/run_server.py", line 105, in main
    server = Server.create(**args, optim_cls=optim_cls, start=True, compression=compression)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/samag19/miniconda3/envs/py311/lib/python3.11/site-packages/hivemind/moe/server/server.py", line 205, in create
    optimizer = ClippingWrapper(optimizer, clip_grad_norm)
[...]

PR #593, opened 2023-10-08
Fix OptimizerWrapper creation, test gradient clipping

mryab:
Currently, OptimizerWrapper fails when creating an optimizer: it attempts to assign the property `defaults` which has no setter defined. The easiest way to address this is to remove inheritance from OptimizerWrapper while keeping the same API for functions required by hivemind.server: as this is an internal component, downstream users are unlikely to be negatively affected.

Fixes https://github.com/learning-at-home/hivemind/issues/592

### DIFF
--- a/hivemind/moe/server/layers/optim.py
+++ b/hivemind/moe/server/layers/optim.py
@@ -1,11 +1,10 @@
 import torch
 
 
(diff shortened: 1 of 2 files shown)


justheuristic (approved):
lgtm, thanks

PR #595, opened 2023-10-08
Fix deprecations and update dependencies for examples/albert

mryab:
The current Docker build fails due to old versions of transformers/datasets depending on an old version of tokenizers, which does not compile in the current environment of the Docker image. The easiest way to address the issue appears to be the bump of dependencies.

I have tested running the example on 2 peers with albert-base-v2, and the code runs and even starts to converge:
image

Also, in recent builds, pip started issuing an error when trying to pass --no-use-pep517 without setuptools and wheel installed (see [this build](https://github.com/learning-at-home/hivemind/actions/runs/6448925844/job/17506719837?pr=593#step:7:16) as an example). This PR addresses the issue by installing setuptools and wheel before calling pip install.


justheuristic (approved):
thanks for fixing it!

PR #598, opened 2023-11-21
Bump p2pd version

dvmazur:


justheuristic:
Note: we had to fork py-multiaddr because the previous version required py-cid that caused a dependency conflict

justheuristic (approved):
LGTM

justheuristic:
The only caveat is that if one person runs 2+ servers on one machine with the same identity file AND at least one of them is old version and at least one is new, THEN the duplicate check may not catch this.

I've asked @borzunov and @mryab  on their opinion. If they approve, let's merge this

requirements.txt
```diff
@@ -9,8 +9,9 @@ uvloop>=0.14.0
...
 protobuf>=3.12.2
 configargparse>=1.2.3
-multiaddr>=0.0.9
-pymultihash>=0.8.2
+py-multihash>=0.2.3
+multiaddr @ git+https://github.com/multiformats/py-multiaddr.git
 cryptography>=3.4.6
 pydantic>=1.8.1,<2.0
 packaging>=20.9
+py-cid>=0.3.0
```
mryab, line 17: If this package is not directly used, there is no need to specify it as a dependency

dvmazur:
@justheuristic solemnly swears to
- show a proof that forwarding kwargs works in a **basic test that is easy to follow**
- show an example of how new clients can work with old servers, at least for all supported basic ops
- show an example of how old clients can work with new servers

PR #601, opened 2023-11-30
Remove pymultihash from datastructures.py

dvmazur:
This PR removes [pymultihash](https://pypi.org/project/pymultihash/) from the codebase and dependencies as it is
in conflict with [py-multihash](https://pypi.org/project/py-multihash/).


hivemind/p2p/p2p_daemon_bindings/datastructures.py
```diff
@@ -17,25 +17,9 @@
 # NOTE: On inlining...
 # See: https://github.com/libp2p/specs/issues/138
 # NOTE: enabling to be interoperable w/ the Go implementation
-ENABLE_INLINING = True
+# ENABLE_INLINING = False
```
justheuristic, line 20: please rm this

Issue #597, opened 2023-11-17
pydantic < 2.0.0 is starting to conflict with other dependencies


Vectorrent:
**Describe the bug**
The version constraint of pydantic < 2.0.0 is conflicting with other dependencies.

**To Reproduce**
You could try to install both `hivemind` and [horde-sdk](https://github.com/Haidra-Org/horde-sdk) with pip. It will fail with dependency conflicts. If you try to install `horde-sdk` in a venv, it breaks `hivemind`:
```
vtx-lab-1  |   File "/usr/local/lib/python3.10/dist-packages/hivemind/dht/schema.py", line 168, in __get_validators__
vtx-lab-1  |     yield from super().__get_validators__()
vtx-lab-1  | AttributeError: 'super' object has no attribute '__get_validators__'
```

**Note**
At some point, I may attempt the fix myself. The problem is, I've never used pydantic, and I don't know what kind of cascading effects it may have on hivemind's dependent packages. For now, I'm stuck - and I would definitely appreciate a fix from someone else if possible.

**Environment**
* python 3.10.12
* hivemind 1.1.5 (as required by [hivemind-lightning](https://github.com/Lightning-Universe/lightning-Hivemind/blob/main/requirements.txt))
* output from pytorch [environment collection script](https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py):
```
PyTorch version: 2.1.1+cu121
Is debug build: False
CUDA used to build PyTorch: 12.1
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.3 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version: 14.0.0-1ubuntu1.1
CMake version: Could not collect
Libc version: glibc-2.35

Python version: 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] (64-bit runtime)
Python platform: Linux-6.5.9-arch2-1-x86_64-with-glibc2.35
Is CUDA available: True
CUDA runtime version: 12.2.140
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: GPU 0: NVIDIA GeForce GTX 1070
Nvidia driver version: 535.113.01
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                       x86_64
[...]

PR #611, opened 2024-05-13
fix dependency for the albert example

samsja:
# Context

this is just a small PR to fix some dependency with the albert example

PR #613, opened 2024-06-04
Migrate to pydantic >= 2.0

Vectorrent:
This is a first pass at upgrading Pydantic to 2.7.3, as discussed in #597. Fortunately, Pydantic is barely used in Hivemind - so this should be fairly straightforward.

So far, I have extended the `BaseModel` class, in order to define a `Config` object with `arbitrary_types_allowed = True`. From what I can tell, this was the simplest way to deal with our validator's regex handling. The `__get_validators__` method was removed in 2.0, and this function needed some work.

I also removed the `_patch_schema` calls, because 1) `field.required` was changed to `field.is_required` in 2.0, and 2) `field.is_required` is now read-only. I couldn't figure out how to set it... and the code seems to work with out it? So is it really necessary here?

DO NOT MERGE; this is an initial attempt. Within the next few days, I will test this in a multi-node environment, and I will test it with Petals as well.

Until then, this version of the code seems to work well. I wanted to open the PR, to get us all started.


Vectorrent:
This PR is as far as I can reasonably take it. I fixed the linting errors introduced by my changes (but not the ones already existing in the master branch), and I reverted Pydantic to 2.5.3 (to make it compatible with CI environment). That said, Hivemind is also compatible with 2.7.3.

I tested this change with a multi-node training run, and it works well. I also tested this with Petals, and was able to run inference there as well.

I'd say this is ready for merge.

Vectorrent:
It looks like #612 might fix the errors we're seeing in tests here.

justheuristic:
NB: We'll need to update hivemind to fix incompatibility with the latest torch update to so we can test this (see failed tests). I intend to do this as soon as I have some bandwidth, but if someone's willing to do this earlier, please do.

Vectorrent:
So, it appears that `torch.cuda.amp` was deprecated, and everything was moved to `torch.amp`. So, I fixed that.

Strangely, some of these tests will install torch 2.3.1, while others are installing 1.9.0 (and thus, failing). Not sure if that is intentional, but I'll keep digging.

samsja:
> It looks like #612 might fix the errors we're seeing in tests here.

yes it should fix all ci error

samsja:
> NB: We'll need to update hivemind to fix incompatibility with the latest torch update to so we can test this (see failed tests). I intend to do this as soon as I have some bandwidth, but if someone's willing to do this earlier, please do.

fy: https://github.com/learning-at-home/hivemind/pull/612

Vectorrent:
It looks like mine and @samsja's PR are addressing the same bug, so we should only merge one.

That said, I have no idea why these tests are failing right now, while theirs worked. All I'm seeing is `Error: The operation was canceled.`. Tests are not failing with my code; they are simply hanging forever, and cancelled after a while.

Both PRs still have the same bug, relating to the Albert test:
```
#15 20.13   × python setup.py egg_info did not run successfully.
#15 20.13   │ exit code: 1
#15 20.13   ╰─> [6 lines of output]
#15 20.13       Traceback (most recent call last):
#15 20.13         File "<string>", line 2, in <module>
#15 20.13         File "<pip-setuptools-caller>", line 34, in <module>
#15 20.13         File "/tmp/pip-install-_8jw_ha0/pathtools_37faf27303d54936a074efc743740ba1/setup.py", line 25, in <module>
#15 20.13           import imp
#15 20.13       ModuleNotFoundError: No module named 'imp'
#15 20.13       [end of output]
```
Apparently, the "imp" module was removed in Python 3.12, so I'm sure that's where the problem lies. I was able to reproduce this bug in Docker locally, so I'll spend some time offline troubleshooting.

Vectorrent:
It seems that `tests/test_optimizer.py` is hanging at `test_progress_tracker()`. Since that code references `BytesWithPublicKey`, and my Pydantic changes touched that code - it's probably related. I just don't know why it hangs without doing anything.

If anyone with Pydantic experience wants to jump in, I'd appreciate it.

samsja:
> It seems that `tests/test_optimizer.py` is hanging at `test_progress_tracker()`. Since that code references `BytesWithPublicKey`, and my Pydantic changes touched that code - it's probably related. I just don't know why it hangs without doing anything.
> 
> If anyone with Pydantic experience wants to jump in, I'd appreciate it.

I just tried your PR with pydantic 2.5.3 and 2.7.3. Seems to be working fine, at least the test is not hanging for me. Happy to help if you hint me on how to reproduce the hanging behavior

Vectorrent:
I made some progress, fixed some Pydantic deprecations, but I've been stumped by this one:
```
         for schema in self._schemas:
            try:
                parsed_record = schema.model_validate(record)
            except pydantic.ValidationError as e:
                if not self._is_failed_due_to_extra_field(e):
                    validation_errors.append(e)
                continue
    
>           parsed_value = parsed_record.dict(by_alias=True)[field_name]
E           KeyError: 'some_field'
```

mryab:
I started digging into this as well, and the easiest way to triage the problem seems to be [replacing](https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features) all `pydantic` imports with `pydantic.v1` imports. Have you tried that approach? I'd be ok with merging such a change in case you were hesitating

Vectorrent:
I didn't even realize `pydantic.v1` was an API. Regardless, I made it this far... I might give it a few more days of trying. There's only a couple tests failing now. But if that doesn't work, I'll do the v1 thing soon. Thanks!

Vectorrent:
I reverted to v1 API, since v2 would probably break Pydantic usage in Petals as well... and I don't want to deal with that.

It seems that most tests are passing now. I have no idea why some of them are being cancelled right now, but it seems like they're timing-out. And these failures are not even consistent; whereas 3.8 was successfully on the previous run, it timed-out on this attempt. Not sure what's happening there, but it looks like a Github issue more than anything.

Vectorrent:
These tests keep failing with random, transient errors. What works on this run, will fail on the next. Hivemind is just being unreliable right now.

Vectorrent:
I'd recommend merging, if you're okay with this code. Tests keep failing, but they are failing with transient errors that were probably not introduced by the Pydantic update:
```
test_averaging.py::test_allreduce_once[0-0] PASSED                       [ 23%]
test_averaging.py::test_allreduce_once[0-1] PASSED                       [ 23%]
test_averaging.py::test_allreduce_once[0-2] PASSED                       [ 24%]
test_averaging.py::test_allreduce_once[1-0] PASSED                       [ 24%]
test_averaging.py::test_allreduce_once[1-1] PASSED                       [ 24%]
test_averaging.py::test_allreduce_once[1-2] PASSED                       [ 25%]
test_averaging.py::test_allreduce_once[2-0] PASSED                       [ 25%]
test_averaging.py::test_allreduce_once[2-1] PASSED                       [ 25%]
test_averaging.py::test_allreduce_once[2-2] PASSED                       [ 25%]
```
Stuff like this is just a quirk of Hivemind. `all_reduce` can execute successfully 9 times in a row, before randomly hanging forever (until the test times-out). There is a deeper issue here, but one that should probably be fixed in a different PR.

justheuristic (approved):
LGTM

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Learning@home authors and collaborators
+Copyright (c) 2024 Learning@home authors and collaborators
```
mryab, line 3: Let's keep the scope of changes limited to the purpose of the PR, please. All other changes can be addressed in parallel with smaller pull requests

docker-compose.yml
mryab, line 1: Unless this is necessary for the PR to work, can you remove this?

mryab (changes requested):
Thanks for the contribution! However, before merging, please reduce the diff of the PR to make sure all changes are Pydantic-related

LICENSE
Vectorrent: I changed this so that tests would re-run, last night. Reverted.

docker-compose.yml
Vectorrent: It was a convenience, for building and running tests locally without headache. I'll remove it though.

tests/test_utils/p2p_daemon.py
```diff
@@ -5,16 +5,24 @@
...
 
 from hivemind.p2p.p2p_daemon_bindings.p2pclient import Client
 
 from test_utils.networking import get_free_port
 
 TIMEOUT_DURATION = 30  # seconds
+
+# hivemind_dir = Path(__file__).parent / "hivemind" / "hivemind_cli"
+
+# P2PD_PATH = str(hivemind_dir / "p2pd")
```
mryab, line 22: Please remove any code that is commented out

tests/test_dht_schema.py
```diff
@@ -12,9 +12,9 @@
 
 
 class SampleSchema(BaseModel):
-    experiment_name: bytes
-    n_batches: Dict[bytes, conint(ge=0, strict=True)]
-    signed_data: Dict[BytesWithPublicKey, bytes]
+    experiment_name: bytes = None
+    n_batches: Dict[bytes, conint(ge=0, strict=True)] = None
+    signed_data: Dict[BytesWithPublicKey, bytes] = None
```
mryab, line 17: Unless these changes are required, can you roll them back as well?

examples/albert/requirements.txt
```diff
@@ -1,7 +1,7 @@
 transformers~=4.6
 datasets~=1.5
 torch_optimizer==0.1.0
-wandb==0.10.26
+wandb==0.17.1
```
mryab, line 4: This does not seem to be relevant to the Pydantic upgrade :)

benchmarks/benchmark_optimizer.py
```diff
@@ -98,7 +98,7 @@ def run_trainer(batch_size: int, batch_time: float, client_mode: bool, verbose:
             grad_scaler = hivemind.GradScaler()
         else:
             # check that hivemind.Optimizer supports regular PyTorch grad scaler as well
-            grad_scaler = torch.cuda.amp.GradScaler(enabled=args.use_amp)
+            grad_scaler = torch.amp.GradScaler(enabled=args.use_amp)
```
mryab, line 101: That is a good catch, but can you make it a separate PR to make sure the scope of changes for every commit to main is clear and focused?

requirements.txt
```diff
@@ -12,5 +12,5 @@ configargparse>=1.2.3
 py-multihash>=0.2.3
 multiaddr @ git+https://github.com/multiformats/py-multiaddr.git@e01dbd38f2c0464c0f78b556691d655265018cce
 cryptography>=3.4.6
-pydantic>=1.8.1,<2.0
+pydantic>=2.5.3
```
mryab, line 15: This version is quite recent, can you provide any reasoning behind how it was chosen? Maybe we can bump to just 2.0? Ideally, we should even keep backwards compatibility with older versions

.github/workflows/run-tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.7', '3.8', '3.9', '3.10', '3.11' ]
+        python-version: [ '3.7', '3.8', '3.9', '3.10', '3.11', '3.12' ]
```
mryab, line 14: This needs to coincide with updating setup.py, so I would also suggest to keep it in a separate pull request

mryab (changes requested):
Thank you again for the pull request! I believe we are close to merging, but there is still a small list of changes that need to be made before this

requirements.txt
Vectorrent: I chose this version because it was the highest possible version I could use, which was still compatible with all of the tests. Remember: the whole reason we're upgrading Pydantic is because this old version has been conflicting with other dependencies, in other projects.

We can revert to 2.0.0 if you still want me to do that, though 2.5.3 seems to be working fine.

Vectorrent:
#619 and #620 to fix these errors again

examples/albert/requirements.txt
samsja: feel free to merge this pr for the albert example https://github.com/learning-at-home/hivemind/pull/611

requirements.txt
samsja: IMO, If the code is still compatible with pydantic v1 it might be worth it to just do `pydantic>1.8.1`

Vectorrent:
@mryab - is there anything else you need from me? Between the 3 open PRs, we should be able to get past these failing tests.

samsja:
> @mryab - is there anything else you need from me? Between the 3 open PRs, we should be able to get past these failing tests.

I have been using this PR for couple of weeks, seems to be stable.

requirements.txt
```diff
@@ -12,5 +12,5 @@ configargparse>=1.2.3
 py-multihash>=0.2.3
 multiaddr @ git+https://github.com/multiformats/py-multiaddr.git@e01dbd38f2c0464c0f78b556691d655265018cce
 cryptography>=3.4.6
-pydantic>=1.8.1,<2.0
-packaging>=20.9
+pydantic>=2.0.0
+packaging>=20.9
```
mryab, line 16: Looks like the file is missing a line break at the end, could you please add it back?

Vectorrent: Done.

mryab (approved):

PR #614, opened 2024-06-09
Update Dockerfile, fix compatibility with PyTorch>=2.3.0

mryab:
The current Dockerfile is based on an outdated version of CUDA, and the Python dependency specification allows Python 3.12 to be installed (which is not supported by Hivemind at the moment). Lastly, this PR fixes an error in `hivemind.optim.grad_scaler` that emerged to a change in the API of PyTorch by using either `torch.amp` or `torch.cuda.amp` depending on the version of PyTorch.

This PR fixes all three issues: to make all of CI workflows pass, both the Python version fix and the `torch.amp` fix are necessary.


justheuristic (approved):
LGTM, thanks!

PR #616, opened 2024-06-09
Improve the process cleanup logic when running tests

mryab:
The current `cleanup_children` function attempts to terminate all child processes, even those that have already stopped. This PR makes the cleanup more robust and avoids terminating dead processes by using `psutil.wait_procs`, which can distinguish dead and alive processes


justheuristic (approved):
LGTM as well. Thanks for finding the time to fix it!

PR #617, opened 2024-06-09
Override the logging level in test_cli_scripts

mryab:
Currently, trying to run a test suite including `test_dht_connection_successful` with `HIVEMIND_LOGLEVEL` other than INFO can lead to intermittent failures, because the first log message might become a debug log entry. This PR overrides the environment when launching `hivemind-dht` from tests to make testing more straightforward


justheuristic (approved):
LGTM

PR #618, opened 2024-06-10
Add downloading time to averager

samsja:
This PR add allows to record the time it takes for downloading from peers. Quite handy (imo) when working with poor bandwith

### DIFF
--- a/hivemind/averaging/averager.py
+++ b/hivemind/averaging/averager.py
@@ -10,6 +10,7 @@ import os
 import random
 import signal
 import threading
+import time
 import weakref
 from dataclasses import asdict
 from typing import Any, AsyncIterator, Dict, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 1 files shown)


samsja:
@mryab feel free to review this one if you have time as well

hivemind/averaging/averager.py
```diff
@@ -722,7 +724,8 @@ async def _load_state_from_peers(self, future: MPFuture, timeout: Optional[float
                             logger.debug(f"Peer {peer} did not send its state")
                             continue
 
-                        logger.info(f"Finished downloading state from {peer}")
+                        t1 = time.time()
```
mryab, line 727: For measuring time intervals, it's best to use [time.monotonic](https://docs.python.org/3/library/time.html#time.monotonic) to make sure the intervals are non-negative

samsja: done

PR #620, opened 2024-06-12
Fix GradScaler import on torch >= 2.3.0 for benchmark_optimizer.py

Vectorrent:


benchmarks/benchmark_optimizer.py
```diff
@@ -98,7 +98,7 @@ def run_trainer(batch_size: int, batch_time: float, client_mode: bool, verbose:
             grad_scaler = hivemind.GradScaler()
         else:
             # check that hivemind.Optimizer supports regular PyTorch grad scaler as well
-            grad_scaler = torch.cuda.amp.GradScaler(enabled=args.use_amp)
+            grad_scaler = torch.amp.GradScaler(enabled=args.use_amp)
```
mryab, line 101: One small question just to make sure: if the user has torch==1.9.0 (earliest supported version in requirements.txt), does this version already have torch.amp? Maybe we need to bump the version in requirements to make sure it won't break existing code?

mryab (commented):
Thanks for the PR! I have one small concern about compatibility, but apart from that we should be good to go

benchmarks/benchmark_optimizer.py
Vectorrent: This part used to exist, but got lost somewhere in splitting all of this work into multiple PRs. I've restored the logic that imports this GradScalar differently here, depending on the installed Torch version.

To answer your question, no, `torch.amp` does not exist in 1.9.

mryab (approved):
Thank you!

PR #621, opened 2024-07-13
Deprecate Python 3.7, switch to Python 3.11 in CI

mryab:
Python 3.7 has reached end-of-line status almost a year ago, which means that we do not need to support it. This PR removes the code that was used to support it (apart from the MPFuture implementation) and switches Actions runners to use Python 3.11 by default


justheuristic (approved):
LGTM

PR #636, opened 2024-11-03
Clean up resources in DHT/P2P, improve test robustness

mryab:
This PR makes several changes to the codebase that aim to make tests less flaky (the situation with their stability is [quite bleak](https://github.com/learning-at-home/hivemind/actions/workflows/run-tests.yml) at the moment).

Specifically, it does the following:
* Disables `fail-fast` for the test matrix to make sure all failing tests can be seen at the same time
* Cleans up resources of DHT/P2P more comprehensively to ensure there are no runaway file descriptors on resource deletion
* Handles SIGTERM gracefully in CLI scripts, as we're using that in tests
* Adds several minor improvements for test robustness and speed (see comments for details)

### DIFF
--- a/hivemind/dht/dht.py
+++ b/hivemind/dht/dht.py
@@ -72,7 +72,7 @@ class DHT(mp.context.ForkProcess):
         self.num_workers = num_workers
 
         self._record_validator = CompositeValidator(record_validators)
-        self._inner_pipe, self._outer_pipe = mp.Pipe(duplex=True)
+        self._inner_pipe, self._outer_pipe = mp.Pipe(duplex=False)
         self.shutdown_timeout = shutdown_timeout
         self._ready = MPFuture()
         self.daemon = daemon
@@ -137,6 +137,7 @@ class DHT(mp.context.ForkProcess):
                     break
 
         loop.run_until_complete(_run())
+        loop.close()
 
     def run_in_background(self, await_ready: bool = True, timeout: Optional[float] = None) -> None:
         """
@@ -154,6 +155,7 @@ class DHT(mp.context.ForkProcess):
         """Shut down a running dht process"""
         if self.is_alive():
             self._outer_pipe.send(("_shutdown", [], {}))
+            self._outer_pipe.close()
             self.join(self.shutdown_timeout)
             if self.is_alive():
                 logger.warning("DHT did not shut down within the grace period; terminating it the hard way")
--- a/hivemind/hivemind_cli/run_dht.py
+++ b/hivemind/hivemind_cli/run_dht.py
@@ -1,6 +1,7 @@
-import time
 from argparse import ArgumentParser
 from secrets import token_hex
+from signal import SIGINT, SIGTERM, signal, strsignal
+from threading import Event
 
 from hivemind.dht import DHT, DHTNode
 from hivemind.utils.logging import get_logger, use_hivemind_log_handler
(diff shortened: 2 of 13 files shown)


hivemind/dht/dht.py
```diff
@@ -72,7 +72,7 @@ def __init__(
         self.num_workers = num_workers
 
         self._record_validator = CompositeValidator(record_validators)
-        self._inner_pipe, self._outer_pipe = mp.Pipe(duplex=True)
+        self._inner_pipe, self._outer_pipe = mp.Pipe(duplex=False)
```
mryab, line 75: Everything works with duplex=False according to tests, best to use pipes since duplex relies on socketpair, which is slower [[1]](https://stackoverflow.com/a/3518624) [[2]](https://stackoverflow.com/a/54164058)

hivemind/hivemind_cli/run_dht.py
```diff
@@ -84,12 +85,19 @@ def main():
     )
     log_visible_maddrs(dht.get_visible_maddrs(), only_p2p=args.use_ipfs)
 
+    exit_event = Event()
+
+    def signal_handler(signal_number: int, _) -> None:
+        logger.info(f"Caught signal {signal_number} ({strsignal(signal_number)}), shutting down")
+        exit_event.set()
+
+    signal(SIGTERM, signal_handler)
```
mryab, line 94: We terminate hivemind-server and hivemind-dht in tests, so a signal handler is necessary to ensure child processes are terminated correctly

hivemind/p2p/p2p_daemon_bindings/utils.py
```diff
@@ -46,6 +46,7 @@ async def write_unsigned_varint(stream: asyncio.StreamWriter, integer: int, max_
             value |= 0x80
         byte = value.to_bytes(1, "big")
         stream.write(byte)
+        await stream.drain()
```
mryab, line 49: drain() is [recommended](https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamWriter.write) after write()

tests/test_util_modules.py
```diff
@@ -262,6 +262,7 @@ def _check_result_and_set(future):
 
     future1, future2 = receiver.recv()
     future1.set_result(123)
+    events[6].wait()
```
mryab, line 265: Without this event, we could `assert future2.done()` in the main process before it's cancelled in the child process

tests/test_util_modules.py
```diff
@@ -514,21 +515,23 @@ async def test_async_context_flooding():
...
     Here's how the test below works: suppose that the thread pool has at most N workers;
     If at least N + 1 coroutines await lock1 concurrently, N of them occupy workers and the rest are awaiting workers;
-    When the first of N workers acquires lock1, it lets coroutine A inside lock1 and into await sleep(1e-2);
+    When the first of N workers acquires lock1, it lets coroutine A inside lock1 and into await sleep();
     During that sleep, one of the worker-less coroutines will take up the worker freed by coroutine A.
     Finally, coroutine A finishes sleeping and immediately gets stuck at lock2, because there are no free workers.
     Thus, every single coroutine is either awaiting an already acquired lock, or awaiting for free workers in executor.
 
     """
+    total_sleep_time = 1
```
mryab, line 524: I don't quite understand the purpose of this test since there are no assert statements, but decided to optimize it a bit (especially on machines with >=128 cores) by setting the total sleep time across all coroutines

dvmazur (approved):

PR #641, opened 2025-02-10
Run CI on Modal, upgrade Bitsandbytes

mryab:
This PR switches the execution of tests from GitHub Actions to [Modal](https://modal.com/), unlocking the option to use GPUs in those tests in the future. Since Modal workers can have multiple CPUs, we can also run tests in parallel, which speeds them up by approximately 4x — from 8 minutes to under 2 minutes. 

Also, as multiple tests seem to be unstable, this PR also disables them for the time being or marks them as flaky with `pytest.mark.xfail`. This will be fixed in future PRs.

Lastly, since the current version of Bitsandbytes is outdated, the PR upgrades it


hivemind/moe/client/moe.py
```diff
@@ -375,19 +379,26 @@ def _collect_responses(
         timeout_total = float("inf") if timeout_total is None else timeout_total
         timeout_after_k_min = float("inf") if timeout_after_k_min is None else timeout_after_k_min
         num_successful_tasks = [0 for _ in range(num_samples)]
-        pending_samples = num_samples  # samples for which we have less than k_min results
+
+        samples_with_tasks = {sample_idx for sample_idx, _ in task_to_indices.values()}
+        pending_samples = len(samples_with_tasks)  # samples for which we have less than k_min results
+        assert pending_samples <= num_samples
```
mryab, line 385: It's better to recompute the number of pending_samples, as some samples might have 0 tasks (for example, in `test_call_many`)

dvmazur (approved):

justheuristic (approved):
Appreciate the version updates and a face lift to the tests. LGTM; pending @dvmazur

PR #652, opened 2025-04-20
Upgrade pytest-asyncio to 0.26.0

mryab:
This PR upgrades pytest_asyncio, following the [migration guide](https://pytest-asyncio.readthedocs.io/en/latest/how-to-guides/migrate_from_0_21.html) given in the plugin documentation


hivemind/utils/asyncio.py
```diff
@@ -195,3 +195,16 @@ async def enter_asynchronously(context: AbstractContextManager):
     """Wrap a non-async context so that it can be entered asynchronously"""
     async with _AsyncContextWrapper(context) as ret_value:
         yield ret_value
+
+
+def cancel_task_if_running(task: Optional[asyncio.Task]) -> None:
```
mryab, line 200: This is introduced to make sure we don't run into errors when trying to cancel tasks with a closed event loop. In particular, this might happen in tests once we remove the current `event_loop` fixture, which seems to be recommended as per the [migration guide](https://pytest-asyncio.readthedocs.io/en/latest/how-to-guides/migrate_from_0_21.html)

justheuristic (approved):
LGTM

PR #664, opened 2025-08-03
Fix Docker build by accepting Conda Terms of Service

mryab:
This PR fixes Conda installation (started erroring recently) by applying the following fix: https://stackoverflow.com/a/79722499


distributedstatemachine:
Please use uv. There is no reason to use conda in 2025. uv has no terms and services

